PR #13, opened 2023-05-31
Bump version as 0.1.1pre

gavanderhoorn:
As per subject.


gavanderhoorn:
I decided not to bump the version displayed in the badge in the readme, as that might be confusing.

Technically we could though, as the `0.1.0` release would contain all `0.1.0` artefacts.

ted-miller (approved):

PR #22, opened 2023-05-31
Clarify INFORM jobs are examples

gavanderhoorn:
As per subject.


gavanderhoorn:
Came up in a private support request.

ted-miller (approved):
I'm wondering if we should take this a step further and actually remove the job folder. We could just outline the contents of the job in the readme.

Issue #21, opened 2023-05-31
Log (also) to `rosout`


gavanderhoorn:
This is on the [roadmap](https://github.com/Yaskawa-Global/motoros2/tree/03e248d6d0c420ba00f2cc00a3425fc71c449317#provisional-roadmap) as:

> integration with ROS logging (`rosout`)

Low priority, as we already have the debug logger, but this could potentially be a nice way to integrate MotoROS2 even more with ROS 2.

PR #23, opened 2023-05-31
Log (also) to rosout

gavanderhoorn:
This is for #21.

Draft currently, as this is a rebase of old code on top of current `HEAD` and I haven't tested this recently.

Enabling `rosout` logging was simpler than I expected, it essentially only needed a call to `rcl_logging_configure_with_output_handler(..)`.

I've opted to make it configurable, as it does introduce overhead *and* isn't really very useful right now.

NOTE: this doesn't actually result in any log msgs on `rosout` *yet*, as we don't use any of the `RCUTILS_LOG_*(..)` macros. We could opt for forwarding all dbg log msgs to `rosout` via `RCUTILS_LOG_DEBUG_NAMED(..., msg)`, but I'm not yet sure I'd like that (for one thing, that would make all debug log msgs have the `file` field set to `Debug.c:<some_line_nr>` which wouldn't be very informative).

Example of the messages received by a subscriber to `rosout`:

```yaml
stamp:
  sec: 1660325294
  nanosec: 991000000
level: 20
name: motoman_36_ad_6e
msg: 'Pinged agent: 0'
file: C:/.../MicroMotoRos/CommunicationExecutor.c
function: Ros_Communication_PingAgentConnection
line: 207
```


gavanderhoorn:
As I wrote elsewhere: I'm a bit on the fence about whether or not to merge this in.

It works, but I'm not sure this gains us anything, unless we start using the ROS logging macros throughout MotoROS2 as well. But even if we do that, the use case

> as a user of MotoROS2, I'd like to be able to see logging output in the terminal, next to the output of my other ROS 2 nodes

will not immediately be addressed, as there's nothing forwarding any of those messages to the terminal which happens to start those "other ROS 2 nodes".

We could add something like that -- and we have had requests to do that -- but that would basically be the debug listener script turned into a ROS 2 node (and so forwarding all msgs coming in over UDP via `RCUTILS_LOG_DEBUG_NAMED(..)` calls). That would not need MotoROS2 using `rosout` as a logging sink.

On the other hand, we also don't lose anything by merging this. It just hooks up some parts of (micro-)ROS which are already included.

src/MotoROS2_AllControllers.vcxproj
```diff
@@ -89,7 +89,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='YRC1000_humble|Win32'">
     <NMakeBuildCommandLine>"$(MP_VS_Install)mpBuilder.exe" -c $(Configuration) -p "$(ProjectDir)\" -n "motoros2_yrc1000_humble" -b "$(SolutionDir)bin\$(Configuration)" -o build -i "$(MP_VS_Install)YRC1000\inc"</NMakeBuildCommandLine>
     <NMakeOutput>$(MP_VS_Install)DevTools\OnlineDownload.exe</NMakeOutput>
-    <NMakePreprocessorDefinitions>$(NMakePreprocessorDefinitions);YRC1000;MOTOROS2_MEM_TRACE_ENABLE</NMakePreprocessorDefinitions>
+    <NMakePreprocessorDefinitions>$(NMakePreprocessorDefinitions);YRC1000;MOTOROS2_MEM_TRACE_ENABLE;RCUTILS_LOG_MIN_SEVERITY=RCUTILS_LOG_MIN_SEVERITY_DEBUG</NMakePreprocessorDefinitions>
```
gavanderhoorn, line 92: I'm not sure this will work as I expect / need it to. I also haven't tested it.

src/CommunicationExecutor.c
```diff
@@ -122,7 +122,8 @@ void Ros_Communication_Initialize()
     motoRosAssert(ret == RCL_RET_OK, SUBCODE_FAIL_SUPPORT_INIT);
 
     //construct a faux command line to pass to rcl_parse_arguments(..)
-    char* faux_argv[MAX_REMAP_RULE_NUM];
+    const size_t FAUX_ARGV_LEN = MAX_REMAP_RULE_NUM + 1 + 1 + 1;
```
gavanderhoorn, line 125: Not too happy with the `1 + 1 + 1` part.

We should probably refactor this at some point.

src/MotoROS2_AllControllers.vcxproj
ted-miller: The `NMakePreprocessorDefinitions` are strictly for intellisense. The actual compiler flags are in `YRC1000_xxxCompilerArguments.mps`

gavanderhoorn: Maybe these should actually be added to the `motoplus_libmicroros_config.h`.

gavanderhoorn:
I'm going to close this one, as, while it works as intended, it does not actually match the expectations of (potential) users.

As mentioned in https://github.com/Yaskawa-Global/motoros2/pull/23#issuecomment-1570168005, log messages would not (automatically) end up in the terminal/console, but on the `/rosout` topic.

Additionally, changes to `rcl` (in micro-ROS) in Iron and newer complicate maintaining this functionality.

PR #27, opened 2023-06-01
Add user lan link state monitoring

gavanderhoorn:
This PR introduces a new 'link state monitoring' task, started by MotoROS2 right after it has (re)connected to a micro-ROS Agent instance.

The responsibility of this task is to check whether the configured ethernet port still reports its ethernet link as being 'up'. As long as it's up, the task does nothing. As soon as a link drop is detected, the task sets `g_Ros_Communication_AgentIsConnected` to `FALSE` and exits.

Setting `g_Ros_Communication_AgentIsConnected` to `FALSE` initiates MotoROS2's shutdown behaviour, resulting in cancelling any active trajectory goals, stopping active motion, relinquishing resources and periodically attempting to reconnect to the Agent.

The current implementation is a *draft*, as while it works, I wanted to discuss a few aspects before marking it as ready for merging into MotoROS2.


src/CommunicationExecutor.c
```diff
@@ -207,6 +207,62 @@ void Ros_Communication_PublishActionFeedback(rcl_timer_t* timer, int64_t last_ca
...
+    {
+        //TODO(gavanderhoorn): make configurable ?
+        Ros_Sleep(0.5 * 1e3);
+
+        //retrieve current state of monitored link, assume it's down
+        BOOL bLinkIsUp = FALSE;
+        STATUS status = Ros_UserLan_IsLinkUp(interface_id, &bLinkIsUp);
+        if (status != OK)
+        {
+            //TODO(gavanderhoorn): should this be fatal for MotoROS2?
```
gavanderhoorn, line 234: Should it perhaps be?

If a user has kept the link state monitor enabled (by not disabling it via the config file), it might be reasonable to say any problem trying to run it should probably cause alarms to be raised.

src/CommunicationExecutor.c
```diff
@@ -207,6 +207,62 @@ void Ros_Communication_PublishActionFeedback(rcl_timer_t* timer, int64_t last_ca
...
+            //yes, goto has its uses
+            goto UserLanMonitorExit;
+        }
+
+        //link was up last time, but no longer is up now -> signal disconnect
+        if (bLinkWasUp && !bLinkIsUp)
+        {
+            //log to console, as network connection is presumable down
+            Ros_Debug_LogToConsole(
+                "Link down (port: %d), flagging Agent as disconnected", interface_id);
```
gavanderhoorn, line 251: We can't UDP broadcast here, as the link is down.

I also didn't want to have to set `log_to_stdout: true` for someone to be able to see this.

This is now one of a handful of log msgs which make it to `stdout`.

src/ConfigFile.c
```diff
@@ -117,6 +117,9 @@ Configuration_Item Ros_ConfigFile_Items[] =
     { "stop_motion_on_disconnect", &g_nodeConfigSettings.stop_motion_on_disconnect, Value_Bool },
     { "inform_job_name", &g_nodeConfigSettings.inform_job_name, Value_String },
     { "allow_custom_inform_job", &g_nodeConfigSettings.allow_custom_inform_job, Value_Bool },
+    { "userlan_monitor_enabled", &g_nodeConfigSettings.userlan_monitor_enabled, Value_Bool },
+    { "userlan_monitor_port", &g_nodeConfigSettings.userlan_monitor_port, Value_Int },
+    { "userlan_monitor_max_num_drops", &g_nodeConfigSettings.userlan_monitor_max_num_drops, Value_Int },
```
gavanderhoorn, line 122: The current implementation does not include any sort of "debouncing logic", so this setting is currently not used.

Personally, I'd like to be notified of *any* drops, but I could be persuaded to allow configuring a specific threshold (ie: nr of drops) after which the disconnection logic should kick in.

(although perhaps a maximum amount of time for a link to remain down would make more sense)

src/MotoROS_PlatformLib.h
```diff
@@ -18,7 +18,7 @@
 // -  ZZ: patch version nr
 //
 // Note: no leading zeros, to avoid octal parsing
-#define MOTOROS_PLATFORM_LIB_VERSION 202
+#define MOTOROS_PLATFORM_LIB_VERSION 203
```
gavanderhoorn, line 21: We'll need to import the updated PlatformLib for this to build correctly.

gavanderhoorn:
In local testing this worked rather nicely.

Disconnecting my YRC1000 from the network resulted in almost immediate shutdown of MotoROS2.

Combined with the reduced shutdown time released as part of `0.1.0`, this makes for a very responsive MotoROS2 in case of connection issues.

gavanderhoorn:
@ted-miller: thoughts?

config/motoros2_config.yaml
```diff
@@ -70,6 +70,33 @@
...
+# Should MotoROS2 monitor the link state of the ethernet port used to
+# communicate with the Agent?
+#
+# If enabled, this will cause immediate notification of loss-of-connection
+# when the link goes down, causing MotoROS2 to activate it's shutdown
+# behaviour (ie: stop all motion (if enabled), release resources, wait for
+# link up and attempt reconnection with the Agent).
+#
+# If this is disabled, Agent disconnection detection uses an application
+# layer based timeout mechanism which takes longer to detecet disconnects (but
```
ted-miller, line 83:
```suggestion
# layer based timeout mechanism which takes longer to detect disconnects (but
```

src/CommunicationExecutor.c
```diff
@@ -353,7 +409,19 @@ void Ros_Communication_StartExecutors(SEM_ID semCommunicationExecutorStatus)
...
     //===========================================================
     //===========================================================
-    
+
+    // See if we should start task which monitors user lan port link state
+    // (This task deletes itself when the agent disconnects.)
+    if (g_nodeConfigSettings.userlan_monitor_enabled)
+    {
+        Ros_Debug_BroadcastMsg("Starting UserLan link state monitor task ..");
+        mpCreateTask(MP_PRI_TIME_NORMAL, MP_STACK_SIZE,
```
ted-miller, line 418: I'm a little concerned about starting up another task. I'm worried that the all the task switching may be contributing to delays/latency.

Could we put this functionality in `main.c` inside the loop in `RosInitTask`?

config/motoros2_config.yaml
```diff
@@ -70,6 +70,33 @@
...
+# might be more robust against intermittent link-layer disconnects which do not
+# cause network disconnections).
+#
+# DEFAULT: true
+#userlan_monitor_enabled: true
```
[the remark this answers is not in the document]
ted-miller: I agree with a default value being true

config/motoros2_config.yaml
```diff
@@ -70,6 +70,33 @@
...
+
+#-----------------------------------------------------------------------------
+# Which port should MotoROS2 monitor, if 'userlan_monitor_enabled' is 'true'?
```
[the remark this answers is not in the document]
ted-miller: I really believe we should automatically detect this. It's nicer for the user.

Plus, a future robot controller could _potentially_ have additional ports which would confuse the issue more.

src/CommunicationExecutor.c
```diff
@@ -207,6 +207,62 @@ void Ros_Communication_PublishActionFeedback(rcl_timer_t* timer, int64_t last_ca
...
+    //connect to it), so the link MUST be up or otherwise this task could
+    //not have been started.
+    //If we start with the assumption the link is down, we'd see a state change
+    //as soon as we call Ros_UserLan_IsLinkUp(..), which would not lead to any
+    //action (as we only act on up->down edges), but would be unnecessary.
+    BOOL bLinkWasUp = TRUE;
+
+    //we set g_Ros_Communication_AgentIsConnected=FALSE in the body below,
+    //so this exits when a link drop is detected
+    while (g_Ros_Communication_AgentIsConnected)
```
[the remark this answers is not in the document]
ted-miller: Reuse is good. It's easier to track a single variable for determining the link status.

src/CommunicationExecutor.c
ted-miller: My concern is based on a little bit more than assumption. I have empirically measured delays with spooling up too many tasks for another project. (Granted, that situation had many more tasks than MotoROS2.)

While the motion task is indeed at a higher priority, a `normal` priority task is the lowest priority of anything else in the robot controller. 

You're _probably_ right that this one task won't cause noticeable delay. I'm just nervous because we frequently have users wanting "maximum" speed and zero latency.

config/motoros2_config.yaml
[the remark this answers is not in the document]
gavanderhoorn: Bah. Auto-detecting means more logic to add :(

But ok, I'll see if I have time to look into it.

src/CommunicationExecutor.c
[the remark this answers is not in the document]
gavanderhoorn: Hm, so technically it wouldn't necessarily be correct, especially not when we're going to add debouncing logic.

But we can leave it like this for now.

src/CommunicationExecutor.c
gavanderhoorn: But that doesn't post any alarms, does it?

src/ConfigFile.c
gavanderhoorn: Do you feel a 'flaky' link would be something we should tolerate?

src/CommunicationExecutor.c
gavanderhoorn: > My concern is based on a little bit more than assumption. I have empirically measured delays with spooling up too many tasks for another project. (Granted, that situation had many more tasks than MotoROS2.)

Would you happen to know what sort of task scheduling is configured on the controller?

I could see round-robin resulting in significantly increased delays with a high nr of tasks.

> While the motion task is indeed at a higher priority, a `normal` priority task is the lowest priority of anything else in the robot controller.

So are you worried the monitoring task might get starved? Or am I misunderstanding you?

> You're _probably_ right that this one task won't cause noticeable delay. I'm just nervous because we frequently have users wanting "maximum" speed and zero latency.

Sure.

I'll see if I can setup some tracing/measurement infrastructure and compare.

src/CommunicationExecutor.c
ted-miller: You're correct there's no alarm. I'd be ok with adding an alarm, but I still don't think it should be fatal. The user can try a different cable w/o rebooting.

src/CommunicationExecutor.c
ted-miller: > Would you happen to know what sort of task scheduling is configured on the controller?

From M+ doc:

> The task scheduling method is preemptive scheduling based on priority.
The tasks in the Ready state with the same priority are executed in a timeshared manner by round robin scheduling.

------------
> So are you worried the monitoring task might get starved? Or am I misunderstanding you?

I'm worried that all the tasks (publishers, executor, services, etc) would be starved and reduce the "responsiveness".

src/MotoROS_PlatformLib.h
gavanderhoorn: 7274a05f18f51243218e388e774f29791fcfbca0

src/ConfigFile.c
gavanderhoorn: I'm going to assume hw debouncing is sufficient for now.

We can always add another layer ourselves, but all my testing to far shows whenever the connection is unstable, it's really either the cable, connector or "the other side" (ie: switch, PC, connector on that side).

src/CommunicationExecutor.c
ted-miller: > Are we sure a single additional task running at 2 Hz will cause problems?

No. But I'm very paranoid about such things. 

> https://github.com/Yaskawa-Global/motoros2/commit/c59e1b8e132bcdbfd7d9eb561d87031351340748 changes the task into a rcl_timer_t callback

I like this idea

config/motoros2_config.yaml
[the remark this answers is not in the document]
gavanderhoorn: I believe I've implemented something in this direction.

It's pretty naive though: it will just check whether the Agent IP is reachable on a particular interface and assume that's the one it should monitor.

config/motoros2_config.yaml
```diff
@@ -70,6 +70,33 @@
...
+#
+# NOTE: this setting only applies to YRC1000 and YRC1000u controllers.
+#       DX200 and FS100 controllers have only a single ethernet port, and any
+#       value other than 1 will be ignored.
+#
+# DEFAULT: 1 (ie: 'ROS_USER_LAN1')
+#userlan_monitor_port: 1
```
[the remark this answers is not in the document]
gavanderhoorn: Decided to implement this.

Changed constants to `USER_LAN1` and `USER_LAN2`.

src/CommunicationExecutor.c
gavanderhoorn: An alarm will now be posted. Non-fatal.

gavanderhoorn:
Ok, so I've:

 1. reworked this to use a timer callback instead of a task
 2. re-implemented cfg file parsing such that it accepts `USER_LAN1` and `USER_LAN2` settings, instead of just plain integers (thought that was nicer for users)
 3. implemented some form of auto-detection (but see my earlier comment about it's limitations)

While it still works, because of the fact the timer is serviced by the same executor as the 'ping agent' callback, depending on when exactly the link drops, the new link state check functionality will have to 'wait' for the ping-to-agent to time out. This sometimes leads to noticeable delay, especially when compared to the previous implementation.

We could give it its own executor, or use the IO one instead (that's in a different task already I believe).

Auto-detection works, but is 'stupid'. Basic algorithm:

```
if agent_on_nic1:
  monitor nic1
else:
  if gateway_on_nic1:
    // assume agent can be reached via gateway
    monitor nic1

// repeat for nic2
```

we're not actively pinging the prospective agent IP, so auto-detection can't really be sure of anything.

If auto-detection is not used (by specifying a port to monitor), the current implementation allows to configure a port to monitor which is *not* used for MotoROS2<->Agent traffic. Technically I could check whether the port configured is the one the Agent IP *should* be reachable on, but that isn't implemented.

All-in-all it still works, it's just a tad slower and a bit more flexible / easier to configure.

src/CommunicationExecutor.c
```diff
@@ -353,7 +420,20 @@ void Ros_Communication_StartExecutors(SEM_ID semCommunicationExecutorStatus)
     //===========================================================
     //===========================================================
     //===========================================================
-    
+
+    // See if we should enable the timer which monitors user lan port link state
+    if (!g_nodeConfigSettings.userlan_monitor_enabled)
+    {
+        rc = rcl_timer_cancel(&timerMonitorUserLanState); RCL_UNUSED(rc);
```
ted-miller, line 427: I don't undersatnd this cancel operation. The timer hasn't been started yet, has it?

src/ConfigFile.h
```diff
@@ -74,6 +78,17 @@ typedef enum
 #define MAX_REMAP_RULE_NUM              16
 #define MAX_REMAP_RULE_LEN              256
 
+typedef enum
+{
+    CFG_ROS_USER_LAN_DISABLED = -2,  //sentinel
```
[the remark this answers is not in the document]
gavanderhoorn: it's a legal `enum` value, but it has a special meaning -- it's not actually a value mapped to a LAN port. It prevents me from having to create an additional constant/`enum`.

gavanderhoorn:
> While it still works, because of the fact the timer is serviced by the same executor as the 'ping agent' callback, depending on when exactly the link drops, the new link state check functionality will have to 'wait' for the ping-to-agent to time out.

I realised this basically makes the new functionality a slightly faster version of the 'ping timeout', which reduces its value significantly.

src/CommunicationExecutor.c
gavanderhoorn: AFAIK, timers don't need to be started explicitly. They start 'running' as soon as the executor starts spinning.

For the `executor_motion_control` this timer is added to, that would be as soon as it enters the `while (g_Ros_Communication_AgentIsConnected)` that follows. 

We unconditionally add the timer to the executor, so if the lan state monitor was disabled via the config file, we need to disable it before the executor spins.

src/CommunicationExecutor.c
```diff
@@ -268,6 +325,12 @@ void Ros_Communication_StartExecutors(SEM_ID semCommunicationExecutorStatus)
     rc = rclc_executor_add_timer(&executor_motion_control, &timerPublishActionFeedback);
     motoRosAssert_withMsg(rc == RCL_RET_OK, SUBCODE_FAIL_TIMER_ADD_ACTION_FB, "Failed adding timer (%d)", (int)rc);
 
+    //NOTE: add userlan monitor timer to the io executor, to prevent timerPingAgent
+    //from blocking it in case agent connection is lost (ie: ping needs to time out)
+    rc = rclc_executor_add_timer(&executor_io_control, &timerMonitorUserLanState);
```
gavanderhoorn, line 330: moved the timer to the io executor (from the motion executor). Both to not block motion callbacks while it runs and also to avoid the ping Agent callback from blocking execution of the lan state monitor callback.

src/ConfigFile.c
```diff
@@ -438,6 +478,38 @@ void Ros_ConfigFile_CheckUsbForNewConfigFile()
         Ros_Debug_BroadcastMsg("No new configuration file found on CN102 USB drive.");
 }
 
+STATUS Ros_ConfigFile_HostOnNetworkInterface(char* const host, USHORT if_no, BOOL* const reachable)
```
gavanderhoorn, line 481: pedantically, this wouldn't belong here. It's not related to config file parsing.

We could decide to move it elsewhere.

gavanderhoorn:
I've just tested the current version again (after the changes in recent commits) and it still easily 'outperforms' the Agent-ping-timeout.

Both are useful though: the Agent could go off-line without the link going down, and the other way around.

I'm not too happy about (ab)using the IO executor to spin this timer, so if we'd want, I could add another executor and spin that from the IO executor's task. That would not increase the nr of tasks, but would decouple the IO callbacks from the link state timer callback.

src/ConfigFile.c
ted-miller: I can't think of a better place for it, unless we make a whole new file.

ted-miller:
> I'm not too happy about (ab)using the IO executor to spin this timer, so if we'd want, I could add another executor and spin that from the IO executor's task. That would not increase the nr of tasks, but would decouple the IO callbacks from the link state timer callback.

I think another executor would just complicate things more. If we add any future additional "utility" type functions, then it would probably be worth making a dedicated executor. But I think it's ok to piggyback for this one.

gavanderhoorn:
@yeu-buchholz: I remember a discussion on the `beta1` discussion forum where you reported either an issue or asked a question about what would happen if MotoROS2 lost connection with the Agent.

While MotoROS2 already detected Agent 'disappearance' using a ping timeout, this PR adds detection of link state, which doesn't use any thresholds (ie: timeout values), but detects almost instantly whether there is no connection to the Agent.

yeu-buchholz:
I agree with you, @gavanderhoorn. Upon disconnecting the ETH cable from the Agent, I immediately observed a detection on the Agent site and the robot motion came to a halt within a time frame of approximately 5-7 seconds. Please note that this estimate may not be entirely accurate.

gavanderhoorn:
But that was without this PR, correct? So purely based on the ping time-out?

That would indeed detect a disappearing Agent in about 5 sec.

yeu-buchholz:
You are correct. I did not have this particular PR. 
Here's the test I conducted:
The robot was being controlled by a program from the PC. I disconnected the cable and immediately noticed that the agent was disconnected. After 5-7 seconds, the robot stopped moving. I did not use the ```ping``` command this time, as I was simply observing out of personal interest. If you, @gavanderhoorn, would like me to conduct any other tests, I am happy to run.

gavanderhoorn:
If you'd be interested, you could see whether the same test now results in faster cancellation of active motion.

This is not high priority though.

yeu-buchholz:
I will update you after the test!

yeu-buchholz:
I ensure that all software is up-to-date and conduct the following tests:
#### without robot motion
1. Verify that there is a connection between ```YRC1000``` and the ```agent``` (referred to as ```terminal 1```).
2. Ping the controller (referred to as ```terminal 2```). Confirm that it is working correctly.
3. Disconnect the ETH cable from the ```agent```.
4. ```terminal 2``` - ping command reports the failure first.
5. ```YRC1000 terminal``` should report disconnection as a second, lasting 3-4 seconds.
6. After an additional 2 seconds, the ```agent``` (referred to as ```terminal 1```) reports: ```Session destroyed.```
7. Reconnect the ETH cable to regain all connections. Confirm that everything is working correctly (```ALL OK```).

#### with robot motion

1. Verify that there is a connection between ```YRC1000``` and the ```agent``` (referred to as ```terminal 1```).
2. Ping the controller (referred to as ```terminal 2```). Confirm that it is working correctly.
3. Start robot motion. Robot moves. 
4. Disconnect the ETH cable from the ```agent```.
5. ```terminal 2``` - ping command reports the failure first.
6.  after 3-4 seconds robot motion halts.
7.  after additional 1-2 seconds ```YRC1000 terminal``` reports disconnection.
8. After an additional 1-2 seconds, the ```agent``` (referred to as ```terminal 1```) reports: ```Session destroyed.```
9. Reconnect the ETH cable to regain all connections  (```ALL OK```). Robot motion is NOT recovered, as expected


If you @gavanderhoorn need more tests to perform, just let me know.

PR #46, opened 2023-06-02
Replace in-line tabs with spaces

gavanderhoorn:
As per subject.

It's a sad day, but this would see the last of the tabs be replaced by spaces in MotoROS2 `.h`s and `.c`s (well, in files where it makes sense).

Our `.editorconfig` already forced supporting editors to use spaces, but it didn't replace the ones that were already there.

No functional changes.


ted-miller (approved):
A sad day, indeed. Goodbye my friends...

Issue #18, opened 2023-05-31
Add support for DX200


gavanderhoorn:
We'll expand this issue to track status of sub issues where we work on adding support for DX200 controllers to MotoROS2.

ted-miller:
Known issues at this time:

- Need to read robot calibration data (obfuscated in PlatformList)
- The command to create the node is returning an error code `RCL_RET_NODE_INVALID_NAME`. (I haven't done any debug on it yet.)
- Setup build system for libmicroros
- Update ParameterExtraction
- Update PlatformList
- Setup VS project configurations and related support files (mps)

gavanderhoorn:
@smith-doug @robinsonmm

gavanderhoorn:
@smith-doug @robinsonmm: would the robot you could potentially use to test happen to be mounted on a rail/track? If it isn't, we might be able to provide you with a test-build sooner.

acbuynak:
If helpful as a test case, I can offer a 12-axis (2 x 6R arms) system controlled by two DX200's in a master-slave configuration.

ted-miller:
hmm... I looks like `Ros_CtrlGroup_GetEncoderTemperature` is not actually used. I suspect this was copied from MotoROS1 in anticipation of future functionality. We should just remove it for now.

gavanderhoorn:
I'm pretty sure `DN3.19.00a` does have `mpGetEncoderTemp(..)`, and that's the version I verified symbols against.

> hmm... I looks like `Ros_CtrlGroup_GetEncoderTemperature` is not actually used. I suspect this was copied from MotoROS1 in anticipation of future functionality. We should just remove it for now.

We (I) added that about a year ago. Preparation for publishing controller statistics.

robinsonmm:
@gavanderhoorn our DX200 is NOT on a track.

gavanderhoorn:
@robinsonmm: then it would be great if you could perhaps also test the `alpha1` binary linked in https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1584968247.

It's expected to *not* work, but the test would help figure out at which point it breaks.

ted-miller:
wrt https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1587956034

I did some more "debugging" and got some weird results. 

[...]

PR #49, opened 2023-06-02
Support for DX200 controllers

gavanderhoorn:
Edit 2: current status:

 - all identified outstanding issues appear to have been resolved (IP address issues: https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1668578286)
 - testing with Humble shows MotoROS2 is functional on DX200 (https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1658947030, https://github.com/Yaskawa-Global/motoros2/pull/49#pullrequestreview-1555598540)
 - Success with Foxy has also been reported (https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1624352260)

Remaining TODO:

 - [x] clean up commit history

---

Edit: current status:

 - Humble and Foxy builds *appear* to work on DX200, *but*
 - ~~IP addresses with last octet `>= 128` will result in a `Must Enable ETHERNET function` (ie: Alarm `8011`) to be raised. Anything with last octet `< 128` is expected to work (as a work-around): https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1625988944~~: this should no longer be a problem.

Available test builds:

 - Humble: https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1638410308
 - Foxy: https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1613766820

---

This is for #18.

A *draft* currently, but first steps towards full DX200 support.


gavanderhoorn:
@ted-miller: could you please add the DX2 version of the `ParameterExtraction` library?

src/MotoROS2_AllControllers.vcxproj.filters
```diff
@@ -285,6 +300,9 @@
     <ClCompile Include="ServiceStartPointQueueMode.c">
       <Filter>Source Files\Services</Filter>
     </ClCompile>
+    <ClCompile Include="Ros_mpGetRobotCalibrationData.c">
+      <Filter>Source Files\Robot Controller</Filter>
```
gavanderhoorn, line 304: @ted-miller: feel free to move this to another 'location'.

gavanderhoorn:
> @ted-miller: could you please add the DX2 version of the `ParameterExtraction` library?

@ted-miller: friendly ping.

src/MotoROS2_AllControllers.vcxproj.filters
ted-miller: This is fine where it is

ted-miller:
The binary has been added. The lib was already included in the `vcxproj` file.

gavanderhoorn:
@acbuynak: we're going to use this PR to distribute test binaries and discuss progress.

Please see [motoros2_dx200_9.9.9-alpha0.zip](https://github.com/Yaskawa-Global/motoros2/files/11692195/motoros2_dx200_9.9.9-alpha0.zip), which, as the name implies, is an `alpha`. I don't have access to a DX200 right now, so I've only build-tested this.

The `.zip` contains only the `.out`, so please take the config file from either the repository or one of the YRC1000(u) releases. The installation process would be largely equivalent to the YRC1000 instructions. We've not yet updated them to account for differences with the DX200.

Please make sure to run the debug listener at all times (even while installing), and start it before you power up the controller.

Just to make it absolutely clear again:

1. this is an `alpha0` version and may absolutely **not** be used for any kind of production or non-testing scenarios
1. we've not tested this ourselves yet
1. TF published by MotoROS2 will be incorrect for your dual group system
1. at this point we're mostly interested in knowing whether it starts up correctly on your system, and the debug log shows the expected output
1. *if* attempting to command motion, you do so at your own risk. In all cases, keep a close watch on your robots and your hand on the e-stop

Whether success or not, we'd be interested in the debug logs. So please zip and then attach them here in a comment. Please also tell us the version of system software on your DX200. A copy of your `PANELBOX.LOG` would also be greatly appreciated.

Thanks already for offering to test 👍

gavanderhoorn:
Seems like more motivation to redesign MotoROS2 as a life-cycle node and use ROS parameters .. (although we still have the problem with string support there).

ted-miller:
@gavanderhoorn : Can you build an `alpha1`? I'm not configured for building DX200 yet

src/ConfigFile.c
```diff
@@ -572,15 +575,32 @@ void Ros_ConfigFile_Parse()
...
+
+        snprintf(configFilePath, CHAR_BUFFER_SIZE, "%s\\%s", storageDrive, CONFIG_FILE_NAME);
 
         Ros_Debug_BroadcastMsg("Checking configuration file: %s", CONFIG_FILE_NAME);
 
-        fd = mpOpen(sramFilePath, O_RDONLY, 0);
+        fd = mpOpen(configFilePath, O_RDONLY, 0);
+
+#if (DX200)
+        if (fd < 0)
```
gavanderhoorn, line 597: Might be nice to add a debug log here.

ted-miller:
To clarify, with the latest commits, a user will need to put the config file on the root of a USB drive. This drive must be connected to `CN106` on the CPU board inside the robot controller. This must reside there permanently. For long term usage, it is recommended to use an industrial-grade drive with a higher temperature rating. 

On production systems for other applications, we usually use these: https://www.verbatim.com/prod/usb-drives/everyday-usb-drives/ToughMAX-usb-drive-sku-70000/

src/ConfigFile.c
ted-miller: I considered that and then figured that a user wouldn't care and that information would just add noise. Physically, it just looks like a USB hub and a user would just plug in the drive arbitrarily. 
image

gavanderhoorn: I haven't checked the changes in detail, but how would the "no usb stick inserted" situation be handled?

gavanderhoorn:
> @gavanderhoorn : Can you build an `alpha1`? I'm not configured for building DX200 yet

yes, I'll do it later today.

> On production systems for other applications, we usually use these: https://www.verbatim.com/prod/usb-drives/everyday-usb-drives/ToughMAX-usb-drive-sku-70000/

hm. 16GB to store 200 bytes? :)

gavanderhoorn:
Looks like the `.md` formatting isn't to the liking of the linter. It's posted a couple of in-line annotations (on the *Files changed* tab).

gavanderhoorn:
@acbuynak: please see [motoros2_dx200_9.9.9-alpha1.zip](https://github.com/Yaskawa-Global/motoros2/files/11709073/motoros2_dx200_9.9.9-alpha1.zip).

This should include the changes needed to work without the SRAM add-on.

No other changes have been made, so refer to my earlier https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1583213189.

ted-miller:
@acbuynak: Please rename the .out file before attempting to load it into the robot controller. I just loaded it in mine and it doesn't appear in the file list. Please rename to `motoros2_alpha1.out` and I expect it to work.

ted-miller:
Scratch that. This build is missing a symbol. We'll work on a new version.

ted-miller:
I got past the missing symbol and was able to get it to build. 

```
[1686343673.21892715]: ---
[1686343673.21892715]: MotoROS2 0.1.1pre - boot
[1686343673.21892715]: M+ libmicroros version: 'humble-20221102-dbg'
[1686343673.21892715]: PlatformLib version: 0.2.4
[1686343673.22693777]: Checking configuration file: motoros2_config.yaml
[1686343673.23651791]: Config: agent_ip_address = 192.168.1.78
[1686343673.23651791]: Config: agent_port_number = 8888
[1686343673.23651791]: Config: sync_timeclock_with_agent = 1
[1686343673.23651791]: Config: stop_motion_on_disconnect = 1
[1686343673.23751831]: Config: publish_tf = 1
[1686343673.23751831]: Config: namespace_tf = 1
[1686343673.23751831]: Config: log_to_stdout = 0
[1686343673.23851824]: Config: executor_sleep_period = 10
[1686343673.23951817]: Config: action_feedback_publisher_period = 20
[1686343673.23951817]: Config: controller_status_monitor_period = 10
[1686343673.23951817]: Config: robot_status = sensor_data
[1686343673.24251914]: Config: joint_states = sensor_data
[1686343673.24251914]: Config: tf = default
[1686343673.24251914]: Waiting for robot alarms to clear...
[1686343675.74765921]: RMW settings 1: 1000; 0; 1000; 256; 256; 512
[1686343675.74765921]: RMW settings 2: 1; 1; 30; 15; 30; 1; -1
[1686343675.74765921]: RMW settings 3: 60; 120; 100; 100
[1686343675.74765921]: Using ROS domain ID: 0
[1686343675.74765921]: Using client key: 0x7385C897
[1686343675.74765921]: Attempting to connect to Micro-ROS PC Agent (at udp://192.168.1.78:8888)
[1686343676.77116156]: Found Micro-ROS PC Agent
[1686343676.77418017]: rclc_support_init_with_options = 0
[1686343676.77418017]: remap_rules str: ''
[1686343676.77418017]: len(remap_rules str): 0
[1686343676.77418017]: num parsed remap rules: 0
[1686343676.77418017]: faux argc: 2
[1686343676.77418017]: rcl_parse_arguments = 0
[1686343677.77579498]: rclc_node_init_with_options = 1
[1686343677.81079793]: motoRosAssert: MotoROS2: Fatal Error (subcode: 5)
```

subcode 5 is `SUBCODE_FAIL_NODE_INIT`
https://github.com/Yaskawa-Global/motoros2/blob/cf9d956d0ff31d962fae305ed9a25779986b608e/src/CommunicationExecutor.c#L155-L160

src/CtrlGroup.c
```diff
@@ -443,32 +471,32 @@ BOOL Ros_CtrlGroup_GetTorque(CtrlGroup* ctrlGroup, double torqueValues[MAX_PULSE
...
+//    if (0 != status)
+//    {
+//        Ros_Debug_BroadcastMsg("Failed to get encoder temperature: %ld", status);
+//        return FALSE;
+//    }
+//
+//    memcpy(encoderTemp, rData.lTemp, MAX_PULSE_AXES * sizeof(long));
+//
+//    return TRUE;
+//}
```
gavanderhoorn, line 499: Seem https://github.com/Yaskawa-Global/motoros2/issues/18#issuecomment-1585503628. I'm not sure this should be needed.

gavanderhoorn:
> `[1686343677.77579498]: rclc_node_init_with_options = 1`

we'll have to trace / enable more debug output, as `1 == RMW_RET_ERROR`, which is the generic "something didn't work" return code (from [here](https://github.com/ros2/rmw/blob/e523e4706ae6e20134fb835a91d18b74a134175e/rmw/include/rmw/ret_types.h#L29-L30)).

ted-miller:
> we'll have to trace / enable more debug output,

I liberally added a bunch of debug broadcasts. I got this deep:

https://github.com/micro-ROS/rmw_microxrcedds/blob/439073381b54a2fcaeeeea64c740a7f6077a8513/rmw_microxrcedds_c/src/utils.c#L39-L44

It seems the failure is occurring in Micro-XRCE-DDS-Client. Will continue digging later in the week.

gavanderhoorn:
It might be worthwhile to see whether [rcutils_get_error_string()](https://docs.ros2.org/latest/api/rcutils/error__handling_8h.html#abb64e38af3108094a76dadeb4a88f773) could help.

If the failing function calls `RCL_SET_ERROR_MSG(..)`, it might give a clue as to what is going wrong.

src/CtrlGroup.c
gavanderhoorn: Would it be OK if I revert 46cfdcda4905333fc53590fac5ab6e83b97a1d85? Just to keep the history clean in this PR?

gavanderhoorn:
@acbuynak: even though it would appear from @ted-miller's tests MotoROS2 will currently not run on DX200, I'd still be curious to know whether you run into the same issue as @ted-miller.

If you would have 5 minutes to test, I'd greatly appreciate it.

src/CtrlGroup.c
ted-miller: Yes.
I did [0b15446](https://github.com/Yaskawa-Global/motoros2/pull/49/commits/0b154464e408c3378fb06a4ace098257a51d2a81), but I'm not sure that's what you intended.

gavanderhoorn: I was thinking of rebasing and dropping it. I'll probably do that later.

acbuynak:
I'll load it up and respond today with my progress.

acbuynak:
### Result
Unable to get YAML file loaded on DX200. 
I think I'm missing something small, minimally I'll get a different flash-drive tomorrow to try that.

### Process Notes
0, Start `debug_listener.py`
1, Successfully loaded the `motoros2_alpha1.out` onto the controller (after renaming to remove vers number).
2, As expected, unable to load via `USER DEFINED FILES` or  initialize `SRAM RAM DRIVE`
 
3, Restart into Normal mode
4, Set `[S2C]` parameters.
5, Turn off controllers.

6, Place USB drive (FAT32 formatted) into CN106 (slot 1 and 2 attempted)
7, Restart into Normal mode
8, Errors

**No output to `debug_listener` or `microros` connection throughout process.

### Errors on Startup
```
Alarm 8011 Task#0: MotoROS2 - Multiple Instances [20]
Alarm 8011 Task#0: Must enable ETHERNET function [19]
```

### Debug Note
Only one copy of `motoros2....` is listed in file list.
Checked my `EtherNet/IP(CPU Board)` configuration and verified it is enabled, has an IP address. 
I'm able to ping the controller and get a response. Successfully logged into `telnet` so I'm not sure what the error is from.

**Multiple MotoPlus Programs**
I have one additional MotoPlus program installed on the controller to support a Lincoln welder. Historically, this did not cause an issue with the original motoros1. A flashdrive for the Lincoln MotoPlus program is kept i n one of the two `CN106` ports at all times. Initially I performed above entire process with this additional drive inserted. No success, then repeated process without drive.

gavanderhoorn:
Thanks for testing @acbuynak.

> I have one additional MotoPlus program installed on the controller to support a Lincoln welder. Historically, this did not cause an issue with the original motoros1. A flashdrive for the Lincoln MotoPlus program is kept i n one of the two CN106 ports at all times.

Regarding `8011[20]`: was `motoros2_alpha1.out` also on this drive?

@ted-miller: could it be loading another copy from the drive perhaps?

acbuynak:
> Regarding 8011[20]: was motoros2_alpha1.out also on this drive?

Originally yes. I have since replicated the error by removing all files but yaml config and restarting.

I verified I can clear the error by deleting the `motoros2_alpha.out` file off the controller and rebooting into normal mode. Error reappears when restarting.

I verified my config yaml syntax and get an error for line length for my joint names, but that's all.
```yaml
joint_names:
  - [r1_joint_1_s, r1_joint_2_l, r1_joint_3_u, r1_joint_4_r, r1_joint_5_b, r1_joint_6_t]
  - [r2_joint_1_s, r2_joint_2_l, r2_joint_3_u, r2_joint_4_r, r2_joint_5_b, r2_joint_6_t]
```

ted-miller:
> Regarding 8011[20]: was motoros2_alpha1.out also on this drive?

This shouldn't matter on the DX200. The feature to load a .out file from the internal USB drive was introduced on YRC1000. (Unless YEC updated the DX200 without telling me... which is possible.)

> Alarm 8011 Task#0: MotoROS2 - Multiple Instances [20]

My guess is that the .out file with the long/complicated name got loaded at some point. It will not appear in the MotoPlus file list and cannot be deleted from the pendant. 

You will need to remove the internal CF card from the CPU board and put it in your PC. The .out files are stored in the `APPLICATION` folder on the drive. Just delete any unwanted .out files.

> I have one additional MotoPlus program installed on the controller to support a Lincoln welder.

This application won't conflict with MotoROS2. But, it does consume some CPU tasks, which are limited. If you run into any `8010` alarms regarding task creation, then we'll need to work with Tech Support to modify your `ALL.PRM` file for the controller.

> Alarm 8011 Task#0: Must enable ETHERNET function [19]

This alarm is interesting. I'm glad that you mentioned you have Ethernet/IP enabled. That option-function does some weird stuff with the network parameters. We haven't tested that on the DX2 (or even on the YRC for that matter).

gavanderhoorn:
> > Alarm 8011 Task#0: Must enable ETHERNET function [19]
> 
> This alarm is interesting. I'm glad that you mentioned you have Ethernet/IP enabled. That option-function does some weird stuff with the network parameters. We haven't tested that on the DX2 (or even on the YRC for that matter).

Oh, I hadn't noticed the `Ethernet/IP` part.

That could complicate things for `Ros_mpNICData(..)`.

src/CtrlGroup.c
gavanderhoorn: Done.

gavanderhoorn:
@acbuynak: please also post debug logs in any case. I'd like to check a few things.

ted-miller:
@smith-doug. Here's a foxy version. As noted above, this has **NOT** been tested or validated in any way beyond establishing a connection. 

[motoros2_dx200_foxy_999_alpha3.zip](https://github.com/Yaskawa-Global/motoros2/files/11910242/motoros2_dx200_foxy_999_alpha3.zip)

ted-miller:
I just tested the `motoros2_dx200_9.9.9-alpha1.zip` version again using latest DX200 controller software and the Humble Agent using a docker image on Ubuntu. Everything appears to work fine. 

It appears that my previous issues were due to old controller software and a bad microros agent.

acbuynak:
I am unable to establish a connection. 

The `Must Enable ETHERNET function` Alarm 8011 is still present.

Attachment: [PANELBOX.LOG](https://github.com/Yaskawa-Global/motoros2/files/11918758/PANELBOX.LOG)

Debug Logs: `Debug_listener` does not report anything. Log file is empty and autodeleted by the script.

ted-miller:
I believe that there is a conflict with the Lincoln ArcLinkXT application. That application uses the same method as MotoROS2 to detect the MAC address. I've found in the past that multiple instances of this logic will cause a failure. 

Please try temporarily removing the Lincoln MotoPlus application. You can just copy the .out file from the CF card to your computer so that you can reload it later. (Or I can provide another copy via email.) It will be located in the APPLICATION folder of the CF card.

acbuynak:
**Summary:** 
Removing Lincoln M+ file did not fix ethernet alarm.

**Detail:**
Moved the Lincoln M+ .out file off the CF card. As I removed it from the CF card, I did not 'delete' the M+ application using the teach pendant. Not sure if there are other files created/loaded which would be automatically cleaned if I deleted via TP.

Reinstalled CF card, rebooted DX200's into maintenance mode, verified Lincoln application was not listed and motoros2 remained, rebooted into normal mode.

Same error. The `Must Enable ETHERNET function` Alarm 8011 is still present.

As a curiosity check, I uninstalled (via TP) the `motoros2.out` application, rebooted, and verified error goes away. So error is caused by presence of `motoros2.out` application.

Reinstalled `motoros2.out` application and error reappeared.

ted-miller:
Can you provide a CMOS.BIN backup of your robot? I'd like to try to recreate this issue on my bot.

acbuynak:
@ted-miller Emailed you the CMOS.BIN file for my current configuration.

ted-miller:
@acbuynak: It seems be an issue with the 10.10.10.203 IP address. I got the same alarm using your CMOS. Then I changed to 192.168.1.x and the alarm magically went away. 

I don't have a clue yet why 10.10.10.x is special... but I'll look into it after the holiday.

smith-doug:
I'm unable to get this working on our DX200, most likely due to the software version being too old (2.33).  Should I contact the usual support email to get an update?

[PANELBOX.LOG](https://github.com/Yaskawa-Global/motoros2/files/11962167/PANELBOX.LOG)

ted-miller:
Yes, the usual Technical Support channel for your region will be able to provide a controller update. In your case (US), techsupport@motoman.com or https://www.motoman.com/en-us/service-training/technical-support

src/Ros_mpGetRobotCalibrationData.c
```diff
@@ -0,0 +1,160 @@
...
+
+#elif defined (DX200) || defined (FS100)
+    //The download and parsing of the DAT file is expensive. I only want to
+    //perform it once. So let's cache the data in an array.
+    //
+    //This is an array of POINTERS. We'll only allocate a slot if that calibration
+    //file is configured.
+    static MP_RB_CALIB_DATA* calibrationFiles[MAX_ROBOT_CALIBRATION_FILES];
+    
+    if (file_no == 0) //first time in, let's parse the whole file
```
gavanderhoorn, line 28: Are all 'files' stored in the single `RBCALIB.DAT` file?

ted-miller: Yes. Example DAT file: 
[RBCALIB.DAT.TXT](https://github.com/Yaskawa-Global/motoros2/files/11969549/RBCALIB.DAT.TXT)

src/Ros_mpGetRobotCalibrationData.c
```diff
@@ -0,0 +1,160 @@
...
+        while (TRUE) //unknown file size. this will `break` when done with file
+        {
+            //---------------------------------------------------------------------
+            bRet = FileUtilityFunctions_ReadLine(fd, buffer, SIZEOFBUFFER); //  //RBCALIB x
+
+            if (bRet && strstr(buffer, "//RBCALIB"))
+            {
+                rcutils_split(buffer, ' ', g_motoros2_Allocator, &splitSpace);
+
+                int fileNo = atoi(splitSpace.data[1]);
```
ted-miller, line 69: Pretty sure this should be `fileNo - 1`

src/Ros_mpGetRobotCalibrationData.c
gavanderhoorn: hm. Would it perhaps not be more robust/defensive to check for an entry in `calibrationFiles` instead of relying on the caller starting at `0`?

smith-doug:
We're up and running with the updated controller software and motoros2_dx200_foxy_999_alpha3.zip.  We had an issue with energy saving mode not resuming after hacking/slashing/renaming to make things work.  I can look into it more later, with a debug log.

ted-miller:
@acbuynak: We found the issue. Any IP address whose last octet is `> 128` will result in this error on the DX200. We'll have a fix for you by Monday. (If you want it sooner than that, just change your IP address and it should work.)

src/Ros_mpGetRobotCalibrationData.c
```diff
@@ -15,9 +18,140 @@ LONG Ros_mpGetRobotCalibrationData(ULONG file_no, MP_RB_CALIB_DATA *rData)
...
 #elif defined (DX200) || defined (FS100)
-    #error Not implemented (yet)
+    //The download and parsing of the DAT file is expensive. I only want to
+    //perform it once. So let's cache the data in an array.
+    //
+    //This is an array of POINTERS. We'll only allocate a slot if that calibration
+    //file is configured.
+    static MP_RB_CALIB_DATA* calibrationFiles[MAX_ROBOT_CALIBRATION_FILES];
+    
+    if (file_no == 0) //first time in, let's parse the whole file
```
gavanderhoorn, line 28: Another question: can the calibration be updated without restarting the controller?

If so, the current implementation would make it impossible to (re)load updated data.

gavanderhoorn:
Rebased on `main`.

acbuynak:
> Any IP address whose last octet is > 128 will result in this error on the DX200. We'll have a fix for you by Monday. 

@ted-miller  Strange error, but glad to hear you have a proposed fix. I'll be able to check this on my DX200 next week (7/17). :+1:

gavanderhoorn:
@ted-miller just making sure, when you wrote (in https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1614788223):

> I just tested the `motoros2_dx200_9.9.9-alpha1.zip` version again using latest DX200 controller software and the Humble Agent using a docker image on Ubuntu. Everything appears to work fine.
> 
> It appears that my previous issues were due to old controller software and a bad microros agent.

were you referring to `rclc_node_init_with_options(..)` failing or something else?

src/Ros_mpGetRobotCalibrationData.c
ted-miller: [a97720f](https://github.com/Yaskawa-Global/motoros2/pull/49/commits/a97720f937a20b0ccd076ddfd000c71b039be331)

src/Ros_mpGetRobotCalibrationData.c
ted-miller: This behavior has been modified in [fe22ac3](https://github.com/Yaskawa-Global/motoros2/pull/49/commits/fe22ac3506855c19873064975acddb3856981963). It now gets parsed during initialization.

gavanderhoorn:
@smith-doug wrote:
> We had an issue with energy saving mode not resuming after hacking/slashing/renaming to make things work. I can look into it more later, with a debug log.

would you happen to have had an opportunity to save that debug log? Could you share it with us?

smith-doug:
[robot_sleep_error_clean.txt](https://github.com/Yaskawa-Global/motoros2/files/12043429/robot_sleep_error_clean.txt)


This has the debug log, along with output from a few command line service calls.  The calling code calls start_traj_mode and waits for the result, sleeps for a few seconds (disabled in these runs), then tries to send motion.  

The first run will usually be fine.  If the drives are in power saving and there isn't a long sleep after the start_traj_mode call, it will fail.  Is start_traj_mode supposed to wait until the servos are ready? 

This code was quickly changed from using the ros1 driver through the bridge to native ros2.  I'm going to try modifying the code to check robot_status.motion_possible.  

I did notice while echoing the topic in a terminal with the servos sleeping, after calling start_traj_mode the drives_powered and motion_possible flags will both change to 1, then motion_possible will briefly change to 0 before changing back to 1.

gavanderhoorn:
Rebased again.

Note: CI is green, including the MSBuild CI workflows, but we don't yet make it build DX200 binaries, so it doesn't really say anything about this DX200 branch.

ted-miller:
@acbuynak: [motoros2_dx200_humble-alpha4.zip..zip](https://github.com/Yaskawa-Global/motoros2/files/12071443/motoros2_dx200_humble-alpha4.zip.zip)
Here's a version that I believe will work on your system. As mentioned previously, this has not undergone full motion testing yet. Please exercise great caution if you attempt to move the arm.

gavanderhoorn:
@acbuynak: have you had any luck in trying the binary posted by @ted-miller?

ted-miller:
I just tested the point queuing service and it worked fine.

There were two minor issues that need to resolved though.

1) On DX200, the `S2C541` and `S2C542` parameters default to `1`. So we need to update the installation instructions.

2) `Couldn't save RBCALIB.DAT to DRAM drive. Error (65535)` (with accompanying alarm at startup.) It turns out, the RBCALIB.DAT file doesn't exist in a single-robot system. So, we need to remove this alarm and just log the error.

ted-miller (approved):
I have also tested the FJT server, services over CLI, and topic echo'ing. I think we're in good shape. 

@gavanderhoorn: Can you think of anything I'm missing to test?

ted-miller (approved):
It appears to be parsing the RBCALIB.DAT file correctly and applying it when publishing /tf.

image

gavanderhoorn:
Do we actually want to provide Galactic builds (8e6d5127)? Initially I only added Humble. Foxy was only done on request.

We could add Galactic, but it should probably then also be in the `it-works-until-it-doesn't` category.

> 2\. `Couldn't save RBCALIB.DAT to DRAM drive. Error (65535)` (with accompanying alarm at startup.) It turns out, the RBCALIB.DAT file doesn't exist in a single-robot system. So, we need to remove this alarm and just log the error.

what about multi-group systems where that file really can't be saved to the drive?

Is the alarm not needed in that case? Or can't that happen?

doc/troubleshooting.md
```diff
@@ -910,3 +913,22 @@ ALARM 8014
...
+*Example:*
+
+```text
+ALARM 8015
+ Failed to parse RBCALIB.DAT
+[1]
+```
+
+*Solution:*
+Check the following robot parameters on the teach pendant:
```
gavanderhoorn, line 928:
```suggestion
Check the following robot parameters on the teach pendant and make sure they are set to the values shown here:
```

gavanderhoorn:
I'd like to retain individual commits instead of squash-merging, so nce we're happy with the PR, I'll clean up history a bit.

ted-miller:
> Do we actually want to provide Galactic builds (https://github.com/Yaskawa-Global/motoros2/commit/8e6d5127654abba242c6166fe2a9b313d877f2a4)?

I assumed we would maintain parity with YRC. 

> We could add Galactic, but it should probably then also be in the it-works-until-it-doesn't category

I agree

> > 2. Couldn't save RBCALIB.DAT to DRAM drive. Error (65535) (with accompanying alarm at startup.) It turns out, the RBCALIB.DAT file doesn't exist in a single-robot system. So, we need to remove this alarm and just log the error.
> 
> what about multi-group systems where that file really can't be saved to the drive?
> 
> Is the alarm not needed in that case? Or can't that happen

I don't know how to differentiate between a lack of calibration and a DRAM failure.

I suppose we could add a check of S2C1117 to see if the DRAM drive is enabled. What do you think?

gavanderhoorn:
> I don't know how to differentiate between a lack of calibration and a DRAM failure.

you wrote:

> It turns out, the RBCALIB.DAT file doesn't exist in a single-robot system. So, we need to remove this alarm and just log the error.

Would `num_grp > 1 && error_saving_rbcalib_dat` be a sufficient and correct condition?

> I suppose we could add a check of S2C1117 to see if the DRAM drive is enabled. What do you think?

We could add that, yes.

ted-miller:
> Would num_grp > 1 && error_saving_rbcalib_dat be a sufficient and correct condition?

Nope. It's very possible for the robots to not be calibrated to each other if they don't have the Coordinated Motion option-function enabled. 

> I suppose we could add a check of S2C1117 to see if the DRAM drive is enabled. What do you think?

We should just check it at startup, like the other S2C parameters. Do you want me to update the ParameterExtractionLibrary on this PR or another? (Probably another)

gavanderhoorn:
> > Would num_grp > 1 && error_saving_rbcalib_dat be a sufficient and correct condition?
> 
> Nope. It's very possible for the robots to not be calibrated to each other if they don't have the Coordinated Motion option-function enabled.

hm. Can we check *the Coordinated Motion option-function [is] enabled*?

But even then it's still up to users to calibrate the groups I guess. Not a requirement?

Just trying to figure out whether it'd be worth it to still show the alarm in situations where we can definitely say something is not right -- as that would be helpful for users.

> > I suppose we could add a check of S2C1117 to see if the DRAM drive is enabled. What do you think?
> 
> We should just check it at startup, like the other S2C parameters. Do you want me to update the ParameterExtractionLibrary on this PR or another? (Probably another)

If `S2C1117` is required for DX200 support, we should do it here.

If it's not, a follow-up PR might be better.

gavanderhoorn (commented):
Some things I noticed.

gavanderhoorn:
I'd like to update CI as part of this PR to include builds of the VS project for DX200.

But our CI currently pulls the necessary `libmicroros` distribution `.zip`s [from the 'latest' release](https://github.com/Yaskawa-Global/motoros2/blob/5a2c6ee40d2fc9f7ad4bd77cd8a11829ef73d74d/.github/workflows/ci_msbuild.yaml#L64-L75), where we attach them [as assets](https://github.com/Yaskawa-Global/motoros2/releases/tag/0.1.0).

Without changing how we distribute the `libmicroros` distribution `.zip`s, I believe we have two options:

 1. edit the `0.1.0` release and add three more assets (`.zip`s for DX200 `libmicroros` -- Foxy, Galactic and Humble). Note: we'd edit the GitHub meta-data, no actual commits nor tags would be changed
 2. publish a *prerelease* of `0.1.1`. This would include MotoROS2 binaries for that version, but *not* the binaries for DX200. We would then attach the `.zip`s of `libmicroros` for DX200 to that prerelease

Option 2 seems cleaner (a published release is sort-of supposed to be treated as immutable), and we could even get away with not including a full changelog, etc -- it'd just be a test.

Option 1 would be less work.

Both would *probably* allow extending the CI to include DX200 builds (ie: the changes from this branch).

ted-miller:
> hm. Can we check the Coordinated Motion option-function [is] enabled?
> But even then it's still up to users to calibrate the groups I guess. Not a requirement?

It is not a requirement to calibrate.

> If S2C1117 is required for DX200 support, we should do it here.

It is required. I'll commit a new lib.

ted-miller:
> I'd like to update CI as part of this PR to include builds of the VS project for DX200.
> Option 1 would be less work.

Personally, I'd suggest option 1. The VS project in `0.1.0` isn't setup to build DX200, so there isn't worry about someone mistakenly trying to use them.

gavanderhoorn:
I've just enabled CI for DX200 for all supported ROS 2 versions.

For now I've added the `libmicroros` distribution `.zip`s to the `0.1.0` release. We might want to rethink our distribution strategy though.

ted-miller (approved):
I found a couple new issues while testing. But they aren't specific to the DX200. 

If you're happy with it, let's merge.

acbuynak:
### Test Results (from 3-week old version)
> @acbuynak: [motoros2_dx200_humble-alpha4.zip..zip](https://github.com/Yaskawa-Global/motoros2/files/12071443/motoros2_dx200_humble-alpha4.zip.zip) Here's a version that I believe will work on your system. As mentioned previously, this has not undergone full motion testing yet. Please exercise great caution if you attempt to move the arm.

3 weeks late and may be no longer relevant, but I verified the older M+ file quoted above works for me.

| Test | Result |
| ---- | ---- |
| IP addresses in 192.168.1.# subnet | :heavy_check_mark: |
| IP addresses in 10.10.10.# subnet  | :heavy_check_mark:  |

For each I verified that I could start/stop trajectory mode, write/read IO.

---

### Issue with Lincoln M+ Program Loaded Alongside MicroROS
:warning:  Note this is from the 3-week old M+ file.

I reloaded my Lincoln M+ program alongside motoROS2 and received the following fatal errors from MotoROS2. I understand this is likely out of scope for this PR, but sharing as a heads up.

```
ALARM 8012 MotoROS2 OUT OF MEMORY [1]
ALARM 8011 MotoROS2: Fatal Error [4]
```

I have attached my [debug log here](https://github.com/Yaskawa-Global/motoros2/files/12285180/20230807T165523_motoros2_debug_log.txt)

gavanderhoorn:
Thanks @acbuynak. It's always helpful to get feedback.

@ted-miller wrote:
> I found a couple new issues while testing. But they aren't specific to the DX200.
> 
> If you're happy with it, let's merge.

I believe we need a final review + update of the readme and related documents to add information about DX200-specific steps and errors/alarms.

818569a99ed931d2cf709c534cb4d39b9b91d661 fi adds a couple of alarms, but *Troubleshooting* wasn't updated.

I'm currently focussing on FS100, so if you could do a first pass?

gavanderhoorn:
@ted-miller: b8bc407f44dda6804fc2fba308bda62ce8df1cc3 appears to commit the MotoROS2 logo `.png` again, but under a different filename (`doc/img/MotoROS2_logo.png`). Was that intentional?

We already have `logo.png`.

ted-miller:
No, it was not intentional. But it should be fixed now

README.md
```diff
@@ -233,7 +241,7 @@ If needed, open a new issue on the [Issue tracker](https://github.com/yaskawa-gl
 
 ## Installation
 
-Place the `.out` (main binary), `.yaml` (configuration), and `.dat` (I/O names) files on an external storage device: Compact Flash (CF), Secure Digital (SD), and USB sticks can be used depending on the controller model.
+Place the `.out` (main binary), `.yaml` (configuration), and `.dat` (I/O names) files on an external storage device: Compact Flash (CF), Secure Digital (SD) and USB sticks can be used depending on the controller model.
```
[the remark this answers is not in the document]
gavanderhoorn: [Boiling frog](https://en.wikipedia.org/wiki/Boiling_frog) ;)

I'll check whether we are consistent with this. If we use it elsewhere, I'll revert.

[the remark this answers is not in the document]
ted-miller: > I'll check whether we are consistent with this

I'm guessing it depends on which one of us wrote which sections.

> If we use it elsewhere, I'll revert.

I don't actually care either way :)

gavanderhoorn:
While I can't test it, I agree with @ted-miller this is in quite an OK shape.

I'll spend some time fixing up the commit history and then we should be good to merge.

gavanderhoorn:
Candidate commit history: https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:dx200_support-squashed.

@ted-miller: I believe those commits succinctly show which changes were/are needed to get the YRC version working on DX2 without including all the fixups and experiments we needed to get to this point.

If you agree with the proposed set of commits, I'll force push it here (ie: overwrite the current `dx200_support` branch).

README.md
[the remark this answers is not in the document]
gavanderhoorn: I've reverted this one in the [squashed history](https://github.com/Yaskawa-Global/motoros2/pull/49#issuecomment-1676731605). Oxford commas appear throughout the text.

ted-miller:
Your branch looks good. Go ahead and push/merge it.

gavanderhoorn:
Going to need a final review (again).

We could then merge this, deal with some of the other outstanding PRs and decide whether that'd be sufficient for `0.1.1`.

We could also decide to go for `0.2.0`, seeing as DX200 support is a bit of a milestone (also: `0.2.0` matches nicely with DX`2`00, but that's minor).

gavanderhoorn:
Thanks for the work @ted-miller 👍 

Thanks for testing early builds and putting up with our mistakes @acbuynak and @smith-doug 👍

PR #61, opened 2023-06-15
Adding a timestamp to debug messages

### LINKED ISSUE #38: Add timestamp to debug udp broadcasts
Debug UDP broadcasts should include the time they were sent:

https://github.com/Yaskawa-Global/motoros2/blob/08329e5a1646315bffb5effdb40857eaea343523/src/Debug.c#L36-L54

SejalBehere:
Changes were made to Debug.c to include a timestamp (Format - Day  Y:M:D  H:M:S.ms) in front of the debug messages being printed. Change was made to MotoROS_PlatformLib.h to add the function localtime_r. These changes were "Fixes #38" to resolve the issue "Add timestamp to debug udp broadcasts" .


src/Debug.c
```diff
@@ -36,6 +36,7 @@ void Ros_Debug_Init()
 void Ros_Debug_BroadcastMsg(char* fmt, ...)
 {
     char str[MAX_DEBUG_MESSAGE_SIZE];
+    char Formatted_Time[300];
```
gavanderhoorn, line 39: Please also use [snake_case](https://en.wikipedia.org/wiki/Snake_case) for variable names.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
 
     if (ros_DebugSocket == -1)
         Ros_Debug_Init();
+    
+   // Timestamp
+
```
gavanderhoorn, line 52:
```suggestion
```

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+    struct timeval tv;
+
+
+    time(&DEBUG_MSG_TIMESTAMP);
+
+    localtime_r(&DEBUG_MSG_TIMESTAMP, &ts);
+
+    strftime(Formatted_Time, sizeof(Formatted_Time), "%a %Y-%m-%d %H:%M:%S", &ts);
+
+    gettimeofday(&tv, NULL);
```
gavanderhoorn, line 64: Just wondering: wouldn't this return a *different* time point (compared to the `DEBUG_MSG_TIMESTAMP`)?

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+    size_t Timestamp_Length = strlen(Formatted_Time);
+    size_t Debug_Message_Length = strlen(str);
+
+    // Pre-pending the timestamp to the debug message 
+
+    if (Timestamp_Length + Debug_Message_Length + 1 < MAX_DEBUG_MESSAGE_SIZE)
+    {
+        memmove(str + Timestamp_Length, str, Debug_Message_Length + 1); // Move existing contents of str buffer to the end by Timestamp_Length to make space for the timestamp and avoiding overwriting the debug message during the move 
+        memcpy(str, Formatted_Time, Timestamp_Length); // Copy the timestamp stored in Formatted_Timestamp buffer to the beginning of str buffer        
+    }
```
gavanderhoorn, line 78: Is there anything we can do if the combined string is too long?

Maybe truncate with an ellipsis?

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+
+    strftime(Formatted_Time, sizeof(Formatted_Time), "%a %Y-%m-%d %H:%M:%S", &ts);
+
+    gettimeofday(&tv, NULL);
+    long msec = tv.tv_usec / 1000;
+
+    snprintf(Formatted_Time, sizeof(Formatted_Time), "%s.%03ld", Formatted_Time, msec);
+
+    size_t Timestamp_Length = strlen(Formatted_Time);
+    size_t Debug_Message_Length = strlen(str);
```
gavanderhoorn, line 70: Consider using `strnlen(..)`.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+
+    // Pre-pending the timestamp to the debug message 
+
+    if (Timestamp_Length + Debug_Message_Length + 1 < MAX_DEBUG_MESSAGE_SIZE)
```
gavanderhoorn, line 74: Why the `+1`?

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+    {
+        memmove(str + Timestamp_Length, str, Debug_Message_Length + 1); // Move existing contents of str buffer to the end by Timestamp_Length to make space for the timestamp and avoiding overwriting the debug message during the move 
+        memcpy(str, Formatted_Time, Timestamp_Length); // Copy the timestamp stored in Formatted_Timestamp buffer to the beginning of str buffer
```
gavanderhoorn, line 77:
```suggestion
        // Move existing contents of str buffer to the end by Timestamp_Length to make space for the timestamp and avoiding overwriting the debug message during the move
        memmove(str + Timestamp_Length, str, Debug_Message_Length + 1);
        // Copy the timestamp stored in Formatted_Timestamp buffer to the beginning of str buffer
        memcpy(str, Formatted_Time, Timestamp_Length);
```

gavanderhoorn (changes requested):
Thanks for the PR 👍 

I've left a couple in-line comments.

Besides those: IIUC, this returns time in 'local time'. Many controllers will not be configured with the correct timezone (if that is even possible), and will likely not have NTP setup.

This means stamps prefixed to debug output would not necessarily show the correct time, when compared to the time of the receiving systems.

In a previous version I believe you used the time from the host, synced using the micro-ROS Agent. That would seem to avoid the timezone issue.

Did that not work correctly?

ted-miller:
> In a previous version I believe you used the time from the host, synced using the micro-ROS Agent. That would seem to avoid the timezone issue.

gavanderhoorn is right, we have used another time function that is synchronized with the PC running the microROS Agent. 

https://github.com/Yaskawa-Global/motoros2/blob/9f1c599549ca89fd4323cf47fbd3101dadf2d3d8/src/PositionMonitor.c#L404-L406

`rmw_uros_epoch_nanos` gets nanoseconds. Then `Ros_Nanos_To_Time_Msg` will break that into seconds and remaining nanoseconds.

gavanderhoorn:
I specifically referred to https://github.com/Yaskawa-Global/motoros2/compare/Yaskawa-Global:9f1c599...SejalBehere:7f163ac.

That seems to do what you also suggest @ted-miller, but there must be a reason this didn't make it into this PR. Hence my question.

SejalBehere:
> 
In a previous version I did use time from the host, synced using the micro-ROS Agent. I was able to get the timestamp in the format of H:M:S.ms but was not able to add the day to it without adding the whole of what I have used now as well. 

I will make some more changes to the previous version and get the current timestamp format using the host time.

src/Debug.c
SejalBehere: Will make sure all the variables are in that format now and for future use as well.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
 
     if (ros_DebugSocket == -1)
         Ros_Debug_Init();
+    
+   // Timestamp
+
+    time_t DEBUG_MSG_TIMESTAMP;
```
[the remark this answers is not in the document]
SejalBehere: I will make it into lowercase.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+    struct tm ts;
+    struct timeval tv;
+
```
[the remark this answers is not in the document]
SejalBehere: Will remove the empty lines.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+
+    time(&DEBUG_MSG_TIMESTAMP);
+
+    localtime_r(&DEBUG_MSG_TIMESTAMP, &ts);
+
+    strftime(Formatted_Time, sizeof(Formatted_Time), "%a %Y-%m-%d %H:%M:%S", &ts);
```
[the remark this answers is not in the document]
SejalBehere: Will keep that in mind and make the change

src/Debug.c
SejalBehere: I just realized it now that you said that. Thanks for pointing that out. Will change it out and use the host time synced using the micro-ROS Agent instead.

src/Debug.c
```diff
@@ -46,6 +47,36 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+
+    gettimeofday(&tv, NULL);
+    long msec = tv.tv_usec / 1000;
+
+    snprintf(Formatted_Time, sizeof(Formatted_Time), "%s.%03ld", Formatted_Time, msec);
```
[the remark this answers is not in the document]
SejalBehere: Okay, will keep that in mind.

src/Debug.c
SejalBehere: Will truncate it with an ellipsis or some other way if that doesn't work.

src/Debug.c
SejalBehere: Okay, will use that instead.

src/Debug.c
SejalBehere: To account for the null character.

src/Debug.c
```diff
@@ -47,37 +47,29 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
-    // Pre-pending the timestamp to the debug message 
-
-    if (Timestamp_Length + Debug_Message_Length + 1 < MAX_DEBUG_MESSAGE_SIZE)
+    // Timestamp
+    struct tm synced_time;
+    int64_t nanosecs = rmw_uros_epoch_nanos();
+    builtin_interfaces__msg__Time debug_msg_timestamp;
+
+    Ros_Nanos_To_Time_Msg(nanosecs, &debug_msg_timestamp);
+    time(&debug_msg_timestamp);
```
ted-miller, line 57: IIUC, the `time` function will put the current time into `debug_msg_timestamp`. But, this is not needed because the current timestamp was already obtained above with `rmw_uros_epoch_nanos`. I think that this line can be deleted.

SejalBehere: I did try by removing that line but what is happening is that for the fist few messages, before the message "Attempting to connect to the micro-ROS agent", the timestamp being showed is the epoch time (THU 1970-01-01 00:00:00.000) and it is only after this message that it shows the current timestamp.

ted-miller: I see. Then I would suggest making the time source conditional on whether the agent is connected.

```
if (g_Ros_Communication_AgentIsConnected)
{
    //get synchronized time from the agent
    int64_t nanosecs = rmw_uros_epoch_nanos();
    Ros_Nanos_To_Time_Msg(nanosecs, &debug_msg_timestamp);
}
else
{
    //rmw_uros_epoch_nanos cannot sync with agent because it's not connected
    time(&debug_msg_timestamp);
}
```

SejalBehere: Okay, I will replace the time function with the conditional

gavanderhoorn: I'm not sure you can pass an instance of `builtin_interfaces__msg__Time` to `time(..)`.

src/Debug.c
SejalBehere: I'm sorry to have missed the removal of this blank line. I have made the changes and removed it.

gavanderhoorn:
@SejalBehere: if this is ready for another review please either request one or comment.

SejalBehere:
@gavanderhoorn This is ready for another review.

src/Debug.c
```diff
@@ -46,7 +47,39 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
 
     if (ros_DebugSocket == -1)
         Ros_Debug_Init();
-
+    // Timestamp
+    //The timestamp for the message "Found Micro-Ros PC Agent" will be the epoch time (THU 1970-01-01 00:00:00.000) as the global flags 
+    //are set to indicate that the Micro-Ros PC Agent is connected but the first sync of the host time using the micro-ROS agent is yet to occur
+    struct tm synced_time;
+    struct timeval tv;
+    int64_t nanosecs = rmw_uros_epoch_nanos();
```
ted-miller, line 55: This is declared a second time below inside the IF statement. Please remove this instance so that there is only call to `rmw_uros_epoch_nanos`. This will also remove the duplicated declaration.

SejalBehere: I will remove that declaration of rmw_uros_epoch_nanos.

src/Debug.c
```diff
@@ -46,7 +47,39 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+    builtin_interfaces__msg__Time debug_msg_timestamp;
+    if (g_Ros_Communication_AgentIsConnected)
+    {
+        //get synchronized time from the agent
+        int64_t nanosecs = rmw_uros_epoch_nanos(); 
+        Ros_Nanos_To_Time_Msg(nanosecs, &debug_msg_timestamp);
+        strftime(timestamp, FORMATTED_TIME_SIZE, "%a %Y-%m-%d %H:%M:%S", localtime_r(&debug_msg_timestamp.sec, &synced_time));
```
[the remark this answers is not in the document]
SejalBehere: I forgot to define it on GitHub. Will do that.

src/Debug.c
```diff
@@ -46,7 +47,39 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+        snprintf(timestamp + strlen(timestamp), FORMATTED_TIME_SIZE - strlen(timestamp), ".%03d", tv.tv_usec / 1000);
+    }
+    // Pre - pending the timestamp to the debug message
+    size_t timestamp_length = strnlen(timestamp, FORMATTED_TIME_SIZE);
+    size_t debug_message_length = strnlen(str, MAX_DEBUG_MESSAGE_SIZE);
+    if (timestamp_length + debug_message_length + 1 < MAX_DEBUG_MESSAGE_SIZE)
+    {
+        // Move existing contents of str buffer to the end by Timestamp_Length to make space 
+        //for the timestamp and avoiding overwriting the debug message during the move 
+        memmove(str + timestamp_length, str, debug_message_length + 1);
```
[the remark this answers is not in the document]
SejalBehere: Okay, I will look into that and alter it accordingly.

SejalBehere:
I have made the changes to address the points raised in the above review. Also, I have removed the +1 as like @ted-miller said str buffer is getting zeroed at the top and hence there's no need for considering the null terminator.

src/Debug.c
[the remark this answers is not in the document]
ted-miller: It doesn't appear that `Formatted_Time` is used in the code below. Looks like a leftover artifact from debugging. I recommend removing this unused variable.
(If you do want to use it for something, be sure to replace the `300` with `FORMATTED_TIME_SIZE`.)

ted-miller (changes requested):
The only other question I have is to confirm whether you have a tested this on a live controller. If so, please post an example copy/paste of the output so we can see the final product.
Thanks!

ted-miller:
Also, I owe you an updated version of `MotoROS_PlatformLib.h`

SejalBehere:
I have made the changes to the branch. I did test this on a live controller and the output was obtained in the following format:


[1687983037.99751401] [192.168.1.31:58649]: WED 2023-06-28 16:10:25.768 Using ROS domain ID: 0
[1687983037.99761701] [192.168.1.31:58649]: WED 2023-06-28 16:10:25.768 Using client key: 0x73B9D4ED
[1687983037.99765611] [192.168.1.31:58649]: WED 2023-06-28 16:10:25.768 Attempting to connect to Micro-ROS PC Agent (at udp://192.168.1.50:8888)
[1687983039.01968718] [192.168.1.31:58649]: THU 1970-01-01 00:00:00.000 Found Micro-ROS PC Agent
[1687983039.02444792] [192.168.1.31:58649]: WED 2023-06-28 16:10:26.795 rclc_support_init_with_options = 0
[1687983039.02451229] [192.168.1.31:58649]: WED 2023-06-28 16:10:26.795 remap_rules str: ''
[1687983039.02464890] [192.168.1.31:58649]: WED 2023-06-28 16:10:26.795 len(remap_rules str): 0
[1687983039.02472425] [192.168.1.31:58649]: WED 2023-06-28 16:10:26.795 num parsed remap rules: 0

ted-miller (dismissed):

ted-miller:
Looks good to me. Thanks for the contribution!

I'll work on resolving the conflicts this afternoon and get this merged.

src/MotoROS_PlatformLib.h
```diff
@@ -55,8 +55,23 @@ extern ULONG tickGet();
...
+
+
+// Attempts to determine whether the specific link is UP (ie: cable is connected).
+// IFF return value is 'OK', 'is_up' will reflect link state. In all other cases,
+// 'is_up' may be left uninitialised.
+#define ROS_USERLAN_STATE_INVALID_IF_NO -10
+#define ROS_USERLAN_STATE_NO_SUCH_IF -11
+#define ROS_USERLAN_STATE_GET_STATE_ERROR -12
+#define ROS_USERLAN_STATE_GENERIC_FAILURE -13
+extern STATUS Ros_UserLan_IsLinkUp(USHORT if_no, BOOL* const is_up /*out*/);
```
gavanderhoorn, line 74: These lines should not be part of this PR. They are already part of #49.

gavanderhoorn (dismissed):
See my comment.

src/MotoROS_PlatformLib.h
ted-miller: I just gave SejalBehere the latest version of the file which included the update for `localtime_r`.

Since the history of this file is being maintained in another repo, I'm not sure how you want to handle it in this repo. Do you just want the `localtime_r` change in here? What about the `MOTOROS_PLATFORM_LIB_VERSION`?

gavanderhoorn: I'd suggest to remove everything *not* related to `localtime_r(..)`. After merging this PR, I'll rebase #49 and sort out any conflicts.

It's a bit awkward with the binary involved, but we'll figure it out.

ted-miller:
I can confirm that YRC does raise a warning. It's wanting a `time_t` instead of `int32_t`. I guess a simple explicit cast should resolve it.

ted-miller:
@gavanderhoorn, I'm also getting a warning for `gettimeofday`. Is that not defined somewhere? (It seems to work/run...)

PR #63, opened 2023-06-16
Add missing enum->str mappings from `Init_Trajectory_Status`

### LINKED ISSUE #7: "generic failure" returned by action server for "final pt must have accel 0" error
After re-enabling the `last-pt-in-a-trajectory-must-have-0-acceleration` check [here](https://github.com/Yaskawa-Global/motoros2/blob/03e248d6d0c420ba00f2cc00a3425fc71c449317/src/MotionControl.c#L327-L336), I submitted a trajectory generated by MoveIt that apparently did not have the final point with an acceleration `== 0`.

MotoROS2 debug log showed:

```
The final point in a trajectory must specify a target acceleration of '0'.
```

but the ROS 2 / MoveIt log only shows:

```
Controller 'follow_joint_trajectory' failed with an unknown error: Trajectory initialization failed. Generic failure."
```

It would be nice(r) to provide the more descriptive error message, like we do for other errors.

gavanderhoorn:
Fixes #7.

### DIFF
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -252,7 +252,8 @@ rcl_ret_t Ros_ActionServer_FJT_Goal_Received(rclc_action_goal_handle_t* goal_han
         if (!bSizeOk)
         {
             motomanErrorCode = INIT_TRAJ_TOO_BIG;
-            rosidl_runtime_c__String__assign(&fjt_result_response.result.error_string, "Trajectory contains too many points (Not enough memory).");
+            rosidl_runtime_c__String__assign(&fjt_result_response.result.error_string,
+                "Trajectory contains too many points (Not enough memory).");
         }
         else if (!bMotionReady)
         {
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
I'm wondering whether we should refactor this and move it to a function in `ErrorHandling` instead. The `enum` keeps growing, and with it the nr of lines mapping them to a `char[]`.

We might also want to store these strings in a ROS `.msg`, like we do with the `MotionNotReadyCode` messages.

ted-miller (approved):
> We might also want to store these strings in a ROS .msg, like we do with the MotionNotReadyCode messages.

Probably not a bad idea, for the sake of consistency.

gavanderhoorn:
> > We might also want to store these strings in a ROS .msg, like we do with the MotionNotReadyCode messages.
> 
> Probably not a bad idea, for the sake of consistency.

Tracking in #67.

PR #70, opened 2023-06-19
doc: add descriptions for missing alarm codes to Troubleshooting

gavanderhoorn:
bc246e4d adds a description for `8011[55]` which will be raised if a problem with the custom `joint_names` setting is encountered. This is not new functionality, but apparently we hadn't yet documented it.

c3296373 removes a superfluous `CONFIGURATION_SUBCODE` entry, which should've been removed as part of #27 (the new 'userlan monitor' functionality).

The other three commits document possible alarms raised by code from #27.


doc/troubleshooting.md
```diff
@@ -732,6 +759,95 @@ Example format: `remap_rules: "joint_states:=my_joint_states read_single_io:=io/
...
+[14]
+```
+
+*Solution:*
+Because `userlan_monitor_enabled` was set to `true` but no value was supplied for `userlan_monitor_port`, MotoROS2 attempted to auto-detect the LAN port to monitor.
+This auto-detection failed, and as a result MotoROS2 has disabled LAN port monitoring for this session.
+
+To rule out a transient failure, reboot the controller.
+
+If the alarm is raised again, and if auto-detection is not needed or desired, make sure `userlan_monitor_port` is not commented out (ie: does not have a `#` at the start of the line) and set it to an appropriate value.
```
ted-miller, line 816: > (ie: does not have a `#` at the start of the line)

We may need to emphasize this point at the top of the instructions. I've seen two or three instances of people not removing the `#` on the agent IP address.

gavanderhoorn: I don't believe there would be a way to draw attention to that in any other way than repeating the same thing everywhere.

I'd rather add it on a case-by-case basis.

ted-miller (approved):

PR #82, opened 2023-07-10
debug: use Ros wrapper of strnlen(..) instead

gavanderhoorn:
As per subject.

`strnlen(..)` is not supported by M+ SDK:

> `warning: implicit declaration of function 'strnlen'`

Introduced by #61.


ted-miller:
Hm. I didn't think to check for build warnings. Just assumed that if it ran/printed correctly that it was ok. 

I will review more closely going forward.

ted-miller (approved):

PR #83, opened 2023-07-10
Explicit casting to resolve warning

ted-miller:
Resolves compiler warning introduced in #61


gavanderhoorn:
@ted-miller: could you please rebase this on `yaskawa-global:main`? Just so we can get CI to run the MSBuild workflow for this?

gavanderhoorn (approved):
With green CI, 👍

gavanderhoorn:
Apologies. This runs into the same issue with repository secrets as I had in #89 and #86 :(.

I need to come up with a work-around for that.

We can either merge without CI, or you'd have to re-submit the PR from a branch on this repository.

PR #86, opened 2023-07-12
Introduce MSBuild based GitHub Actions CI builds

gavanderhoorn:
As per subject.

This uses a `windows-2022` runner (which comes with VS22), the M+ SDK and MSBuild to build the `MotoROS2.sln` project as a regular user with the SDK would.

The intent is to catch build issues early and have an additional layer of build-testing which should help us prevent inadvertently introducing problems which only occur on other/clean build environments.

See [gavanderhoorn/motoros2/actions/runs/5530961659](https://github.com/gavanderhoorn/motoros2/actions/runs/5530961659) for a test build (on my fork).

### DIFF
--- a/.github/matchers/mpbuilder.json
+++ b/.github/matchers/mpbuilder.json
@@ -1,0 +1,16 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "mpBuilder",
+      "pattern": [
+        {
+          "regexp": "(\\w+\\.\\w+)\\((\\d+)\\):\\s+(\\w+)\\s+:\\s+(.*)\\s+\\[\\S+\\]$",
+          "file": 1,
+          "line": 2,
+          "severity": 3,
+          "message": 4
+        }
+      ]
+    }
+  ]
+}
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: 2023, Yaskawa America, Inc.
+# SPDX-FileCopyrightText: 2023, Delft University of Technology
+#
+# SPDX-License-Identifier: CC0-1.0
+
+name: "CI: build VS proj"
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  msbuild:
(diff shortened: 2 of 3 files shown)


gavanderhoorn:
It looks like the build failed. I'm not sure why yet, as the test env does successfully complete (see the run linked in the OP).

I'll investigate and update when I know something.

ted-miller:
> exit code: 3 something from mpBuilder?

No. I believe that is from the `The system cannot find the path specified.` message.

I think that this command is wrong:
```
unzip -q "/c/build/1684764548_libmicroros_humble-20221102-dbg_yrc1000.zip" \
    -d /c/build/motoros2/libmicroros_yrc1000_humble
```
I think it should be `-d /c/build/libmicroros_yrc1000_humble`

ted-miller:
> I think it should be -d /c/build/libmicroros_yrc1000_humble

Nope. I take that back

ted-miller:
```
C:\msys64\usr\bin\wget.exe -q -O C:/build/mpsdk_ci.7z ""
  "C:\Program Files\7-Zip\7z.exe" x C:/build/mpsdk_ci.7z -oC:/build/
  del /q C:\build\mpsdk_ci.7z
  shell: C:\Windows\system[3](https://github.com/Yaskawa-Global/motoros2/actions/runs/5532899977/jobs/10095630017?pr=86#step:9:3)2\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
ERROR: C:\build\mpsdk_ci.[7](https://github.com/Yaskawa-Global/motoros2/actions/runs/5532899977/jobs/10095630017?pr=86#step:9:8)z

Cannot open the file as archive
```

Maybe missing a space on the second line after `-o`???

gavanderhoorn:
No, I was being dense: I forgot to configure the required repository secret ..

gavanderhoorn:
Ok, so this will never work, as the workflow is trying to access a repository secret. But I opened the PR from my fork, which means from an "external account" and those don't get access to any secrets (for security reasons).

I'll see whether opening the same PR from a branch on this repository works. If not, I suggest we merge this in based on the fact I've got successful runs of the workflow on my fork.

gavanderhoorn:
closing in favour of #87.

PR #87, opened 2023-07-12
Introduce MSBuild based GitHub Actions CI builds (attempt 2)

gavanderhoorn:
Replacement for #86, but now from a branch on this repository instead of my fork -- to see whether the repository secrets are accessible this way.

### DIFF
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -19,10 +19,8 @@ jobs:
       # controllers and for which ROS 2 versions builds fail / succeed)
       fail-fast: false
       matrix:
-        # controller: [yrc1000, yrc1000u]
-        # ros2_codename: [humble, galactic, foxy]
-        controller: [yrc1000]
-        ros2_codename: [humble]
+        controller: [yrc1000, yrc1000u]
+        ros2_codename: [humble, galactic, foxy]
 
     steps:
     - name: Uppercase controller identifier
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: 2023, Yaskawa America, Inc.
+# SPDX-FileCopyrightText: 2023, Delft University of Technology
+#
+# SPDX-License-Identifier: CC0-1.0
+
+name: "CI: build VS proj"
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  msbuild:
+    runs-on: windows-2022
+
+    strategy:
+      # keep jobs running even if one fails (we want to know on which
+      # controllers and for which ROS 2 versions builds fail / succeed)
+      fail-fast: false
+      matrix:
+        # controller: [yrc1000, yrc1000u]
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Seems like access to the repository secret was indeed the main issue.

.github/workflows/ci_msbuild.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+    strategy:
+      # keep jobs running even if one fails (we want to know on which
+      # controllers and for which ROS 2 versions builds fail / succeed)
+      fail-fast: false
+      matrix:
+        # controller: [yrc1000, yrc1000u]
+        # ros2_codename: [humble, galactic, foxy]
+        controller: [yrc1000]
+        ros2_codename: [humble]
```
gavanderhoorn, line 25: We could extend this to include all the controllers we currently support, for all the ROS 2 versions we currently support.

```suggestion
        controller: [yrc1000, yrc1000u]
        ros2_codename: [humble, galactic, foxy]
```

.github/workflows/ci_msbuild.yaml
```diff
@@ -0,0 +1,110 @@
...
+      run: |
+        C:\msys64\usr\bin\wget.exe -q -O C:/build/mpsdk_ci.7z "${{ secrets.MPSDK_DOWNLOAD_URL }}"
+        "C:\Program Files\7-Zip\7z.exe" x C:/build/mpsdk_ci.7z -oC:/build/
+        del /q C:\build\mpsdk_ci.7z
+    - name: Add M+ SDK to PATH
+      shell: bash
+      run: echo "C:/build/mpsdk" >> $GITHUB_PATH
+
+    - name: Register mpBuilder problem matcher
+      run: echo "::add-matcher::motoros2/.github/matchers/mpbuilder.json"
```
gavanderhoorn, line 95: The problem matcher actually works, and should result in warnings and/or errors in the output of `mpBuilder` to be converted into annotations on the files contributed in a PR.

However, `mpBuilder` currently doesn't output the full path to files, which makes it impossible for GitHub to match the errors to the respective files in the repository. The best we can do right now is to have the annotations appear on the summary page of a specific run.

ted-miller: It still shows the file and line number. I think we'll be fine.
(Though I can work on a future enhancement for M+SDK at some point.)

image

ted-miller (approved):

.github/workflows/ci_msbuild.yaml
gavanderhoorn: > It still shows the file and line number.

yes, but the links in the summary only take you to the log.

With the full path, Github would be able to map it to the file in the repository which would:

 - allow you to click on the warning/error and be taken to the actual file at the line causing the issue
 - make GitHub show in-line annotations for all warnings/errors reported in the log

gavanderhoorn:
One thing left to decide: should we add `-warnAsError` as option to the MSBuild invocation?

Currently the build succeeds, but there are warnings, which should probably be treated as errors.

It would mean no build would succeed with any warnings though, which is probably a good thing, but could be annoying.

gavanderhoorn:
Thanks for the review @ted-miller.

PR #88, opened 2023-07-13
ci: only run msbuild if needed

gavanderhoorn:
As per subject.

See the commit comment for rationale.

Summarising: this only starts the (full) build of MotoROS2 if a PR actually changes files that would be part of such builds. No point in starting 6 builds if only one `.md` file (ie: documentation) was changed.

### DIFF
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -11,8 +11,26 @@ on:
   workflow_dispatch:
 
 jobs:
+  msbuild_filter:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
The checks for this PR already showcase the functionality. Notice how the `msbuild` job was *skipped*, as this PR only changes a workflow file.

We might want to consider adding the workflow `.yaml` itself to the list of files to check in the `needs_msbuild` filter, as it might actually skip builds even if we change the workflow itself (which would make testing changes to the workflow difficult).

ted-miller (approved):

Issue #60, opened 2023-06-14
Reuse of error (sub)codes


ted-miller:
There are multiple instances of the `SUBCODE_FAIL_MP_NICDATA` error code in different source files. Although they both essentially have the same meaning, it is impossible to know exactly which line of code raised the error. This makes troubleshooting more difficult.

We should go through and make sure that no error code is repeated anywhere in the code.

`SUBCODE_FAIL_MP_NICDATA` instances:

https://github.com/Yaskawa-Global/motoros2/blob/9f1c599549ca89fd4323cf47fbd3101dadf2d3d8/src/CommunicationExecutor.c#L67

and

https://github.com/Yaskawa-Global/motoros2/blob/9f1c599549ca89fd4323cf47fbd3101dadf2d3d8/src/ConfigFile.c#L134

gavanderhoorn:
I've got something for this in https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:use_updated_get_mac_addr.

PR #89, opened 2023-07-13
Update PlatformLib to `v0.2.10`

gavanderhoorn:
As per subject.

Only build-tested (but we'll see what CI thinks of it), although we have tested this version of the platform lib as part of #49.

0ec05f5ea8db598e4a855da72dda4a659fec165a adds the changes necessary to use the new version of `Ros_GetMacAddress(..)` (which now requires an argument to indicate for *which* interface the MAC should be retrieved).

e202c7f63076c0026bbe6ecfb68635dcd6f634ec addresses #60, at least for `SUBCODE_FAIL_MP_NICDATA` (it was used 4 times in total).


gavanderhoorn:
Closing in favour of #90.

Issue #58, opened 2023-06-09
Debug log always broadcast on `LAN2`


gavanderhoorn:
With `motoros2_config.yaml` setup to use `LAN3` on a YRC1000 (or at least: using an IP address in the range configured on `LAN3`) it appears debug log messages are only sent to `LAN2`.

Without `LAN2` having a connection, no messages will be received.

gavanderhoorn:
We have it hard-coded to always only use `LAN2` (ie: `ROS_USER_LAN1`):

https://github.com/Yaskawa-Global/motoros2/blob/8c1a524e8cad8e45e865454bb01678cabc416e38/src/Debug.c#L18-L34

Note: this could be *by-design*, but I though to report it anyway.

ted-miller:
> Note: this could be by-design, but I though to report it anyway.

I don't think this is by-design. Just an oversight.

jimmy-mcelwain:
So would the preferred solution be to publish to all enabled LAN ports simultaneously? Or maybe to add an option in motoros2_config.yaml for the user to configure which LAN port to send the message?

ted-miller:
I vote for the first option

gavanderhoorn:
I'd actually suggest implementing the configuration option, as that would cover multiple use-cases:

1. not configuring anything: broadcast on all active interfaces
2. configuring a specific interface: broadcast on that
3. disable broadcasting completely

@ted-miller's prefers option 1.

Option 2 would be my choice.

Option 3 is something which I believe we've discussed before and would be nice to have: in some cases it would/could be preferable to disable the debug log completely -- for security reasons fi -- and this would be one way to implement that.

We'd have to come up with a nice way to encode for "don't use any of the available interfaces".

Or perhaps we should not combine these, and have two options: global (en|dis)able of debug logging and one which configures which interface to use.

Similar to the user lan monitoring: this would only really make sense for controllers with multiple interfaces. On anything else, the setting will be ignored. Except disabling logging completely of course.

jimmy-mcelwain:
Resolved by #309 and #321

Issue #57, opened 2023-06-09
Incorrect MAC used if LAN3 is used instead of LAN2


gavanderhoorn:
Ran into this while testing #27.

With `motoros2_config.yaml` setup to use `LAN3` on a YRC1000 (or at least: using an IP address in the range configured on `LAN3`), the following are seemingly still based on the MAC of `LAN2`:

 - micro-ROS client key
 - generated MotoROS2 node name

As a user, I would expect anything based on MAC address to take that address from the NIC that is used to connect to the Agent.

gavanderhoorn:
Looks like [this](https://github.com/Yaskawa-Global/motoros2/blob/8c1a524e8cad8e45e865454bb01678cabc416e38/src/CommunicationExecutor.c#L62-L68) and [this](https://github.com/Yaskawa-Global/motoros2/blob/8c1a524e8cad8e45e865454bb01678cabc416e38/src/ConfigFile.c#L130-L137) should first figure out which NIC to retrieve the MAC from.

jimmy-mcelwain:
I am working to address this and #58. I have a [branch](https://github.com/Yaskawa-Global/motoros2/tree/LanFix) where I pushed some code. There are several problems that stem from my misunderstandings the first time I tried to solve this. But I have a couple of questions.

First, for the following [code](https://github.com/Yaskawa-Global/motoros2/blob/5820d8d06189af91569cf7020e552f23ee417d73/src/CommunicationExecutor.c#L62-L86). The TODO states `get MAC only from interface which is used for ROS traffic`. I originally thought that this was saying that the NIC for the `Ros_GetMacAddress()` function should always be the same as the NIC that is used for `userlan_monitor_port`, but it doesn't look that way to me anymore. Are `userlan_monitor_enabled` and `userlan_monitor_port` only used to check liveliness, and if so, why would they be different from the port that is used to communicate with the agent? It would be easier if they were the same, but I assume there is some good reason to allow the monitoring and the communication to happen over 2 different ports. If there is, then I plan on "saving" the port that is on the same subnet as the agent during `Ros_ConfigFile_ValidateCriticalSettings()` and getting the MAC address from the interface on that port for this call. If I "save" it there, then I can also use that value for auto-detection if `userlan_monitor_port` is not specified, and I could use it for 

Also, with regards to debugging, is `Ros_Debug_BroadcastMsg()` meant to send data over the `userlan_monitor_port`, or over the port that is used to communicate with the agent? 

[...]

PR #90, opened 2023-07-13
Update PlatformLib to `v0.2.10` (resubmit for CI)

gavanderhoorn:
(This is a replacement for #89 to get CI working (secrets don't work for workflows triggered by events from forks))

As per subject.

Only build-tested (but we'll see what CI thinks of it), although we have tested ~~this~~ *similar* versions of the platform lib as part of #49.

0ec05f5ea8db598e4a855da72dda4a659fec165a adds the changes necessary to use the new version of `Ros_GetMacAddress(..)` (which now requires an argument to indicate for *which* interface the MAC should be retrieved).

e202c7f63076c0026bbe6ecfb68635dcd6f634ec addresses #60, at least for `SUBCODE_FAIL_MP_NICDATA` (it was used 4 times in total).


src/CommunicationExecutor.c
```diff
@@ -61,11 +61,29 @@ void Ros_Communication_ConnectToAgent()
...
-    motoRosAssert_withMsg(
-        Ros_GetMacAddress(macId) == OK,
-        SUBCODE_FAIL_MP_NICDATA,
-        "Must enable ETHERNET function");
+    STATUS status = Ros_GetMacAddress(ROS_USER_LAN1, macId);
+    if (status != OK)
+    {
+        Ros_Debug_BroadcastMsg("%s: Ros_GetMacAddress: iface: %d; error: %d",
+            __func__, ROS_USER_LAN1, status);
+        motoRosAssert_withMsg(FALSE, SUBCODE_FAIL_MP_NICDATA_INIT0,
```
ted-miller, line 72: On YRC, this won't allow the code to get to the point where it checks LAN2

src/ConfigFile.c
```diff
@@ -132,10 +132,29 @@ void Ros_ConfigFile_SetAllDefaultValues()
...
-        "Must enable ETHERNET function");
+
+    //TODO(gavanderhoorn): get MAC only from interface which is used for
+    //ROS traffic (https://github.com/Yaskawa-Global/motoros2/issues/57)
+    STATUS status = Ros_GetMacAddress(ROS_USER_LAN1, macId);
+    if (status != OK)
+    {
+        Ros_Debug_BroadcastMsg("%s: Ros_GetMacAddress: iface: %d; error: %d",
+            __func__, ROS_USER_LAN1, status);
+        motoRosAssert_withMsg(FALSE, SUBCODE_CONFIGURATION_FAIL_MP_NICDATA0,
```
ted-miller, line 143: Same issue here

ted-miller (changes requested):

src/CommunicationExecutor.c
gavanderhoorn: I know. But do we want to just try any interface which works, or should we actually just raise an alarm if we can't retrieve the MAC for the interface we're asking for?

I would probably suggest to do the first, as the previous behaviour could lead to situations where MotoROS2 ends up trying to use the 3rd interface while the 2nd is actually the one with the network cable (ie: for some reason, 2nd interface MAC can't be read, 3rd interface succeeds -> we just continue with that interface).

ted-miller: > I would probably suggest to do the first, as the previous behaviour could lead to situations where MotoROS2 ends up trying to use the 3rd interface while the 2nd is actually the one with the network cable (ie: for some reason, 2nd interface MAC can't be read, 3rd interface succeeds -> we just continue with that interface).

I don't understand this comment. I'm not sure which of the two you're suggesting.

Personally, I'd just go with any interface that works.

Also, if we do maintain this assertion, then the check for the second MAC will never logically happen. Either status is bad and there's an alarm, or status is good and the second one never gets checked.

gavanderhoorn: Failure to read the MAC is not necessarily related to whether or not an interface is `UP` and/or `RUNNING`.

We're already tracking the case where with two interfaces connected, MotoROS2 always uses the first interface (ie: `ROS_USER_LAN1`) to get the MAC (#57) and to broadcast the debug logs (#58).

I've learned that failing early (and loudly) is better than trying to continue with whatever fallback does work.

Hence the question/comment.

> Also, if we do maintain this assertion, then the check for the second MAC will never logically happen. Either status is bad and there's an alarm, or status is good and the second one never gets checked.

yes, agreed.

0ec05f5ea8db598e4a855da72dda4a659fec165a was essentially just the old pattern but adapted to:

 - the changed `Ros_GetMacAddress(..)` implementation, and
 - no longer reuse alarm codes

gavanderhoorn: @ted-miller: would implementing the previous behaviour (ie: just try interfaces until one works) unblock this PR?

If so, I'll make that change and push it here.

We can update/correct later if desired.

ted-miller (approved):

ted-miller (approved):

PR #105, opened 2023-08-01
Fix `Ros_Controller_GetActiveAlarmCodes(..)`

### LINKED ISSUE #104: A pendant error causes failure in `Ros_Controller_GetActiveAlarmCodes`
I got an error on the pendant because I had the wrong media-type selected.

image

Then I noticed on the debug log that it was spamming `Error retrieving active alarms: -2` at a high frequency.
It stopped when I cleared the error on the pendant.

Turns out that this line is treating the "error number" as the "number of errors" instead of an error code. 
https://github.com/Yaskawa-Global/motoros2/blob/c2453273e80dfe2841d5338b08d489af943c5214/src/ControllerStatusIO.c#L739

There can only ever be one error.

gavanderhoorn:
Fixes #104.

Should fix it, at least. I haven't had an opportunity to test it.

### DIFF
--- a/src/ControllerStatusIO.h
+++ b/src/ControllerStatusIO.h
@@ -140,7 +140,7 @@ extern void Ros_Controller_SetIOState(ULONG signal, BOOL status);
 extern int Ros_Controller_GetAlarmCode();
 
 //retrieve all active alarms (and a possible error) and store them in 'active_alarms'
-extern int Ros_Controller_GetActiveAlarmCodes(USHORT active_alarms[MAX_ALARM_COUNT + 1]);
+extern int Ros_Controller_GetActiveAlarmCodes(USHORT active_alarms[MAX_ALARM_COUNT + MAX_ERROR_COUNT]);
 
 
 //#define DUMMY_SERVO_MODE 1    // Dummy servo mode is used for testing with Yaskawa debug controllers
--- a/src/MotoPlusExterns.h
+++ b/src/MotoPlusExterns.h
@@ -15,4 +15,11 @@ extern size_t mpNumBytesFree(void);
 
 extern MP_GRP_ID_TYPE mpCtrlGrpNo2GrpId(int grp_no);
 
+//M+ only defines MAX_ALARM_COUNT, but we'd like to avoid magic nrs
+//as much as possible, so we define our own here for errors.
+//But only if something hasn't been defined already.
+#ifndef MAX_ERROR_COUNT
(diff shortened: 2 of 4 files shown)


src/MotoPlusExterns.h
```diff
@@ -15,4 +15,11 @@ extern size_t mpNumBytesFree(void);
 
 extern MP_GRP_ID_TYPE mpCtrlGrpNo2GrpId(int grp_no);
 
+//M+ only defines MAX_ALARM_COUNT, but we'd like to avoid magic nrs
+//as much as possible, so we define our own here for errors.
+//But only if something hasn't been defined already.
+#ifndef MAX_ERROR_COUNT
+#define MAX_ERROR_COUNT 1
+#endif
```
gavanderhoorn, line 23: we could perhaps also put this in the platform lib, but seeing as it's basically something that should be part of M+, I decided to treat it like something that *is* part of M+, but just not `define`d in the (default) headers.

src/ControllerStatusIO.h
```diff
@@ -8,6 +8,9 @@
 #ifndef MOTOROS2_CONTROLLER_STATUS_IO_H
 #define MOTOROS2_CONTROLLER_STATUS_IO_H
 
+#include "MotoPlusExterns.h" // for MAX_ERROR_COUNT
```
ted-miller, line 11: It would be more consistent (style-wise) to remove this line and bump [the include](https://github.com/Yaskawa-Global/motoros2/blob/c2453273e80dfe2841d5338b08d489af943c5214/src/MotoROS.h#L98) up above [this line](https://github.com/Yaskawa-Global/motoros2/blob/c2453273e80dfe2841d5338b08d489af943c5214/src/MotoROS.h#L86).

gavanderhoorn: Probably.

It also highlights the fragility of the current style of `#include`-ing MotoROS2 currently uses.

We should really switch to `include-what-you-use` instead.

gavanderhoorn: I've re-ordered the `#include`s, but instead of squashing it into e8e66c77073e6fdf98e295b24446fb0e9a73657d, I've kept it as a separate commit.

Just to have a clear reminder.

ted-miller:
> I don't know how to (reliably, easily) trigger an error, so I'm not sure how to test this.

Just push PLAY when the robot is in TEACH mode.

I tested. All good.

ted-miller (approved):

PR #106, opened 2023-08-01
Fixed Cryptic Error Codes in mpStartJob

### LINKED ISSUE #102: Cryptic error codes from mpStartJob
When enabling a motion mode, the software tries to start the INFORM job. If it fails to start that job, it returns a cryptic error code. You would need to reference the MotoPlus documentation to find the definition. 

Ex: `Can't start 'INIT_ROS' because: 0x3040`

We should map the error codes to a lookup table with human-readable strings.

```
0x0000 Normal end
0x2010 Robot is in operation
0x2030 In HOLD status (PP)
0x2040 In HOLD status (External)
0x2050 In HOLD status (Command)
0x2060 In error/alarm status
0x2070 In SERVO OFF status
0x2080 Wrong operation mode
0x3040 The home position is not registered
0x3050 Out of range (ABSO data)
0x3400 Cannot operate MASTER JOB
0x3410 The JOB name is already registered in another task.
0x4040 Specified JOB not found
```

aigallegos:
fixes #102. I wrote a new function that kept the error string format as an f-string for mpStartJob errors in MotionControl.c by just adding a new function called Ros_ErrorHandling_Only_ErrNo_ToString in ErrorHandling.c that did not use strncpy and therefore did not need a buffer to get passed in, similar to Ros_ErrorHandling_MotionNotReadyCode_ToString it just returns a char* instead of having a destination for the chars and can be directly passed into the error f-string.

### DIFF
--- a/src/ErrorHandling.c
+++ b/src/ErrorHandling.c
@@ -15,6 +15,8 @@ const char* const Ros_ErrorHandling_ErrNo_ToString(int errNo)
 {
     switch (errNo)
     {
+    //Note: returning literals here as 'const char*' is OK, as they are stored
+    //in an anonymous array with static storage.
     case 0x2010: return "Robot is in operation";
     case 0x2030: return "In HOLD status (PP)";
     case 0x2040: return "In HOLD status (External)";
--- a/src/ErrorHandling.c
+++ b/src/ErrorHandling.c
@@ -32,41 +32,6 @@ const char* const Ros_ErrorHandling_ErrNo_ToString(int errNo)
     }
 }
 
-const char* const Ros_ErrorHandling_Only_ErrNo_ToString(int errNo)
-{
-    switch (errNo)
-    {
-    case 0x2010:
-        return "Robot is in operation";
-    case 0x2030:
-        return "In HOLD status (PP)";
-    case 0x2040:
-        return "In HOLD status (External)";
-    case 0x2050:
-        return "In HOLD status (Command)";
-    case 0x2060:
-        return "In ERROR/ALARM status";
-    case 0x2070:
-        return "In SERVO OFF status";
-    case 0x2080:
-        return "Wrong operation mode";
-    case 0x3040:
-        return "The home position is not registered";
-    case 0x3050:
-        return "Out of range (ABSO data)";
-    case 0x3400:
-        return "Cannot operate MASTER JOB";
-    case 0x3410:
-        return "The JOB name is already registered in another task";
-    case 0x4040:
-        return "Specified JOB not found";
-    case 0x5200:
-        return "Over data range";
-    default:
-        return "Unspecified reason";
-    }
-}
-
 const char* const Ros_ErrorHandling_MotionNotReadyCode_ToString(MotionNotReadyCode code)
 {
     //messages defined in motoros2_interfaces/msg/MotionReadyEnum.msg
--- a/src/ErrorHandling.c
+++ b/src/ErrorHandling.c
@@ -11,24 +11,24 @@
 //-------------------------------------------------------------------
 // Convert error code to string
 //-------------------------------------------------------------------
-void Ros_ErrorHandling_ErrNo_ToString(int errNo, char* const errMsg, int errMsgSize)
+const char* const Ros_ErrorHandling_ErrNo_ToString(int errNo)
 {
     switch (errNo)
     {
-    case 0x2010: strncpy(errMsg, "Robot is in operation", errMsgSize); break;
-    case 0x2030: strncpy(errMsg, "In HOLD status (PP)", errMsgSize); break;
-    case 0x2040: strncpy(errMsg, "In HOLD status (External)", errMsgSize); break;
-    case 0x2050: strncpy(errMsg, "In HOLD status (Command)", errMsgSize); break;
-    case 0x2060: strncpy(errMsg, "In ERROR/ALARM status", errMsgSize); break;
-    case 0x2070: strncpy(errMsg, "In SERVO OFF status", errMsgSize); break;
-    case 0x2080: strncpy(errMsg, "Wrong operation mode", errMsgSize); break;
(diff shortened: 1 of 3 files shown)


src/ErrorHandling.c
```diff
@@ -32,6 +32,41 @@ void Ros_ErrorHandling_ErrNo_ToString(int errNo, char* const errMsg, int errMsgS
     }
 }
 
+const char* const Ros_ErrorHandling_Only_ErrNo_ToString(int errNo)
```
gavanderhoorn, line 35: This seems to duplicate what `Ros_ErrorHandling_ErrNo_ToString(..)` does above. It also appears to return pointers to temporary objects.

Could we use `Ros_ErrorHandling_ErrNo_ToString(..)` instead of adding a new function?

src/MotionControl.c
```diff
@@ -1287,7 +1287,7 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
     {
         //TODO(gavanderhoorn): special check for "job is not loaded"
         Ros_Debug_BroadcastMsg(
-            "Can't start '%s' because: 0x%04X", g_nodeConfigSettings.inform_job_name, rData.err_no);
+            "Can't start '%s' because: 0x%04X %s", g_nodeConfigSettings.inform_job_name, rData.err_no, Ros_ErrorHandling_Only_ErrNo_ToString(rData.err_no));
```
gavanderhoorn, line 1290: please see my https://github.com/Yaskawa-Global/motoros2/issues/102#issuecomment-1661579552.

I would suggest we replace the other case where we print `rData.err_no` directly as well.

gavanderhoorn (dismissed):

gavanderhoorn:
> I wrote a new function that kept the error string format as an f-string for mpStartJob errors in MotionControl.c by just adding a new function called Ros_ErrorHandling_Only_ErrNo_ToString in ErrorHandling.c that did not use strncpy and therefore did not need a buffer to get passed in

thing is, that buffer & copy are needed, as functions can't return pointers to temporaries (ie: variables on the stack).

> similar to Ros_ErrorHandling_MotionNotReadyCode_ToString it just returns a char* instead of having a destination for the chars and can be directly passed into the error f-string.

you could consider defining all error 'strings' as

```c
static char something[] = "...";
```

which would allow just `return`-ing `something` directly.

`Ros_ErrorHandling_MotionNotReadyCode_ToString(..)` used to do this, but that's invisible now, as the storage for the 'strings' returned by it is the responsibility of the code generated by the ROS 2 message generators now.

src/ErrorHandling.c
aigallegos: Sure! I just had though you wanted it to function similarly to `Ros_ErrorHandling_MotionNotReadyCode_ToString(..)` and return a temporary object which `Ros_ErrorHandling_ErrNo_ToString(..)` does not do. What should the object that should be the destination (the 2nd parameter) that is passed into `Ros_ErrorHandling_ErrNo_ToString(..)` be? Can I create a new buffer specifically for this?

gavanderhoorn: > Sure! I just had though you wanted it to function similarly to `Ros_ErrorHandling_MotionNotReadyCode_ToString(..)` and return a temporary object which `Ros_ErrorHandling_ErrNo_ToString(..)` does not do. 

`Ros_ErrorHandling_MotionNotReadyCode_ToString(..)` does not return a pointer-to-temporary. It returns the `const char*` which points to a `static`ally allocated string inside the type support library for `motoros2_interface`.

As follows (from `target_ws\install\include\motoros2_interfaces\msg\detail\motion_ready_enum__struct.h`, generated code):

<details><summary>Click to expand</summary>

```c
// Constants defined in the message

/// Constant 'READY'.
/**
  * Result codes which may be returned by services and actions which attempt to
  * execute motion (or are executing motion).
  *
  * Values other than 1 would indicate an active error or other active inhibition
  * on the Yaskawa controller preventing MotoROS2 from enabling trajectory mode
  * or executing any motion.
  *
  * Applications are encouraged to refer to the named constants generated by
  * rosidl as part of this message instead of using the integer values directly.
 */
enum
{
  motoros2_interfaces__msg__MotionReadyEnum__READY = 1
};

/// Constant 'READY_STR'.
static const char * const motoros2_interfaces__msg__MotionReadyEnum__READY_STR = "Ready";

...
```

</details>

The `static` there changes the lifetime of the strings to essentially `as-long-as-the-program-runs`.

> What should the object that should be the destination (the 2nd parameter) that is passed into `Ros_ErrorHandling_ErrNo_ToString(..)` be? Can I create a new buffer specifically for this?

There are two options I believe:

 1. you just use `Ros_ErrorHandling_ErrNo_ToString(..)` as-is. That would mean you'd have to provide a `char[]` for the `strncpy(..)` to copy into.
 2. we could change `Ros_ErrorHandling_ErrNo_ToString(..)` to also have it return `const char*` to `static`ally allocated strings. That should solve the issue identified here. `Ros_ErrorHandling_ErrNo_ToString(..)` is not currently used, so it would be a relatively minor change without too much fall-out.

I'll push a commit which implements option 2.

gavanderhoorn:
Note: we should consider squash-merging this if possible. All my commits are essentially fix-ups.

ted-miller (approved):
Looks good.

```
THU 2023-08-03 15:03:27.596 Ros_MotionControl_StartMotionMode: enter
THU 2023-08-03 15:03:29.876 Can't start 'INIT_ROS' because: 'The home position is not registered' (0x3040)
```

Thank you both for the PR!

gavanderhoorn:
Thanks @aigallegos 👍

aigallegos:
> Thanks @aigallegos 👍

Thank YOU! I learned so much from this rabbit hole we went down and thanks for putting up with my newbie negligence that just so happened to be correct haha. Thanks for the help and happy coding!

gavanderhoorn:
If you see any other issues you believe you could help with, please comment on them and we can discuss potential approaches.

PR #109, opened 2023-08-03
Proposal: switch to shorter names for MotoROS2 binaries

gavanderhoorn:
The main change is in f44a53715c92f43f201ebff2e7c375289bbc2598: it updates the VS project to change the names of the `.out` files to their shorter alternatives as discussed in #107.

See the commit comment for the naming scheme used.

The second commit updates the readme to refer to the binaries by their new names.


gavanderhoorn:
@ted-miller: this uses `m` for the YRC1000micro, mimicking the `.yrcmLib` extension.

That's different from what we currently use for the micro (`motoros2_yrc1000u_humble.out`). I don't really have a preference, so we could change it to `u` if you'd prefer.

ted-miller (approved):
Are you going to do this on the DX2 PR too?

gavanderhoorn:
> Are you going to do this on the DX2 PR too?

I'll rebase #49 and add the missing changes there.

We need to review the readme and other files for missing DX200 references a final time anyway.

PR #115, opened 2023-08-07
Don't try to start jobs/drives when in a fault state

### LINKED ISSUE #113: `Ros_MotionControl_StartMotionMode` should check `Ros_Controller_IsMpIncMoveErrorActive`
I was running a point-queue. Then I hit estop. I cleared the estop. Then I tried to activate FJT-mode. 

The drives power up and the job starts. But the service returns a nasty message about `There was an error with the internal motion API`. I was very confused.

I do now recall that we're intentionally not resetting the internal `bMpIncMoveError` flag after an estop until the user explicitly calls `/reset_error`. But I think that `Ros_MotionControl_StartMotionMode` should not have tried to enable the servos and start the job. Additionally, it should return a better error message indicating that the robot is in a (clearable) error state.

ted-miller:
Fixes #113

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -1177,6 +1177,12 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
     }
 #endif
 
+    if (Ros_Controller_IsAnyFaultActive())
+    {
+        Ros_Debug_BroadcastMsg("Controller is in a fault state. Please call /reset_error");
+        return FALSE;
+    }
+
     // Check if currently in operation, we don't want to interrupt current operation
     if (Ros_Controller_IsOperating())
     {


src/MotionControl.c
```diff
@@ -1208,6 +1208,12 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
         }
     }
 
+    if (Ros_Controller_IsAnyFaultActive())
+    {
+        Ros_Debug_BroadcastMsg("Controller is in a fault state. Please call /reset_error");
+        return FALSE;
+    }
+
```
gavanderhoorn, line 1216: I would perhaps suggest to move this up, to right after the call to `Ros_Controller_IsRemote()`.

Doesn't seem much point to 'wait this long' to check whether there is an active alarm if we could check early on and just bail out.

It should still work in case we're just trying to get servos out of eco mode, as that would not work with an active alarm/error in any case.

ted-miller: Updated

gavanderhoorn (approved):

gavanderhoorn:
PS: I would suggest to use a separate branch next time. If you push it to the main repository, we also get the benefit of having CI run against the branch.

PR #118, opened 2023-08-14
Pass in a specific `rcl_allocator_t` instance to initialise

gavanderhoorn:
As per subject.

Instead of assuming it's always `g_motoros2_Allocator`.

Functionally equivalent, but I have a need to use multiple allocators in another project and this allows me to reuse this bit of MR2.


gavanderhoorn:
Argh. I forgot about the restrictions wrt CI. I'll push to the main repository and re-open.

PR #123, opened 2023-08-15
Remove duplicated `min(..)` & `max(..)` macros

gavanderhoorn:
As per subject.

The are already provided by `motoPlus.h` in the M+ SDK.


gavanderhoorn:
Yes, I remember you adding these (about two years ago).

It's indeed in all versions of `motoPlus.h` we use, so it didn't seem like we should keep it around.

PR #126, opened 2023-08-16
Include ROS 2 version in `AppInfo` on controller

gavanderhoorn:
I noticed we actually don't have any way of determining _which_ ROS 2 version is supported by which MR2 binary, other than asking the user.

Now that #109 is merged, it's possible to derive this information from the name of the `.out`, but that's not very convenient.

This PR changes the way MotoROS2 registers itself with the "application information" infrastructure to add the ROS 2 codename to the version nr/string.

It changes the line in the `PANELBOX.LOG` from :

```
//MOTOPLUS APP
01: MotoROS2                         0.1.0            micro-ROS based ROS 2 interface 
```

to:

```
//MOTOPLUS APP
01: MotoROS2 (humble)                0.1.0            micro-ROS based ROS 2 interface 
```

From this, it's clear MotoROS2 version `0.1.0` supporting ROS 2 Humble is installed on this controller.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -40,11 +40,12 @@ void Ros_ReportVersionInfoToController()
     MP_APPINFO_SEND_DATA appInfoSendData;
     MP_STD_RSP_DATA stdResponseData;
 
-    sprintf(appInfoSendData.AppName, "%s (%s)",
-        APPLICATION_NAME, MOTOPLUS_LIBMICROROS_ROS2_CODENAME);
-    sprintf(appInfoSendData.Version, "%s", APPLICATION_VERSION);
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
I've kept this a *draft*, as it requires updates to the `libmicroros` distribution `.zip`s.

Until that's done, CI will fail.

I'll probably update the `.zip`s currently attached to the `0.1.0` release.

gavanderhoorn:
As an alternative, we could put the ROS 2 codename string in the `AppName` field (it's longer), but that didn't seem right, as the ROS 2 codename isn't part of the name of the application itself.

gavanderhoorn:
With the updated M+ `libmicroros` distribution `.zip`s CI now succeeds.

ted-miller (approved):
looks good

PR #136, opened 2023-08-25
Cleanup `Ros_Debug_BroadcastMsg(..)` a bit

gavanderhoorn:
I noticed `Ros_Debug_BroadcastMsg(..)` duplicated some code, and seemed to (ab)use `builtin_interfaces__msg__Time` and `Ros_Nanos_To_Time_Msg(..)` to convert from nanoseconds to seconds.

This PR introduces a new convenience conversion function (nanoseconds -> `struct timespec`) and reduces duplication by making only the retrieval of the 'current time' conditional on whether or not the Agent is connected (the rest is the same for both cases).

Build-tested only.


src/Debug.c
```diff
@@ -48,28 +48,28 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
...
+        Ros_Nanos_To_Timespec(nanosecs, &tp);
     }
     else
     {
         //rmw_uros_epoch_nanos cannot sync with agent because it's not connected
         clock_gettime(CLOCK_REALTIME, &tp);
-        strftime(timestamp, FORMATTED_TIME_SIZE, "%Y-%m-%d %H:%M:%S", localtime_r(&tp.tv_sec, &synced_time));
-        snprintf(timestamp + strlen(timestamp), FORMATTED_TIME_SIZE - strlen(timestamp), ".%06d ", tp.tv_nsec / 1000);
     }
+    localtime_r(&tp.tv_sec, &synced_time);
```
gavanderhoorn, line 69: Using `localtime_r(..)` like this facilitates use on FS100 a bit as well, as we don't use the return value directly any more (which is different between FS100 and the rest).

ted-miller (approved):

PR #157, opened 2023-09-20
Port FSU speed limit handling from MotoROS1

iydv:
As discussed at [here](https://github.com/ros-industrial/motoman/pull/542) I have implemented motoros2 version of FSU speed limit handling. Additionally, I have introduced following changes to resolve issues in the implementation:

1) `prevPulsePosData` is defined as global variable in `CtrlGroup`. This is required, since this variable is constantly changing in `Ros_MotionControl_IncMoveLoopStart` function. The `prevPulsePos` variable cannot be used instead, because it will break the logic of [`Ros_MotionControl_AddToIncQueueProcess`](https://github.com/iydv/motoros2/blob/75055e1e6b60f73e6abe80dab213f0e3ce889a1f/src/MotionControl.c#L518). 

This change is required to allow initialization of `prevPulsePosData` when trajectory mode is disabled/enabled. Otherwise, the robot will always tried to return to its last position every time the trajectory mode is enabled. 

2) `feedback_FollowJointTrajectory.feedback.desired.positions` is estimated differently now. Previously, it was accumulating all `moveData` increments in every iteration of `Ros_MotionControl_IncMoveLoopStart` function. However, the robot cannot reach desired increments due to the speed limit. Which means that `Ros_ActionServer_FJT_UpdateProgressTracker` function is called multiple times with increments that are never actually executed by the robot. 
I have changed `feedback_FollowJointTrajectory.feedback.desired.positions` to represent a current desired increment of actual positions, which better reflects the behavior of `Ros_MotionControl_IncMoveLoopStart` with current changes.

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -773,8 +773,6 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
     BOOL isMissingPulse;                                                // Flag that there are pulses send in last cycle that are missing from the command (pulses were not processed)
     BOOL hasUnprocessedData;                                            // Flag that at least one axis (any group) still has unprecessed data. (Used to continue sending data after the queue is empty.)
 
-    ctrlGrpData.sCtrlGrp = 0;
-
     bzero(newPulseInc, sizeof(LONG) * MP_GRP_AXES_NUM * MAX_CONTROLLABLE_GROUPS);
     bzero(toProcessPulses, sizeof(LONG) * MP_GRP_AXES_NUM * MAX_CONTROLLABLE_GROUPS);
     bzero(processedPulses, sizeof(LONG) * MP_GRP_AXES_NUM);
@@ -796,6 +794,8 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
     {
         moveData.ctrl_grp |= (0x01 << i);
         moveData.grp_pos_info[i].pos_tag.data[0] = Ros_CtrlGroup_GetAxisConfig(g_Ros_Controller.ctrlGroups[i]);
+
+        ctrlGrpData.sCtrlGrp = g_Ros_Controller.ctrlGroups[i]->groupId;
         mpGetPulsePos(&ctrlGrpData, &prevPulsePosData[i]);
     }
 
@@ -918,6 +918,7 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
                 // Get the current controller command position and substract the previous command position
                 // and check if it matches the amount if increment sent last cycle.  If it doesn't then
                 // some pulses are missing and the amount of unprocessed pulses needs to be added to this cycle.
+                ctrlGrpData.sCtrlGrp = g_Ros_Controller.ctrlGroups[i]->groupId;
                 mpGetPulsePos(&ctrlGrpData, &pulsePosData);
                 isMissingPulse = FALSE;
                 for (axis = 0; axis < MP_GRP_AXES_NUM; axis++)
@@ -1103,6 +1104,7 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
             hasUnprocessedData = FALSE;
             for (i = 0; i < g_Ros_Controller.numGroup; i++)
             {
+                ctrlGrpData.sCtrlGrp = g_Ros_Controller.ctrlGroups[i]->groupId;
                 mpGetPulsePos(&ctrlGrpData, &prevPulsePosData[i]);
             }
         }
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -751,17 +751,27 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
     MP_PULSE_POS_RSP_DATA prevPulsePosData[MAX_CONTROLLABLE_GROUPS];
     MP_PULSE_POS_RSP_DATA pulsePosData;
 
-    LONG newPulseInc[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];		// Pulse increments that we just retrieved from the incQueue
-    LONG toProcessPulses[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];	// Total pulses that still need to be sent to the command
-    LONG processedPulses[MP_GRP_AXES_NUM];							// Amount of pulses from the last command that were actually processed (accepted) 
-    LONG prevMaxSpeed[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];
-    LONG prevMaxSpeedRemain[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];
-    LONG maxSpeed[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];
-    LONG maxSpeedRemain[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];
-    BOOL skipReadingQ[MAX_CONTROLLABLE_GROUPS];
-    BOOL queueRead[MAX_CONTROLLABLE_GROUPS];
-    BOOL isMissingPulse;
-    BOOL hasUnprocessedData;
+    // --- FSU Speed Limit related ---
+    // When FSU speed limitation is active, some pulses for an interpolation cycle may not be processed by the controller.
+    // To track the true amount of pulses processed, we keep track of the command position and by substracting the
+    // the previous position from the current one, we can confirm the amount if pulses precessed.
+    // If the amount processed doesn't match the amount sent at the previous cycle, we resend the unprocessed pulses amount.
+    // To keep the motion smooth, the 'maximum speed' (max pulses per cycle) is tracked and used to skip reading
+    // more pulse increment from the queue if the amount of unprocessed pulses is larger than the detected speed.
+    // The 'maximum speed' is also used to prevent exceeding the commanded speed once the FSU Speed Limit is removed.
+    //
+    // The following set of variables are used to track FSU speed limitation.
+    LONG newPulseInc[MAX_CONTROLLABLE_GROUPS][MP_GRP_AXES_NUM];         // Pulse increments that we just retrieved from the incQueue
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Hi, thanks for the PR :+1: 

Could I be difficult and ask you to split it into multiple commits?

Ideally a first commit showing just the 'import' of @EricMarcil's work here in MotoROS2, then further commits with the changes you describe you had to make.

That would facilitate review as it would very clearly show what changed, where and when.

thanks.

iydv:
> Hi, thanks for the PR 👍
> 
> Could I be difficult and ask you to split it into multiple commits?
> 
> Ideally a first commit showing just the 'import' of @EricMarcil's work here in MotoROS2, then further commits with the changes you describe you had to make.
> 
> That would facilitate review as it would very clearly show what changed, where and when.
> 
> thanks.

Hi, I have split it into 3 commits

ted-miller:
I will try to do some testing next week. I don't currently have a system with FSU configured.

EricMarcil:
@iydv 
I probably should have looked at you solution before working on the MotoRos1 version on my PR.
[FSU Speed Limit Handling](https://github.com/ros-industrial/motoman/pull/542)
Please review the way I have done it.  I believe that in your correction, there will still be a small motion.  During testing, I also noticed a small motion that was caused by the "toProcessPulses" not being re-initialized.  
See [commit](https://github.com/ros-industrial/motoman/pull/542/commits/0654dfaba390033b21e98c6729ea09ad286e6edd)

iydv:
@EricMarcil thank you for the update. I have checked your solution and it is better than mine. I will implement and test it in MotoROS2 (which I can do only next week). 
Unfortunately, I am no longer using ROS1 so I cannot test the original MotoROS1 implementation.

iydv:
I have integrated changes from the latest MotoROS1 [commit](https://github.com/ros-industrial/motoman/pull/542/commits/0654dfaba390033b21e98c6729ea09ad286e6edd) to reset `prevPulsePosData ` and
 `toProcessPulses`. 

@EricMarcil unfortunately, I encountered another issue with this solution: when the robot is moved away and enabled again the following error is triggered:
`mpExRcsIncrementMove returned UNREADY: -1 (Could be PFL Active)` 

I believe the issue is [this condition](https://github.com/iydv/motoros2/blob/c738598462973ebc4d737e685a1e28233cc347cd/src/MotionControl.c#L796). The variable `hasUnprocessedData` is never reset inside the main loop, which means it is always `TRUE` and when the robot is enabled it always enters the queue processing loop. However, by original logic, it should enter the loop only when the queue has data. I added a reset of `hasUnprocessedData` [here](https://github.com/iydv/motoros2/blob/c738598462973ebc4d737e685a1e28233cc347cd/src/MotionControl.c#L1079). And I think it should also reset during normal operation similar to [`isMissingPulse`](https://github.com/iydv/motoros2/blob/c738598462973ebc4d737e685a1e28233cc347cd/src/MotionControl.c#L901C17-L901C32)

ted-miller:
I have tested this on a few trajectories and it seems to work well. Admittedly, my trajectories are rather primitive and hard-coded. But the behavior with speed-limiting enabled always matches the behavior with speed-limiting disabled. I can even switch back and forth mid trajectory. 

The only unfortunate part is that this makes the most-complicated section of code even more complicated. But I guess that's inevitable. 

@gavanderhoorn, I think this is a great addition and should in the `0.0.12` milestone. Any arguments against it?
(Just today I had a user contact me regarding this issue. So, it'd be good timing for a fix.)

EricMarcil:
I've also been reviewing and testing the last recommendation from @iydv.  I agree with his change but in the process, I also found a new condition where the robot can still move.  If in the middle of a motion, the mode key is switch to TEACH mode, and then the robot is manually restarted by switching back to PLAY, turning on the servos and then switch back to REMOTE.  The robot continues execution what's left in the motion buffer.
Note: I haven;t tested yet, but I don't believe this is specific to the change in this PR.  It's probably an issue that has always existed and was never detected.

ted-miller:
Are you testing with MotoROS2 or MotoROS1?
I'm not seeing this behavior on MR2.

EricMarcil:
I'm testing on MotoRos1.  I'm working on fixing it.  I'll commit to the PR on MotoRos1 and then you can review to see if it applied on MotoRos2.

ted-miller:
> you can review to see if it applied on MotoRos2

It looks like this does _not_ apply to MR2.

ted-miller (approved):

gavanderhoorn:
Could we perhaps add some more comments? As you write @ted-miller, this is already some of the most -- if not *the* most -- complex code in MotoROS (1 and 2) and this PR makes it even more complex.

Seeing the discussion on https://github.com/ros-industrial/motoman/pull/542, it's a non-trivial change which adds quite a bit of bookkeeping and handling for new corner-cases.

I'd like to see more comments explaining what's going on, perhaps also at a conceptual level. https://github.com/ros-industrial/motoman/pull/542 has some of that in the PR description and the subsequent comments, but we should embed that and have it right next to the source.

I would perhaps also suggest setting an IO signal whenever the new code detects the FSU is 'active'. That would allow us to use that in other places without increasing coupling even more than we already do.

gavanderhoorn (changes requested):
Marking as *request changes* to prevent us from merging this accidentally.

gavanderhoorn:
Note: @iydv, as this is essentially @EricMarcil's code, I'm not asking you add comments of course.

As maintainers we can add more commits to this PR, which I what I suggest we do @ted-miller and @EricMarcil.

gavanderhoorn:
@EricMarcil / @ted-miller: friendly ping?

ted-miller:
It's on the todo-list. 

I'd be in favor of merging this and opening another issue for the next milestone regarding comments/documentation. We also need to update the [EDS document](https://wiki.ros.org/motoman_driver?action=AttachFile&do=view&target=MotoRos_EDS.pdf) to explain the overall architecture of the code.

gavanderhoorn:
@iydv: I've just pushed a commit which cherry-picked @EricMarcil's comments from https://github.com/ros-industrial/motoman/commit/9c6ff7d6ef98490b09eed15ca124be0a2051c23e to your branch.

gavanderhoorn:
I don't have an FSU configured, so I can't test this myself right now.

@ted-miller: if you could verify again this works as intended -- also with different (ie: multi-group) systems?

I'll leave the decision to merge to you.

gavanderhoorn:
It would also still be good to address my previous comment:

> I would perhaps also suggest setting an IO signal whenever the new code detects the FSU is 'active'. That would allow us to use that in other places without increasing coupling even more than we already do.

but we can do that in a follow up PR.

src/MotionControl.c
```diff
@@ -755,88 +796,236 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
...
+                    toProcessPulses[i][axis] -= processedPulses[axis];
+                    if (toProcessPulses[i][axis] != 0)
+                        isMissingPulse = TRUE;
+
+                    // Add the new pulses to be processed for this iteration 
+                    toProcessPulses[i][axis] += newPulseInc[i][axis];
+
+                    if (toProcessPulses[i][axis] != 0)
+                        hasUnprocessedData = TRUE;
+                }
```
gavanderhoorn, line 940: @EricMarcil: could we assume the ratio between submitted and processed pulses could be used to calculate the amount / percentage by which the FSU is currently throttling robot performance?

EricMarcil: @gavanderhoorn It might give some idea but it is difficult to evaluate.  The FSU Speed Limit is usually a restriction in mm/s in the Cartesian space but the MotoRos is working in the joint space.  It would probably be easier to just observe the motion resulting speed at the end of arm.

iydv:
I have tested all recent changes including [this](https://github.com/Yaskawa-Global/motoros2/pull/157/commits/13511c033f88806035222da87588f942075ff773) and everything still works fine.

ted-miller:
Great. I'm setting up a multi-group system now to verify. If there are no issues, I'll merge.

iydv:
@ted-miller can [this](https://github.com/iydv/motoros2/blob/13511c033f88806035222da87588f942075ff773/src/MotionControl.c#L776) be the issue? In [here](https://github.com/iydv/motoros2/blob/13511c033f88806035222da87588f942075ff773/src/MotionControl.c#L921) the current controller command position is retrieved. But if the command group is always 0, the positions for all groups are set to position of group 0?

ted-miller:
@iydv, Good catch. That does look wrong. 

Unfortunately, it's not the only problem. Now the robots come to immediate violent halt after they start moving. 
image

I'm guessing R2 is getting commanded with some uninitialized value.

iydv:
maybe initialization [here](https://github.com/iydv/motoros2/blob/13511c033f88806035222da87588f942075ff773/src/MotionControl.c#L799) ? The `prevPulsePosData` for all groups is getting initialized by the data from group 0. And then [here](https://github.com/iydv/motoros2/blob/13511c033f88806035222da87588f942075ff773/src/MotionControl.c#L926) the computation is wrong for any group other than 0.

ted-miller:
Yep. There were a few instances.

It is working now (with only minimal physical damage done...)

Given that the main functionality has been thoroughly vetted at this point, and that we've now identified a fix for other control groups, I'd say this is good to merge.

As a side note, I was surprised to find that activating the speed limit for _any_ group in the job will activate a speed limit for _all_ groups in the job. This makes sense in hindsight, but I was surprised. This is good because it keeps everything in sync.

I also attempted to create a custom INIT_ROS job which `pstart`ed individual jobs for each group to see how the speed limit would cause things to go out of sync. Turns out that you can't do that. The motion API returns:

image

Thank you @EricMarcil and @iydv for the implementation and debugging.

ted-miller (approved):

ted-miller:
> As a side note, I was surprised to find that activating the speed limit for any group in the job will activate a speed limit for all groups in the job. This makes sense in hindsight, but I was surprised. This is good because it keeps everything in sync.

Well... turns out this isn't entirely true. If the FSU board is not configured for a particular servo board, then group(s) on that board will not be limited.

image

I can't imagine a scenario where it would be intentionally configured like this. But its apparently possible.

gavanderhoorn:
That looks like something we might want to detect and warn about?

Issue #19, opened 2023-05-31
Add support for FS100


gavanderhoorn:
We'll expand this issue to track status of sub issues where we work on adding support for FS100 controllers to MotoROS2.

gavanderhoorn:
Added this to the *Untargeted* milestone, as this is `low-priority` for now.

gavanderhoorn:
@destogl FYI also.

gavanderhoorn:
Update: seems there are some more endianness issues to resolve: https://github.com/micro-ROS/micro-ROS-Agent/issues/197#issuecomment-1692789200.

gavanderhoorn:
Hi @saahu27.

> I want to understand, contribute and make use of fs100 controller using ros2.

great. We're always looking for more community participation.

> Can we please get in touch so that i can explain more about my project and work and discuss how can i contribute.

unless you feel what you're doing is (highly) sensitive, I'd suggest starting a discussion [on the Discussion forum](https://github.com/Yaskawa-Global/motoros2/discussions/categories/general).

At the moment I'm debugging some endianess issues, as you can read in my previous https://github.com/Yaskawa-Global/motoros2/issues/19#issuecomment-1693305876.

I expect basic MotoROS2 to work on FS100 once those get fixed (ie: publications and IO services, no motion yet). But no guarantees.

saahu27:
I have learnt a lot on how to use ros from your previous contributions for universal robots  and i feel very happy to have interacted with you.Thank you for your response.  I was not able to express myself clearly earlier. i wanted to request to drop your email so i can get in touch with you. 

Nothing that i do is classified  i wanted some help on how to go about controlling the motoman 100fs using ROS 2 and Move it2.
So, i am currently working on porting motoman ros1 drivers using ros1 to ros2 bridge.  i came across the new ros2 drivers you were contributing. Been trying your work on ROS1 Drivers and just wanted to ask some pointers on how to understand your new drivers, what kind of control can we expect a point to point control/Cartesian? Trajectory control?   with what baud rate so i can decide ros2  controller update frequency  etc which I will open a discussion in the page you mentioned.

gavanderhoorn:
> i wanted to request to drop your email so i can get in touch with you.

nothing personal, but I try to avoid discussing ROS topics in private emails if at all possible.

Most of my work is open-source, and I'd like to be able to reuse diagnostic sessions and discussions as much as possible as well.

> I will open a discussion in the page you mentioned.

[...]

PR #169, opened 2023-10-03
Raise alarm if calib data can't be loaded on multi-group systems with TF enabled

### LINKED ISSUE #54: Warn about failure to retrieve multi-robot calib data
The current implementation silently ignores failure to call `mpGetRobotCalibrationData(..)`:

https://github.com/Yaskawa-Global/motoros2/blob/8c1a524e8cad8e45e865454bb01678cabc416e38/src/ControllerStatusIO.c#L116-L135

MotoROS2 won't crash if the call fails, but it will result in TF transforms being incorrect, which will surprise users of multi-robot systems (especially if [they have performed the robot-to-robot calibration](https://github.com/Yaskawa-Global/motoros2/tree/8c1a524e8cad8e45e865454bb01678cabc416e38#incorrect-transform-tree-origin-with-multi-robot-setups)).

A new failure code should be defined and reported on the pendant.

gavanderhoorn:
But only if it hasn't been explicitly disabled by the user.

Should fix #54.

For the discussion: see also #54.

Haven't tested it yet.


gavanderhoorn:
Note: I intentionally didn't update subcode documentation as I'd like to test the new subcode doc linter.

gavanderhoorn:
Tested, works:

```
Ros_Controller_Initialize: calib loaded ok: 0, should warn: 1
Ros_Controller_Initialize: no calibration files loaded: TF potentially incorrect -> warn (disable with 'disable_missing_calib_warning')
```

on a YRC1 with a GP25 on a track without calibration.

ted-miller:
> on a YRC1 with a GP25 on a track without calibration

This comment got me thinking. I think I went about the base-track-tf all wrong. I don't think that base tracks use the same calibration method as multi-robot and station-axes. 

Do you even have a `Robot Calib` or `Grp Combination` button on your pendant?

image

image

I think the base track uses different parameters for its calibration. It's basically an implied group combination. 

If you jog your track, does your /tf visualization move with it?

Assuming I'm right about the base calibration (or lack thereof), then this alarm will always occur on a system with a base. Also, it would mean that /tf is never going to work right.

gavanderhoorn:
> Do you even have a `Robot Calib` or `Grp Combination` button on your pendant?

nope.

> If you jog your track, does your /tf visualization move with it?

nope.

> Also, it would mean that /tf is never going to work right.

with the current implementation I guess? What's needed would be to broadcast the current Yaskawa `Base` to `RF` IIUC?

gavanderhoorn:
Opened #170 to track the TF-is-incorrect-for-robots-on-tracks issue.

ted-miller:
> > Also, it would mean that /tf is never going to work right.
> 
> with the current implementation I guess? 

Yeah.

> What's needed would be to broadcast the current Yaskawa Base to Yaskawa RF IIUC?

The calibration from BF to RF exists. It's just not in the mpGetCalibration data. I think it's in SCG parameters.

src/ControllerStatusIO.c
```diff
@@ -114,23 +162,35 @@ BOOL Ros_Controller_Initialize()
...
+    const BOOL bCalibLoadedOk = Ros_Controller_LoadGroupCalibrationData(&g_Ros_Controller);
+
+    //It's OK if calibration did not load (for single-group systems fi, or systems
+    //that haven't been calibrated), but post an alarm just in case IFF:
+    // - this is a multi-group system
+    // - TF broadcast is enabled
+    // - none of the calibration files loaded, and
+    // - the warning has not been disabled
+    const BOOL bShouldWarnAboutCalib =
+        g_Ros_Controller.numGroup > 1 &&
```
ted-miller, line 174: This needs more intelligence to not count base tracks. I'm not sure the best way to do this, short of brute force. Just loop through each group and look at the `groupId`?

gavanderhoorn: Is it just base tracks/groups, or do we need to check for a specific combination of ids / relation between ids?

But we could easily add a new function which does this. Intended use would be during initialisation, so performance isn't too much of a concern directly (and even then: `O(n)` for small lists would not be problematic).

gavanderhoorn: @ted-miller: just to clarify: would you suppress the alarm if there are *any* base groups used for tracks? Or only in other specific cases?

ted-miller: > Is it just base tracks/groups, or do we need to check for a specific combination of ids / relation between ids?

AFAICT, just the base axes.

> would you suppress the alarm if there are any base groups used for tracks? Or only in other specific cases?

Specific cases.
R1+B1: No alarm
R1+B1+S1: Alarm
R1+B1+R2: Alarm

gavanderhoorn: Would it correct to say that any multi-group system with non-base-axis groups would fall into the `raise-alarm-if-no-calibration-files-loaded` category?

IIUC, base axes groups are always associated with robot groups, correct? So that leaves just station axes and proper robots.

ted-miller: That sounds correct

gavanderhoorn:
I believe this should now work.

In essence, this checks whether there are groups which don't have an associated base group and if there are, and if no calibration files loaded, TF is enabled and the warning is not suppressed, it raises the alarm.

In all other cases it stays silent.

gavanderhoorn:
@ted-miller: could you perhaps test this on some more complex group configurations?

ted-miller:
Sure. It'll be Monday or Tuesday. I've currently got everything configured for single-robot.

gavanderhoorn:
Friendly ping @ted-miller.

ted-miller:
So, with R1+R2+S1, if I don't have any calibrations, I get the alarm. If I then add a cal for R2+S1 and reboot, there is no alarm.

src/ControllerStatusIO.c
```diff
@@ -114,23 +171,40 @@ BOOL Ros_Controller_Initialize()
...
+    //that haven't been calibrated), but post an alarm just in case IFF:
+    // - this is a multi-group system
+    // - there are groups other than base groups (so R1+R2, or R1+S1)
+    // - TF broadcast is enabled
+    // - none of the calibration files loaded, and
+    // - the warning has not been disabled
+    const BOOL bNeedToWarnAboutCalib =
+        g_Ros_Controller.numGroup > 1 &&
+        //NOTE: a base group 'takes care' of 1 regular group and itself, hence x2
+        (g_Ros_Controller.numGroup - 2*numBaseGroups) > 0 &&
```
ted-miller, line 187: What if both robots are on a track, `R1+R2+B1+B2`? This would evaluate to false.

ted-miller (changes requested):

src/ControllerStatusIO.c
gavanderhoorn: Could you perhaps provide a few more examples of possible group configurations which you believe should raise the alarm, and a few which shouldn't?

I don't have sufficient information right now to derive a correct condition.

gavanderhoorn: @ted-miller: could a Station be calibrated against another Station? Or only Robots against Robots + Stations?

I'm thinking of just counting the nr of edges in a fully connected graph (recognising this is an undirected graph, so `/2`) where all `R` and `S` groups count as vertices. But if `S` groups can't be calibrated against each other (which I sort-of expect), I should subtract those from the nr of relations to take into account.

ted-miller: No, I don't think you can calibrate station-to-station.

I think the logic is simply, "If there is more than one group, ignoring bases". There has to be at least one robot. After that, it doesn't matter how many robots/stations there are.

gavanderhoorn: I have an updated implementation here: https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:warn_about_missing_calib. But I haven't had a chance to test it yet.

I'll update this PR if/when I do.

gavanderhoorn: I have a feeling the new unit tests are causing the MotoROS2 binary to run out of stack. [This test](https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:warn_about_missing_calib#diff-c6374f863bbf386ebe4c03efe704d83985947777c2af2952604d085b44fdd4ccR279) specifically tries to allocate a `Controller` and four `CtrlGroup`s. It doesn't like that.

ted-miller: Default stack size is supposedly 40k. Might be worth checking the `sizeof` some of those.

gavanderhoorn: According to `mpTaskCtrl.h`, we're currently using "the optimum value" for the stack. The next option would be `MP_STACK_LARGE_SIZE`, which is 50KB. But that just crashed my controller, so not sure what's going on.

> Might be worth checking the `sizeof` some of those.

that was indeed going to be my next step. I seem to remember especially `Controller` can be large, but I'm just mocking some simple `struct`s. Not even populating all fields/lists.

gavanderhoorn: @ted-miller: could you please test https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:warn_about_missing_calib instead of this PR? Note: the branch has the same name, but it's on my fork and is a different version of the changes in this PR.

Ignore the tests (so make sure to *not* set `MOTOROS2_TESTING_ENABLE`), I still haven't had time to look into the crashes.

ted-miller: My R1R2S1 configuration is broken right now. I'll need to re-init sometime tomorrow.

ted-miller: (Unless the `WRONG DATA` alarms are somehow caused by the code. I'm assuming your crashes are not WRONG DATA alarms.)

gavanderhoorn: Unless you've still got `MOTOROS2_TESTING_ENABLE` defined, there should not be any errors/alarms raised by this, other than the `8013[16]` user alarm.

src/ControllerStatusIO.c
```diff
@@ -751,3 +763,86 @@ int Ros_Controller_GetActiveAlarmCodes(USHORT active_alarms[MAX_ALARM_COUNT + MA
...
+}
+
+//TODO(gavanderhoorn): should be able to make this 'static' once Yaskawa-Global/motoros2#173
+//is implemented and we can unit test static functions
+BOOL Ros_Controller_ShouldWarnNoCalibDataLoaded(Controller const* controller, BOOL bCalibLoadedOk, BOOL bPublishTfEnabled)
+{
+    size_t numRobotsOrStations = 0;
+    for(size_t i=0; i < controller->numGroup; ++i)
+        if (Ros_CtrlGroup_IsRobot(controller->ctrlGroups[i]) || Ros_CtrlGroup_IsStation(controller->ctrlGroups[i]))
+            numRobotsOrStations += 1;
```
gavanderhoorn, line 823: I don't remember exactly any more why I'm counting robot groups here as well again -- we already have `Controller::numRobot` ..

src/CtrlGroup.c
```diff
@@ -724,6 +724,22 @@ BOOL Ros_CtrlGroup_IsRobot(CtrlGroup* ctrlGroup)
     return((ctrlGroup->groupId >= MP_R1_GID) && (ctrlGroup->groupId <= MP_R8_GID));
 }
 
+//-------------------------------------------------------------------
+// Returns TRUE if the specified group is defined as a base
+//-------------------------------------------------------------------
+BOOL Ros_CtrlGroup_IsBase(CtrlGroup const* const ctrlGroup)
+{
+    return ((ctrlGroup->groupId >= MP_B1_GID) && (ctrlGroup->groupId <= MP_B8_GID));
```
gavanderhoorn, line 732: these duplicate base functionality, but we can't use those on all platforms, hence these custom implementations.

(they will need to be updated for #19 as those controllers support fewer groups)

gavanderhoorn:
@ted-miller: I've updated the PR with the changes from the branch I mentioned in https://github.com/Yaskawa-Global/motoros2/pull/169#discussion_r1384674728. I've also included the changes to address your review commits from #169.

src/ControllerStatusIO.h
```diff
@@ -70,7 +70,7 @@ typedef struct
 {
     UINT16 interpolPeriod;                                  // Interpolation period of the controller
     int numGroup;                                           // Actual number of defined group
-    //int numRobot;                                         // Actual number of defined robot
+    int numRobot;                                           // Actual number of defined robot
```
ted-miller, line 73: I don't think this is needed. The only place that I see it being used is to check if it's `>= 1`. 

It's not possible for this software to be running on a controller w/o a robot.

src/ControllerStatusIO.c
ted-miller: Robots and stations have an explicit calibration file defined. The base axis calibration is done via robot parameters.

For this alarm to occur, we need to have multiple groups. But the base axes don't count. 

I think you should keep `numRobotsOrStations` and remove `numRobot`.

src/ControllerStatusIO.c
```diff
@@ -751,3 +763,86 @@ int Ros_Controller_GetActiveAlarmCodes(USHORT active_alarms[MAX_ALARM_COUNT + MA
...
+    // - there are groups other than base groups (so R1+R2, or R1+S1)
+    // - TF broadcast is enabled, and
+    // - none of the calibration files loaded
+    //
+    //in that case there could be Robot groups that have not been calibrated (to
+    //either other robots and/or Station groups) and this would make their TF
+    //broadcasts incorrect.
+    return (
+        //single-group systems don't have calibration files at all
+        controller->numGroup > 1 &&
```
ted-miller, line 840: This line is probably not needed, since `numRobotsOrStations > 1` exists below

gavanderhoorn: Getting back to this: I mostly added the `numRobot` and `numGroup` check because of what we discussed in https://github.com/Yaskawa-Global/motoros2/pull/169#discussion_r1353202048:

@gavanderhoorn wrote:
> Would it be correct to say that any multi-group system with non-base-axis groups would fall into the `raise-alarm-if-no-calibration-files-loaded` category?
> 
> IIUC, base axes groups are always associated with robot groups, correct? So that leaves just station axes and proper robots.

If it's impossible to have just a station, and base axes are always associated with a robot group, then just counting robots and stations would indeed be enough.

gavanderhoorn:
@ted-miller: I believe this addresses your latest feedback.

I've tested it on my controller, and it appears to work -- as in: I don't get an alarm with just my base axis, even though I have not disabled it nor have I any calibration file(s).

Could you perhaps try this on your setup?

ted-miller:
My multi-group setup is being used for testing something else. It should free up later this week.

gavanderhoorn:
> My multi-group setup is being used for testing something else. It should free up later this week.

you guys do other things beside maintaining/developing MotoROS2?

PS: running it on a single-group system would also be a good test.

ted-miller:
> you guys do other things beside maintaining/developing MotoROS2?

Of course. We're scoping out the framework for MotoROS3. What else would we be doing?

gavanderhoorn:
Friendly ping @ted-miller :)

ted-miller:
Still on the todo list. Haven't forgotten this.

gavanderhoorn:
Friendly ping @ted-miller :)

ted-miller:
> Still on the todo list. Haven't forgotten this.

Ok... I did forget this. 
I won't get to it today. But I should be able to this week.

gavanderhoorn:
I don't really have time to figure out why the tests crash the controller. I have a hunch it's stack related, but haven't been able to check it.

I would be OK with merging without the tests, as I have been able to run them by disabling the other tests. They passed.

As it's been a while, I'd like to rebase this and then merge.

gavanderhoorn:
Oh, right. CI is broken (https://github.com/Yaskawa-Global/motoros2/pull/212#issuecomment-1963952062).

Well, it builds, I promise :crossed_fingers: 

@ted-miller: ok with the merge?

ted-miller (approved):

PR #176, opened 2023-10-10
PositionMonitor: assert on unexpected base track motion types

gavanderhoorn:
As per title.

Without this, `mpBuilder` will log a warning:

```
warning: enumeration value ‘MOTION_TYPE_NOT_USED’ not handled in switch
```

which is annoying, and if configured to be more strict, will also 'break' the build.


src/ErrorHandling.h
```diff
@@ -165,6 +165,7 @@ typedef enum
     SUBCODE_CONFIGURATION_FAIL_MP_NICDATA0,
     SUBCODE_CONFIGURATION_FAIL_MP_NICDATA1,
     SUBCODE_FAIL_MP_NICDATA_INIT1,
+    SUBCODE_FAIL_INVALID_BASE_TRACK_MOTION_TYPE,
```
gavanderhoorn, line 168: Alternatively, we could define an "unexpected internal error" assertion subcode, and report those kinds of problems with that subcode.

We might otherwise quickly use up quite a few of the 255 subcodes available to us in the assertion alarm code.

The subcode text could then be used to clarify what's going on exactly, and the debug log could be used to log even more information.

ted-miller: I _really_ don't like catch-all error codes. They should only be used as a last resort for stuff like unhandled exceptions.

ted-miller (approved):

src/ErrorHandling.h
gavanderhoorn: Well .. this would be such an unhandled exception. We basically just assert and then sit there, not doing anything. It's like crashing, but without actually crashing.

ted-miller: Nah, we _handled_ it. 
Not gracefully... but it was caught and we're able to return a unique/specific error code.

gavanderhoorn: Just pointing out again that we "only" have 255 sub codes .. of which we already use about 1/4th.

ted-miller: We also have 1000 main codes. Of which we're using 6 :)

gavanderhoorn: Are those all available to MR2?

I was under the impression `8000` was the generic "user application" range, which is shared with all other M+ applications?

gavanderhoorn:
Just ran into a situation where I'm seeing `8011[63]` where I wouldn't expect it.

I'll open an issue.

Issue #184, opened 2023-10-19
Investigate how to best make use of F/T data on HC platforms


gavanderhoorn:
The HC series of robots have built-in torque sensors in all joints and come with various collaborative features built-in.

Some of the data used for those features is accessible to M+ applications: either via specific M+ APIs or fi via M-registers.

We should investigate if we can, and if so, how to, make use of that data. Either to improve existing interfaces or to make new ones available, specifically for users of HC robots.

An example improvement could be to use the values reported by the torque sensors in `JointState` messages instead of the commanded torque we currently use to populate them. We could also see whether it'd be possible to publish the estimated F/T on the TCP as a `geometry_msgs/Wrench`.

For information on M-registers used on HC platforms for this sort of data refer to *181437-1CD (HW1484764) Collaborative Operation Instructions*, specifically the *Register List* on pages 174 and 175 (10-1 and 10-2).

iydv:
I have been investigating this for a while:

1) I can obtain F/T values from registers. Unfortunately, it is not convenient to constantly send requests via service call, the values should be published on the topic.

2) MotoPlus has functions to read F/T values directly. But it doesn't work for me and I cannot find any description of these functions in MotoPlus manuals. I have tried to use the following functions:

`extern int mpFcsGetForceData(MP_FCS_ROB_ID rob_id, int coord_type, int uf_no, MP_FCS_SENS_DATA sens_data);
`
`extern int mpFcsGetSensorChannelData(MP_FCS_SENS_ID sens_id, MP_FCS_SENS_DATA sens_data);`

 they always return an error when I call them as follows:
```
MP_FCS_SENS_DATA sens_data;
bzero(sens_data, MP_FCS_AXES_NUM);
int f_ret = mpFcsGetForceData(MP_FCS_R1ID, 0, 0, sens_data);
```

and 
```
MP_FCS_SENS_DATA sens_data;
bzero(sens_data, MP_FCS_AXES_NUM);
int f_ret = mpFcsGetSensorChannelData(MP_FCS_SENS1ID, sens_data);
```
@ted-miller is there an updated manual for MotoPlus API? I have checked all available manuals on [motoman](https://www.motoman.com/en-us/service-training/product-documentation) page and did not find any description of these functions.

gavanderhoorn:
@iydv: thanks for looking into this.

> I can obtain F/T values from registers.

Are you reading the registers referenced by the documentation I mentioned in the OP?

[...]

PR #188, opened 2023-11-08
Initial implementation of HC10 F/T and ext torque publisher

iydv:
Based on discussion #184 I have implemented the first version of the force torque monitor. I have tested the code on HC10 and the data is published correctly on both topics. 

However, there are several issues which require clarification: 

1) No multi-group support. Based on my information, the collaborative robots work only on YRC1000/ YRC1000u controllers which do not support multi-groups. Can @ted-miller please clarify this? 
2) Register addresses for TCP/ joint torque values should be obtained from parameters in `ctrlGroup` similar to `Ros_CtrlGroup_GetFBServoSpeed` function. However, I have not found where I can define register addresses. Are they defined in `ParameterExtraction.yrcLib` ? How can I edit it?


gavanderhoorn:
Nice. Thanks for the PR.

I'll try to go over it in the next couple days.

Could you perhaps post some example/representative `ros2 topic echo` output?

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,180 @@
...
+    g_messages_ForceTorqueMonitor.jointExternalTorque = sensor_msgs__msg__JointState__create();
+    rosidl_runtime_c__String__assign(&g_messages_ForceTorqueMonitor.jointExternalTorque->header.frame_id, "");
+    rosidl_runtime_c__String__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->name, g_Ros_Controller.totalAxesCount); // Number of joints in message
+
+    for (int jointIndex = 0; jointIndex < g_Ros_Controller.totalAxesCount; jointIndex += 1)
+    {
+        rosidl_runtime_c__String__assign(&g_messages_ForceTorqueMonitor.jointExternalTorque->name.data[jointIndex], g_nodeConfigSettings.joint_names[jointIndex]);
+    }
+    rosidl_runtime_c__float64__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->position, g_Ros_Controller.totalAxesCount);
+    rosidl_runtime_c__float64__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->velocity, g_Ros_Controller.totalAxesCount);
```
gavanderhoorn, line 65: Quick comment: seeing as this message will only ever contain torque estimates, I would leave these fields at 0 length.

The resulting `JointState` message would then just contain values in the `effort` field, with `position` and `velocity` empty lists. That's perfectly OK.

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    //convert result to double array
+    for (int i = 0; i < 6; i += 1)
+    {
+        forceTCPValues[i] = registerValues[i];
+        forceTCPValues[i] = (forceTCPValues[i] - 10000.0) * 0.1;
+    }
+
+    //TODO: create better value assigment
```
gavanderhoorn, line 128: Would you have any ideas already on what you would consider 'better'?

src/ForceTorqueMonitor.c
iydv: I can change that

gavanderhoorn:
High-level comment: please avoid merging in `main`. I'd suggest to rebase instead. Keeps history clean, and as long as we don't have any related, concurrent development in `main` or other branches, does not seem like it violates any consistency / coherency on the Git history side.

src/ForceTorqueMonitor.c
iydv: Unfortunately, no. It is possible to just leave it like that.

gavanderhoorn: What made you add that comment there? If the size and alignment of the two arrays is exactly the same, and both are contiguous, a `memcpy(..)` might work. We do that elsewhere in MotoROS2.

gavanderhoorn: Alternatively: assign the result of the conversion to the message fields directly.

Would avoid needing the temporary `forceTCPValues` array I believe.

iydv:
yes, sorry I have deleted the merge.

src/ForceTorqueMonitor.c
iydv: > What made you add that comment there? If the size and alignment of the two arrays is exactly the same, and both are contiguous, a `memcpy(..)` might work. We do that elsewhere in MotoROS2.

Yes, I wanted to use `memcpy ` there, but I do not know if the data will be converted correctly to the `geometry_msgs__msg__Wrench` format. The size and alignment of both arrays should be the same. I can test it and see if it works

gavanderhoorn: > I do not know if the data will be converted correctly to the `geometry_msgs__msg__Wrench` format.

it should just be a `double[]` in the end. You can check the `struct` definition in the generated `.h`.

gavanderhoorn: Actually, that's not true. Apologies.

It's as follows:

```c
typedef struct geometry_msgs__msg__WrenchStamped
{
  std_msgs__msg__Header header;
  geometry_msgs__msg__Wrench wrench;
} geometry_msgs__msg__WrenchStamped;
```

and then:

```c
typedef struct geometry_msgs__msg__Wrench
{
  geometry_msgs__msg__Vector3 force;
  geometry_msgs__msg__Vector3 torque;
} geometry_msgs__msg__Wrench;
```

then:

```c
typedef struct geometry_msgs__msg__Vector3
{
  double x;
  double y;
  double z;
} geometry_msgs__msg__Vector3;
```

so ideally, it will be contiguous in memory as 6 `double`s.

But I'm not sure that's guaranteed by the C99 standard.

iydv: ok. In this case, I will just assign values directly like before but without `forceTCPValues`

iydv:
> Could you perhaps post some example/representative `ros2 topic echo` output?

here are examples of topic outputs (I will change later `joint_external_torque` to output empty position and velocity) :
```
ros2 topic echo /tcp_force_torque 
header:
  stamp:
    sec: 1699449961
    nanosec: 592319543
  frame_id: motoros2_r1/base
wrench:
  force:
    x: -22.6
    y: 4.6000000000000005
    z: 23.1
  torque:
    x: -2.2
    y: -9.1
    z: -0.4
```

```
ros2 topic echo /joint_external_torque 
header:
  stamp:
    sec: 1699450224
    nanosec: 656734054
  frame_id: ''
name:
- joint_1_s
- joint_2_l
- joint_3_u
- joint_4_r
- joint_5_b
- joint_6_t
position:
- 0.0
- 0.0
- 0.0
- 0.0
- 0.0
- 0.0
velocity:
- 0.0
- 0.0
- 0.0
- 0.0
- 0.0
- 0.0
effort:
- -1.5
- -20.5
- 11.4
- -2.5
- 0.1
- -0.30000000000000004
```
To verify the values you can use teach pad: Safety Func. -> ForceMonitor 
IMG_20231108_142709413

Unfortunately, the F/T values oscillates a lot, so you might not see exact values on teach pad

gavanderhoorn:
Those are some serious nrs, are you pushing on the robot?

> Unfortunately, the F/T values oscillates a lot, so you might not see exact values on teach pad

yes. I expect some filtering to be beneficial/necessary on the ROS side in actual applications.

iydv:
`Those are some serious nrs, are you pushing on the robot?`
I did not select correct tool in settings with proper weight, so the values are off since it uses tool weight for calculation

ted-miller:
> No multi-group support. Based on my information, the collaborative robots work only on YRC1000/ YRC1000u controllers which do not support multi-groups. Can @ted-miller please clarify this?

The YRC1000 does support multi-group. But we don't do multi-HC configurations.

> Register addresses for TCP/ joint torque values should be obtained from parameters in ctrlGroup similar to Ros_CtrlGroup_GetFBServoSpeed function. However, I have not found where I can define register addresses. Are they defined in ParameterExtraction.yrcLib ? How can I edit it?

The registers are hardcoded. There is now way to change them.
[register list.pdf](https://github.com/Yaskawa-Global/motoros2/files/13300071/register.list.pdf)

I would suggest defining these as contants/defines in a header file

gavanderhoorn:
@iydv: have you had any chance to look at some of things commented on above?

Just a friendly ping again.

iydv:
> @iydv: have you had any chance to look at some of things commented on above?

I was waiting for full review, in case a major code rework is required.  

If general code structure is fine, I can integrate changes from above today

iydv:
> If general code structure is fine, I can integrate changes from above.
done

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+static void Ros_ForceTorqueMonitor_Initialize_Publishers(rmw_qos_profile_t const* const qos_profile)
+{
+    //create a publisher for TCP force torque
+    rcl_ret_t ret = rclc_publisher_init(
+        &g_publishers_ForceTorqueMonitor.forceTorqueTCP,
+        &g_microRosNodeInfo.node,
+        ROSIDL_GET_MSG_TYPE_SUPPORT(geometry_msgs, msg, WrenchStamped),
+        TOPIC_NAME_FORCE_TORQUE_TCP,
+        qos_profile);
+    motoRosAssert(ret == RCL_RET_OK, SUBCODE_FAIL_CREATE_PUBLISHER_JOINT_STATE_ALL);
```
ted-miller, line 35: I like to have a unique error code for every single alarm. That way, the numerical code identifies the exact line line of code which raised the alarm. There's no speculation about possible causes.

Please add a new entry to the end of the `ALARM_ASSERTION_FAIL_SUBCODE` enum. (If you add to the middle of the enum, then it throws off the existing troubleshooting documentation.) Then we'll also need to add this entry to `troubleshooting.md`.

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+    Ros_Debug_BroadcastMsg("Initializing ForceTorqueMonitor publishers");
+
+    //==================================
+    //create publishers for TCP force and external joint torque
+    const rmw_qos_profile_t* qos_profile_js = Ros_ConfigFile_To_Rmw_Qos_Profile(g_nodeConfigSettings.qos_joint_states);
+    Ros_ForceTorqueMonitor_Initialize_Publishers(qos_profile_js);
+
+}
+
+static void Ros_ForceTorqueMonitor_Initialize_Publishers(rmw_qos_profile_t const* const qos_profile)
```
ted-miller, line 26: Can you add a new entry in the config file parsing to enable/disable this publisher? I think this topic should default to OFF. HC robots are a small minority of Motoman robots. So when this data is not available, there's no need to add a topic which would increase publish latency.

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
+// ForceTorqueMonitor.c
+
+// SPDX-FileCopyrightText: 2022-2023, Yaskawa America, Inc.
+// SPDX-FileCopyrightText: 2022-2023, Delft University of Technology
```
ted-miller, line 4: I'm not really sure how the licensing stuff works. But I assume RobotCloud should be added here.

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+        ROSIDL_GET_MSG_TYPE_SUPPORT(geometry_msgs, msg, WrenchStamped),
+        TOPIC_NAME_FORCE_TORQUE_TCP,
+        qos_profile);
+    motoRosAssert(ret == RCL_RET_OK, SUBCODE_FAIL_CREATE_PUBLISHER_JOINT_STATE_ALL);
+
+    //create message for TCP force torque
+    g_messages_ForceTorqueMonitor.forceTorqueTCP = geometry_msgs__msg__WrenchStamped__create();
+
+    char formatBuffer[MAX_JOINT_NAME_LENGTH];
+    snprintf(formatBuffer, MAX_TF_FRAME_NAME_LENGTH, "%sr%d/base", g_nodeConfigSettings.tf_frame_prefix, 1);
```
ted-miller, line 41: I don't understand the naming convention here. Could you elaborate?

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+    ret = rclc_publisher_init(
+        &g_publishers_ForceTorqueMonitor.jointExternalTorque,
+        &g_microRosNodeInfo.node,
+        ROSIDL_GET_MSG_TYPE_SUPPORT(sensor_msgs, msg, JointState),
+        TOPIC_NAME_JOINT_EXTERNAL_TORQUE,
+        qos_profile);
+    motoRosAssert(ret == RCL_RET_OK, SUBCODE_FAIL_CREATE_PUBLISHER_JOINT_STATE_ALL);
+
+    //create message for joint external torque
+    g_messages_ForceTorqueMonitor.jointExternalTorque = sensor_msgs__msg__JointState__create();
```
ted-miller, line 56: The existing `/joint_states` topic already publishes effort. Would it make sense to replace that field with these values? Or is there value in knowing the difference between internal/external torque?

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+    for (int jointIndex = 0; jointIndex < g_Ros_Controller.totalAxesCount; jointIndex += 1)
+    {
+        rosidl_runtime_c__String__assign(&g_messages_ForceTorqueMonitor.jointExternalTorque->name.data[jointIndex], g_nodeConfigSettings.joint_names[jointIndex]);
+    }
+    rosidl_runtime_c__float64__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->position, 0);
+    rosidl_runtime_c__float64__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->velocity, 0);
+    rosidl_runtime_c__float64__Sequence__init(&g_messages_ForceTorqueMonitor.jointExternalTorque->effort, g_Ros_Controller.totalAxesCount);
+}
+
+void Ros_ForceTorqueMonitor_Cleanup()
```
ted-miller, line 69: I'd suggest adding the `MOTOROS2_MEM_TRACE` macros to the Initialize and Cleanup routines. We've already got known memory leak in the app. So it'd be helpful to know we're not adding any more. (See PositionMonitor.c for examples.)

src/ForceTorqueMonitor.c
```diff
@@ -0,0 +1,164 @@
...
+    //Read TCP force data from registers
+    status = mpReadIO(registerInfo, registerValues, 6);
+
+    if (status != OK)
+    {
+        Ros_Debug_BroadcastMsg("Failed to get TCP force torque data: %u", status);
+    }
+
+    //convert register data to proper format and assign to corresponding wrench fields
+    g_messages_ForceTorqueMonitor.forceTorqueTCP->wrench.force.x = (registerValues[0] - 10000.0) * 0.1;
```
ted-miller, line 121: Why is 10000 being subtracted? 
Is this for scaling the units? If so, that should be division.

ted-miller (changes requested):
Overall, I think the general structure is good. I just have a few comments on the details.

src/ForceTorqueMonitor.c
iydv: The formula taken directly from official manual 
image

src/ForceTorqueMonitor.c
iydv: In the header we have to set a correct frame ID, which is `base`. Additionally, if the tf prefix is set in config file, I also add it to the frame ID. And since the publisher always process data for group 1, the name will be `prefix_r1/base`, which should correspond to the name of published base tf frame for group 1.

iydv:
@ted-miller thank you for feedback, I will try to integrate all changes.

src/ForceTorqueMonitor.c
destogl: @ted-miller I can also confirm that the uint16 values come from M-registers and those are represented with `10000` for `0`. Therefore, this calculation is also correct for our HC20SDTP robot.

gavanderhoorn: that `10000` is in the documentation (also quoted by @iydv), so it's clear it needs to be included in the conversion.

destogl:
@gavanderhoorn or @iydv is it possible to get a compiled version for YRC1000m for testing this?

I have a compilation issue, as VS 2017 is hard to get and then end up in the dependencies hell already for more then a day...

gavanderhoorn:
VS 2017 isn't a requirement: I use VS 2022 myself. Do you have the MotoPlus SDK?

destogl:
@gavanderhoorn thank you for this! It works great on our HC20SDTP model and directly out of the box. Thank you for this!

I have managed to set up the VS for the compilation, but the key needs a few more days to come so we can also compile this.

We will report our experiences, so you guys have more feedback, and let's hope this lands soon into release.

destogl:
Hi all, I can confirm that we were using this feature happily and successful over multiple months. Can we contribute somehow to get this merged? @ted-miller and @gavanderhoorn

ted-miller:
Hi @destogl 
I left a number of comments/requests in a [previous review](https://github.com/Yaskawa-Global/motoros2/pull/188#pullrequestreview-1762749538). If you're able to address these topics, it would be very helpful.

IMO, the most important issue is adding an entry to the yaml-configuration-file to explicitly enable this topic.

gavanderhoorn:
It would be really great if we could auto-detect the presence of the sensor / PFL and enable the additional publishers based on that. An additional config entry to override that detection would be good though.

ted-miller:
There's this:
https://github.com/Yaskawa-Global/motoros2/blob/48dd4a1758fec1840d076b1914611670e8e45040/src/ControllerStatusIO.c#L51-L55

gavanderhoorn:
That could work. I'm still not a fan of assuming a PFL-enabled controller *always* has the whole FT-infrastructure setup, accessible and working (https://github.com/Yaskawa-Global/motoros2/issues/184#issuecomment-1781077543) -- as it also prevents us from publishing FT data if there is a sensor but no PFL, but for now it seems like calling `GP_isPflEnabled()` would indeed be OK for this first implementation.

src/ForceTorqueMonitor.c
gavanderhoorn: This should be one of the places where we'd check whether the controller actually has PFL enabled (and we can thus assume FT data is available). See the discussion below.

src/ForceTorqueMonitor.c
gavanderhoorn: This would depend on whether @iydv chooses to or not. But it might be nice to do.

src/ForceTorqueMonitor.c
gavanderhoorn: We should perhaps add a comment explaining which frame the FT data is in -- on the Yaskawa side -- and how we 'map' that to ROS TF frames.

src/ForceTorqueMonitor.c
gavanderhoorn: > Or is there value in knowing the difference between internal/external torque?

yes.

The `effort` field in the msgs on `/joint_states` is supposed to represent effort applied by the joints to their surroundings. The FT data we're publishing here is something else.

Fadi-AlMasalmah:
> > Could you perhaps post some example/representative `ros2 topic echo` output?
> 
> here are examples of topic outputs (I will change later `joint_external_torque` to output empty position and velocity) :
> 
> `ros2 topic echo /tcp_force_torque`:
> 
> Click to expand
> ```yaml
> header:
>   stamp:
>     sec: 1699449961
>     nanosec: 592319543
>   frame_id: motoros2_r1/base
> wrench:
>   force:
>     x: -22.6
>     y: 4.6000000000000005
>     z: 23.1
>   torque:
>     x: -2.2
>     y: -9.1
>     z: -0.4
> ```

Hello @gavanderhoorn @iydv, thank you for this PR, I tried it with Yaskawa HC20DTP and it works well, however, the values don't make much sense. With no motion, no tools and no contacts, the robot outputs values that goes up to +-40N depending on the joints positions, especially when far from the calibrated Torque-Origin-Position. I realize that this is not a problem from the code here, but are these values even usable? Is it just my robot or am I missing some step? If you have used it in some application, is there anything I can do to enhance the results?
Thanks a lot!

ted-miller:
hello @Fadi-AlMasalmah 
I'm not sure. Please give us and @jimmy-mcelwain a few days to test this setup. I think we have access to an HC20 to see what the expected behavior is. We'll try the existing code and see if improvements are needed.

Issue #190, opened 2023-11-16
mpExRcsIncrementMove is constantly sending zero pulse increments after each motion


iydv:
After testing #157 I have discovered another issue: `mpExRcsIncrementMove` with zero pulse increments is constantly being sent after each movement. Such behavior did not happen before the update.  

I have done following experiment to test it:

1) Test 1: checkout bd95e65679d582b13b3887e5277d34f7406db69a and add debug message before `mpExRcsIncrementMove` [here](https://github.com/Yaskawa-Global/motoros2/blob/bd95e65679d582b13b3887e5277d34f7406db69a/src/MotionControl.c#L847)

`Ros_Debug_BroadcastMsg("Send mpExRcsIncrementMove tool: %d, pos data: %d %d %d %d %d %d", moveData.grp_pos_info[0].pos_tag.data[2], moveData.grp_pos_info[0].pos[0], moveData.grp_pos_info[0].pos[1], moveData.grp_pos_info[0].pos[2], moveData.grp_pos_info[0].pos[3], moveData.grp_pos_info[0].pos[4], moveData.grp_pos_info[0].pos[5]);`

Select tool 4 and sent simple trajectory with single point. Resulting log file [test1.txt](https://github.com/Yaskawa-Global/motoros2/files/13378122/test1.txt) clearly indicate that when the trajectory is complete `mpExRcsIncrementMove` is no longer being sent to the controller. 

2) Test 2: checkout main and add same debug message [here](https://github.com/Yaskawa-Global/motoros2/blob/464758bb0337a7b567e450af273fcc2a0ab8b849/src/MotionControl.c#L1048). 

Similar to Test 1, I select tool 4 and move robot using simple trajectory. Resulting log file 
[test2.txt](https://github.com/Yaskawa-Global/motoros2/files/13378161/test2.txt) shows that after trajectory is complete, `mpExRcsIncrementMove` with zero pulse increments is being constantly sent to the robot. 

The main problem for my application is that after the motion is finished, empty `mpExRcsIncrementMove` changes tool from 4 to 0 and I cannot obtained correct Force/ Torque data. Additionally, I do not think it is normal that zero pulses are constantly being sent to controller. 

[...]

PR #191, opened 2023-11-16
reset hasUnprocessedData to prevent zero pulses being sent by mpExRcsIncrementMove after trajectory is finished

iydv:
Fix for #190 . Currently, `hasUnprocessedData` is never reset after the queue is fully processed inside `Ros_MotionControl_IncMoveLoopStart`, which means that empty queue is constantly being read and zero pulse increment is constantly being sent by `mpExRcsIncrementMove`

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -932,6 +932,7 @@ void Ros_MotionControl_IncMoveLoopStart() //<-- IP_CLK priority task
                 ctrlGrpData.sCtrlGrp = g_Ros_Controller.ctrlGroups[i]->groupId;
                 mpGetPulsePos(&ctrlGrpData, &pulsePosData);
                 isMissingPulse = FALSE;
+                hasUnprocessedData = FALSE;
                 for (axis = 0; axis < MP_GRP_AXES_NUM; axis++)
                 {
                     // Check how many pulses we processed from last increment


iydv:
I have conducted a test similar to #190  and resulting debug log shows that FSU speed handling part still works fine but empty `mpExRcsIncrementMove` is no longer sent when trajectory is finished.
[20231116T135527_motoros2_debug_log.txt](https://github.com/Yaskawa-Global/motoros2/files/13378508/20231116T135527_motoros2_debug_log.txt)

ted-miller:
Hi @iydv
Good catch. I definitely missed that during testing. 

This won't work for a multi-group system. The `hasUnprocessedData` would get cleared when calculating for R2. 

I think we can simply move the `hasUnprocessedData = FALSE;` up a few lines so that it outside of the for-each-group loop.

What do you think?

iydv:
Yes, you are right, I forgot about the multi-group system. I guess it should be moved [here ](https://github.com/iydv/motoros2/blob/ba8c0cdae3f705819650e6503a861ec9e2b335cb/src/MotionControl.c#L913).

ted-miller:
Looks good to me. My system is currently configured for single-group. But I should be able to reconfigure it on Wednesday for a quick test.

ted-miller (approved):
Test was ok. Thanks for the fix.

PR #205, opened 2024-01-30
Bump version as 0.1.3pre

gavanderhoorn:
As per subject.

Facilitate distinguishing from-source builds from binary releases.


gavanderhoorn:
Note: I opted to just set it to `0.1.3pre`, even though we might actually release `0.2.0` later.

We'll adjust/correct in a later PR.

ted-miller (approved):

PR #209, opened 2024-02-01
ci: config md lint for our 'strong style'

gavanderhoorn:
As per title.

Seems this check was added 'recently' to `mdlint` and we got that version in #202.

Not sure why this didn't trigger CI for that PR, but let's just fix it.


gavanderhoorn:
Waiting to see if this makes `mdlint` happy.

PR #212, opened 2024-02-15
Issue 203 cycle switch mode

### LINKED ISSUE #203: Cycle-switch-mode during setup procedure
Re: https://github.com/Yaskawa-Global/motoros2/discussions/201#discussioncomment-8224715

Need to add this to setup procedure. Perhaps we can also detect this in real time via CIO.

(I vaguely recall this being an issue in the past. But I can't find anything about it in the repos.)

yai-rosejo:
Fixes #203 
Added in a CIO read to check for continuous cycle mode being set

### DIFF
--- a/src/ControllerStatusIO.c
+++ b/src/ControllerStatusIO.c
@@ -428,6 +428,10 @@ MotionNotReadyCode Ros_Controller_GetNotReadySubcode()
     if(!Ros_Controller_IsPlay())
         return MOTION_NOT_READY_NOT_PLAY;
 
+    // Check if in continuous cycle mode (Here due to being checked before starting servor power)
+    if (!Ros_Controller_IsContinuousCycle())
+        return MOTION_NOT_READY_NOT_CONT_CYCLE_MODE;
+
 #ifndef DUMMY_SERVO_MODE
     // Check remote
     if(!Ros_Controller_IsRemote())
@@ -460,10 +464,6 @@ MotionNotReadyCode Ros_Controller_GetNotReadySubcode()
     // Check ready I/O signal (should confirm wait)
     if(!Ros_Controller_IsWaitingRos())
         return MOTION_NOT_READY_WAITING_ROS;
-
-    // Check if in continuous cycle mode
-    if (!Ros_Controller_IsContinuousCycle())
-        return MOTION_NOT_READY_NOT_CONT_CYCLE_MODE;
 
     return MOTION_READY;
 }
--- a/src/ControllerStatusIO.c
+++ b/src/ControllerStatusIO.c
@@ -461,6 +461,10 @@ MotionNotReadyCode Ros_Controller_GetNotReadySubcode()
     if(!Ros_Controller_IsWaitingRos())
         return MOTION_NOT_READY_WAITING_ROS;
 
+    // Check if in continuous cycle mode
+    if (!Ros_Controller_IsContinuousCycle())
+        return MOTION_NOT_READY_NOT_CONT_CYCLE_MODE;
+
     return MOTION_READY;
 }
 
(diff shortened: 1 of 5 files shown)


ted-miller:
Depends on https://github.com/Yaskawa-Global/motoros2_interfaces/pull/10

Code looks good to me, and I saw yai-rosejo test it. 

@gavanderhoorn, do you have any concerns about message version compatiblity?

gavanderhoorn:
> @gavanderhoorn, do you have any concerns about message version compatiblity?

no, not in this case.

We're not embedding the `MotionReadyEnum` msg itself, we just use it to store "constants".

gavanderhoorn:
High-level comment: should we perhaps also post a (non-blocking) alarm? That would make it *really* obvious something is misconfigured (at least for use with MotoROS2).

ted-miller:
> should we perhaps also post a (non-blocking) alarm

I don't think so. I don't think it's any different than the other motion-not-ready codes.

The status message indicates `not-ready` and the user will get an appropriate error message if they attempt either of the `start` services.

What would be your argument for an alarm?

ted-miller (approved):
I think this is in good shape. Merging.

gavanderhoorn:
I'm almost certain this PR broke CI builds.

The [msbuild filter](https://github.com/Yaskawa-Global/motoros2/actions/runs/8023609181/job/21920407657#step:3:35) did determine the PR needed a full build, but that didn't happen because @yai-rosejo [submitted from a fork](https://github.com/Yaskawa-Global/motoros2/blob/7574bd2d849efda1ad631c07bdc059b585db5c81/.github/workflows/ci_msbuild.yaml#L41-L45).

I'm not sure why [the merge](https://github.com/Yaskawa-Global/motoros2/pull/212#event-11882390095) by @ted-miller didn't trigger a full build either.

In any case: CI is probably broken until `libmicroros` sees a new build *and* release, because `motoros2_interfaces__msg__MotionReadyEnum__NOT_READY_NOT_CONT_CYCLE_MODE_STR` and `motoros2_interfaces__msg__MotionReadyEnum__NOT_READY_NOT_CONT_CYCLE_MODE` are not defined in the current release(s) but are used here:

https://github.com/Yaskawa-Global/motoros2/blob/7574bd2d849efda1ad631c07bdc059b585db5c81/src/ErrorHandling.c#L72-L73

I'll trigger a manual full build (of MotoROS2) and verify.

gavanderhoorn:
See https://github.com/Yaskawa-Global/motoros2/actions/runs/8048602951.

It indeed fails due to the undeclared constants.

gavanderhoorn:
@ted-miller wrote:
> @gavanderhoorn wrote:
> > should we perhaps also post a (non-blocking) alarm
> 
> I don't think so. I don't think it's any different than the other motion-not-ready codes.
> 
> The status message indicates `not-ready` and the user will get an appropriate error message if they attempt either of the `start` services.
> 
> What would be your argument for an alarm?

mostly just to draw extra attention to it.

The cycle mode is a rather invisible setting I believe, and perhaps also not too obvious where to change.

This in contrast to most (all?) other failure reasons we check for when starting a traj mode.

With an alarm, we could also more easily document a solution for users, as we'd have a specific number to refer to.

But we don't really need to do anything here. It was just a suggestion.

gavanderhoorn:
> I did notice that the CI didn't automatically run, as I expected. Normally, I've seen it just say either success or failure. 

It should also tell you *why* it failed, but you'd have to go to the specific run and inspect the results.

If/when we get an improved `mpBuilder`, GitHub should show annotations with compiler warnings/errors directly associated with the offending line(s) in the PR itself.

> In this case, I had to manually instantiate it. I'm not sure if this is relevant.

what do you mean by "had to manually instantiate it"?

> > `motoros2_interfaces__msg__MotionReadyEnum__NOT_READY_NOT_CONT_CYCLE_MODE` are not defined in the current release(s) but are used
> 
> This is something that I didn't consider. How do we handle a PR that require changes to the libmicroros? I guess we would need to make release of that before handling the PR. John and I have been working from source, and not considering this scenario for external users.

Current `HEAD` of MotoROS2 can't be built with the M+ `libmicroros` distribution `.zip`s we have made available until we release updated distribution `.zip`s. From-source builds are essentially broken right now.

This ties into the discussion around how we version, build and distribute M+ `libmicroros`.

The sort of breakage we discovered here being one of the reasons we're looking into this.

gavanderhoorn:
> CI didn't run automatically. There was a button to make it run. I clicked it.

ah, that's probably because the PR was submitted by a new/external contributor. We've configured the repository such that all Action workflows must then be *approved* before they run. It's not the cause of what we're discussing here.

> Do you think it's worth reverting this PR in the short-term? (I'm not personally aware of any external users building from source)

Technically we only 'release' M+ `libmicroros` with each [release of MotoROS2](https://github.com/Yaskawa-Global/motoros2/releases), so you could say that we're allowed to break `main`.

For now, users building from source can't build `HEAD`, only `0.1.2` or earlier.

Issue #114, opened 2023-08-07
`/start_traj_mode` failed incorrectly


ted-miller:
I was in an error state created by an estop. I tried to start traj mode. It failed due to the error state. But, the job did start and the drives came on. 

Eventually servo-sleep mode kicked in with the job still running. 

I then call `/reset_error` to clear the previous error state.

When I call start_traj_mode, the servos woke back up and the job was restarted. But the service returned `result_code=motoros2_interfaces.msg.MotionReadyEnum(value=109), message='INFORM job is not on a WAIT command. Check the format of INIT_ROS')`

I'm wondering if it was due to the servo-sleep or my error-reset situation.

A second call to /start_traj_mode worked ok.

ted-miller:
> A second call to /start_traj_mode worked ok.

Well maybe not. It indicated success. But when I tried to submit a goal, I got `control_msgs.action.FollowJointTrajectory_Result(error_code=-100208, error_string='Must call start_traj_mode service.')`

yai-rosejo:
Found a sequence that makes this message to appear and is repeatable:
e-stop -> turn off e-stop -> send reset_error -> send start_traj_mode -> wait till energy save mode turns on -> send fjt command -> inform error
```
[2024-02-16 19:34:21.624265] [192.168.1.31:65119]: reset: attempting to reset controller
[2024-02-16 19:34:21.624265] [192.168.1.31:65119]: reset: success
[2024-02-16 19:34:30.722604] [192.168.1.31:65119]: Ros_MotionControl_StartMotionMode: enter
[2024-02-16 19:34:31.211004] [192.168.1.31:65119]: Robot job is ready for ROS commands.
[2024-02-16 19:34:31.311004] [192.168.1.31:65119]: Robot job is ready for ROS commands.
[2024-02-16 19:34:31.367204] [192.168.1.31:65119]: Ros_MotionControl_StartMotionMode: exit
[2024-02-16 19:34:31.367204] [192.168.1.31:65119]: Ros_MotionControl_ActiveTrajMode = 1
[2024-02-16 19:34:31.367204] [192.168.1.31:65119]: Ros_ServiceStartTrajMode_Trigger: activated
[2024-02-16 19:35:54.783605] [192.168.1.31:65119]: Trajectory contains 61 points
[2024-02-16 19:35:54.783805] [192.168.1.31:65119]: Energy saving function is active. Re-enabling the robot.
[...]

PR #214, opened 2024-02-20
Issue 114 Start Traj Mode failure exiting Eco Mode

yai-rosejo:
#114 Covers when /start_traj_mode fails while eco mode is set. While testing out /start_traj_mode I found that fjt also failed when in eco mode. By increasing the fjt delay and moving it into startMotionMode this got rid of eco issues with fjt and potentially resolves the start_traj_mode issues.

Tested on controller with no arm and controller with arm.
Test procedures:
1. Run microros
2. Hit e-stop
3. Reset e-stop
4. Send: ros2 service call /reset_error motoros2_interfaces/srv/ResetError
5. Send: ros2 service call /start_traj_mode motoros2_interfaces/srv/StartTrajMode
6. Wait until energy-saving mode turns on
7. Send fjt command

### DIFF
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -190,11 +190,6 @@ rcl_ret_t Ros_ActionServer_FJT_Goal_Received(rclc_action_goal_handle_t* goal_han
 
         Ros_Debug_BroadcastMsg(responseMsg->message.data);
 
-        //Give time for the controller to recognize that the INFORM cursor is sitting on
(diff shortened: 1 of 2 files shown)


ted-miller:
I'm still not clear exactly _why_ increasing the delay is necessary when eco-mode is involved. 

But I do like that the delay is moved from the FJT handling to the more generalized `StartMotionMode`. And the testing was pretty clear.

ted-miller (approved):

ted-miller:
waiting on ci

PR #215, opened 2024-02-26
ci: don't prevent manually triggered msbuild jobs

gavanderhoorn:
The conditional on the `msbuild` job prevented it from running manually triggered `workflow_dispatch` jobs.

Extend the conditional by explicitly checking for manually triggered builds and skipping the changed-files filter in those cases.

### DIFF
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -38,11 +38,17 @@ jobs:
   msbuild:
     runs-on: windows-2022
     needs: msbuild_filter
-    # only run this job if source files were changed,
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Test build [here](https://github.com/Yaskawa-Global/motoros2/actions/runs/8048602951).

Seems to work as intended (as in: `msbuild_filter` sees no changes, but because it's a `workflow_dispatch` run, the `msbuild` job still runs).

gavanderhoorn:
Friendly ping @ted-miller.

ted-miller (approved):

PR #220, opened 2024-03-07
readme: don't start Agent as root

gavanderhoorn:
As per title.

The Agent does not have to be run as `root` on Linux systems. The example command-line we show by default will do just that.

By adding `--user=$(id -u):$(id -g)` we ask Docker to start the Agent binary as the same user that runs the command, which is much better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,7 @@ docker run \
   -it \
   --rm \
   --net=host \
+  --user=$(id -u):$(id -g) \
   microros/micro-ros-agent:humble \
     udp4 \
     --port 8888


ted-miller:
I don't understand why this is a concern. Since it's running in an isolated container, why does it matter?

gavanderhoorn:
Docker containers are typically started using the `docker` cli. This talks to the Docker daemon via a socket. The Docker daemon typically runs with super user privileges.

Running processes inside a Docker container as `root` makes it 'easier' to escape the container and then use those privileges to attack the host system.

If there are any security issues with the Agent, this would complicate exploiting them just a little bit.

It's a minor change in a command users typically copy-paste so should not really change UX.

ted-miller (approved):

PR #226, opened 2024-03-15
Support for FS100 controllers

gavanderhoorn:
This is for #19.

A draft currently, but first steps towards FS100 support.

PR #227, opened 2024-03-15
Support for FS100 controllers

gavanderhoorn:
This is for #19 (and a resubmit of #226 from a branch here on the main repository).

A draft currently, but first steps towards FS100 support.


gavanderhoorn:
CI will fail as we currently have no release of M+ `libmicroros` for FS100.

We'll deal with that later.

gavanderhoorn:
@ted-miller: after I rebased this branch on `main` my FS100 now complains because the binary includes #212 and I can't seem to set the cycle mode to *continuous*.

There's only `AUTO`, `CYCLE` and `STEP`. If I set it to `AUTO`, `SOUT#035` becomes high, but as soon as I switch the TP to `Remote`, cycle gets reset to `CYCLE`.

Any ideas?

gavanderhoorn:
Something to decide: the current work-around/hack mentioned in https://github.com/Yaskawa-Global/motoros2/issues/19#issuecomment-1930072900 only works for Agents running on little-endian machines.

I expect that to be the typical setup for use with MotoROS2 (ie: `amd64` based ROS PC running the Agent), *but* I could see someone trying to use an RPi or similar `aarch64` or `aarch32` platform to delegate the Agent to.

That would not work for the FS100.

Personally, I feel we could document this, but:

1. the asymmetry with the other controllers is not nice (™)
2. MotoROS1 had no such limitation
3. it's likely such a note in our documentation will be overlooked, leading to users posting issues and/or opening discussions about their installation not working

Given the fact the FS100 is a legacy platform at this point this may all be acceptable, but we need to decide how we want to approach this.

ted-miller:
> There's only AUTO, CYCLE and STEP. If I set it to AUTO, SOUT#035 becomes high, but as soon as I switch the TP to Remote, cycle gets reset to CYCLE.
> Any ideas?

Go to `Setup` > `Operate Cond`. Then change the `CYCLE SWITCH IN REMOTE MODE` setting.

> Given the fact the FS100 is a legacy platform at this point this may all be acceptable, but we need to decide how we want to approach this

I'm of this opinion. Just document the limitation and move on.

There will be a limited set of users using MR2 on FS100 and a _much_ smaller set of those using a bigendian cpu.

ted-miller:
@gavanderhoorn Beyond the potential endianness issue, what concerns do you have? Given the success of the previous testing, I'm not really sure why this is a 'draft' PR. 

My biggest concern is that we haven't validated on a non-dummy controller. I'm going to try to scrounge up one of the old educational carts.

Now that the rebase is complete, I'm happy to test all topics/services again. Please provide a binary of the libmicroros. (I'll also validate the VS settings for building.)

ted-miller:
I've located an FS100 with an arm. It should be here tomorrow afternoon.

gavanderhoorn:
It's partly a draft because I haven't actually moved anything, neither real hw nor virtual.

It also lacks documentation updates, and I still need to verify which parameters need to be set to which values.

ted-miller:
> I still need to verify which parameters need to be set to which values.

From my personal notes regarding the FS100:

S1D320 = 1  //Allow ".out" files to be loaded
(I think this gets set with the MOTOMAN DRIVER option function)

I don't think that S2C1402 exists.

Everything else should be covered by our existing setup-instructions.

ted-miller:
@gavanderhoorn 
Can you provide the latest libmicroros or .out?
Or would you prefer I validate using https://github.com/Yaskawa-Global/motoros2/files/14606238/20240314_mr2_fs1_h_9.9.9-FS100-alpha46.zip ?

gavanderhoorn:
> Or would you prefer I validate using [Yaskawa-Global/motoros2/files/14606238/20240314_mr2_fs1_h_9.9.9-FS100-alpha46.zip](https://github.com/Yaskawa-Global/motoros2/files/14606238/20240314_mr2_fs1_h_9.9.9-FS100-alpha46.zip) ?

that wouldn't work, as it has `DUMMY_SERVO_MODE` enabled.

I'll build a fresh alpha.

gavanderhoorn:
@ted-miller: here you go: [20240327_mr2_fs1_h_9.9.9-FS100-alpha50.zip](https://github.com/Yaskawa-Global/motoros2/files/14789640/20240327_mr2_fs1_h_9.9.9-FS100-alpha50.zip).

ted-miller:
I'll do some more thorough testing next week. But doing a quick test, I was able to get the robot to move with the point-queue service.

Other quick notes:
- `effort` data in joint_states is garbage data
- `velocity` fields in joint_states are always zero
- The feedback position is 'laggy'. If I jog the robot, the position immediately updates. But when I release the jogging button, the position coasts for about a second before coming to a stop. I see this same behavior on the pendant, so it's not MR2. But it is something different with the FS100.

gavanderhoorn:
> I'll do some more thorough testing next week. But doing a quick test, I was able to get the robot to move with the point-queue service.

:+1: 

> Other quick notes:
> 
> * `effort` data in joint_states is garbage data

hm. Is this a known issue? I'm running `FS4.00` here.

> * `velocity` fields in joint_states are always zero

Probably due to the M-register for velocity feedback not being supported on FS100? At least that's what logged on my controller when it boots up MotoROS2.

> * The feedback position is 'laggy'. If I jog the robot, the position immediately updates. But when I release the jogging button, the position coasts for about a second before coming to a stop. I see this same behavior on the pendant, so it's not MR2. But it is something different with the FS100.

I believe I've seen something similar, but wasn't sure whether it was caused by the fact I don't have any manipulator connected.

ted-miller:
> > effort data in joint_states is garbage data
>
> hm. Is this a known issue? I'm running FS4.00 here.

No. The M+ manual says the API should work. But I looked at MR1 code and see that `Ros_CtrlGroup_GetTorque` never actually gets called anywhere. So, I don't have any known examples of the API working on the FS100.


> > velocity fields in joint_states are always zero
> 
> Probably due to the M-register for velocity feedback not being supported on FS100? At least that's what logged on my controller when it boots up MotoROS2.

Yup. That'll do it.

image

I guess we need to convert `Ros_CtrlGroup_GetFBServoSpeed` to use `mpSvsGetVelTrqFb`. But that's assuming `mpSvsGetVelTrqFb` actually works... (It's what is being used for `effort` data.)

gavanderhoorn:
> I guess we need to convert `Ros_CtrlGroup_GetFBServoSpeed` to use `mpSvsGetVelTrqFb`. But that's assuming `mpSvsGetVelTrqFb` actually works... (It's what is being used for effort data.)

I believe we have an open issue on the internal tracker (`#547`) where in the end I commented:

> According to yourself ([here](https://github.com/ros-industrial/motoman/blob/63e624215180110474e0d89289f7434eb961d3b0/motoman_driver/MotoPlus/CtrlGroup.c#L342-L435)), M-registers are better ;)

I was confused by the comments in that issue as you also suggested changing to `mpSvsGetVelTrqFb(..)`, but that was/is already used. At least in MotoROS1.

ted-miller:
_(Not sure if this note is a repost. I know I had something typed up, but I think I closed the window prematurely.)_

I made a small app that just polls `mpSvsGetVelTrqFb` continuously.

For velocity:
With servos off and the brakes engaged, I'm getting random +/- 1248 on one or more axes every few cycles. But while jogging, the numbers look reasonable.

For effort:
I don't know what to think of the numbers being displayed. I need to do more testing.
The pendant only displays `percent of rated max` torque, so I don't have a good correlation between this and Nm. 

But it seems like there's a lot of fluctuation on multiple axes for even a very simple motion, such as jogging the S axis. Also, the resting-position seems unusual. If I jog S+ the value spikes. When motion is complete, the value settles around 15000. But then if I jog S- and stop, the value settles around -10000. I understand the sign inversion, but I don't get why the absolute value is significantly different. 

-----------------------------

I tested a motion with the FJT script. The motion worked, but the final position was outside of tolerance on both axes that I moved by 0.012 rad. 

I'm wondering if this is related to the 'laggy' response I get when jogging. 

----------------------------

In the error message about the trajectory being outside tolerance, the joint names were non-ascii characters.

https://github.com/Yaskawa-Global/motoros2/blob/7f95156533a90f5efa0a46dd7f00c74a32c439bd/src/ActionServer_FJT.c#L571-L573

This has me questioning if something is overflowing an array. Or perhaps [this](https://github.com/Yaskawa-Global/motoros2/blob/7f95156533a90f5efa0a46dd7f00c74a32c439bd/src/ActionServer_FJT.c#L231-L232) isn't working properly.

gavanderhoorn:
Hm. Sounds like something isn't working.

re: velocities & torques: could this be rate-dependent?

gavanderhoorn:
@ted-miller: I've added 422e2b2b735ac4c11464e9bdb096b29c132295f6 which (im)ports the MotoROS1 version of `Ros_CtrlGroup_GetFBServoSpeed(..)` to use `mpSvsGetVelTrqFb(..)` instead of the M-registers.

gavanderhoorn:
@ted-miller: I just remembered I haven't actually built this branch after I added d4fb2fd997832330b6a6d7b6c3914109946ecb4e.

gavanderhoorn:
@ted-miller wrote:
> In the error message about the trajectory being outside tolerance, the joint names were non-ascii characters.
> 
> https://github.com/Yaskawa-Global/motoros2/blob/7f95156533a90f5efa0a46dd7f00c74a32c439bd/src/ActionServer_FJT.c#L571-L573

Should that be `feedback_FollowJointTrajectory.feedback.joint_names.data[axis].data`? Technically `feedback_FollowJointTrajectory.feedback.joint_names.data[axis]` is a `rosidl_runtime_c__String`, not a `const char*`.

> This has me questioning if something is overflowing an array. Or perhaps [this](https://github.com/Yaskawa-Global/motoros2/blob/7f95156533a90f5efa0a46dd7f00c74a32c439bd/src/ActionServer_FJT.c#L231-L232) isn't working properly.

Technically I believe we could use `rosidl_runtime_c__String__copy(..)` instead of 'abusing' `_assign(..)`, but it wouldn't really make much difference IIUC.

ted-miller:
> I just remembered I haven't actually built this branch after I added https://github.com/Yaskawa-Global/motoros2/commit/d4fb2fd997832330b6a6d7b6c3914109946ecb4e.

I was able to build this yesterday. Though, I did have to shrink `SIZEOF_BUFFER_FJT_GOAL` yet again to make it load without 1020 [6]. (Admittedly, I didn't strip the binary after building and it was 300-400k larger than `alpha50`. So perhaps that would have made a difference.)

Also, there were a bunch of annoying warnings about an unneeded `const` identifier. I'm thinking that we should remove the `-Wall` from the compiler arguments for FS100. Presumably, any "important" warnings would also be raised by the DX/YRC compilers. 

> Should that be feedback_FollowJointTrajectory.feedback.joint_names.data[axis].data

Technically, yes. But since the string pointer is the first entry in that structure, it should be ok.

gavanderhoorn:
> Though, I did have to shrink `SIZEOF_BUFFER_FJT_GOAL` yet again to make it load without 1020 [6]. (Admittedly, I didn't strip the binary after building and it was 300-400k larger than `alpha50`. So perhaps that would have made a difference.)

that should indeed not really be needed.

I may be doing something 'different' when building it here. I'll take a look at my settings.

> Also, there were a bunch of annoying warnings about an unneeded `const` identifier. I'm thinking that we should remove the `-Wall` from the compiler arguments for FS100. Presumably, any "important" warnings would also be raised by the DX/YRC compilers.

I don't remember anything about `const` modifier (I guess?), I do have warnings about return types, for which I added `-Wno-return-type`. But only for the FS100 builds.

> > Should that be feedback_FollowJointTrajectory.feedback.joint_names.data[axis].data
> 
> Technically, yes. But since the string pointer is the first entry in that structure, it should be ok.

technically, yes. After I changed that when working on #233, the issues with `double`s and `Ros_Debug_BroadcastMsg(..)` stopped.

gavanderhoorn:
@ted-miller: have you had any time to take a further look at this?

ted-miller:
No I haven't. 
I'm definitely not going to be able to get to it until at least the end of the month.

Sanchitkedia:
@gavanderhoorn I have a Yaskawa Motoman SIA20F with an FS100 controller and can try to run this on my end. Could you help me with the necessary files and instructions on how to get started?

gavanderhoorn:
@Sanchitkedia: could you provide some more information about your FS100? The current minimal controller software version we support is `4.00` for instance, which is relatively new (for FS100s, that is).

gavanderhoorn:
@ted-miller: I've drafted a *prerelease* version of M+ `libmicroros` for FS100 here: [Yaskawa-Global/micro_ros_motoplus/releases](https://github.com/Yaskawa-Global/micro_ros_motoplus/releases).

Sanchitkedia:
> @Sanchitkedia: could you provide some more information about your FS100? The current minimal controller software version we support is `4.00` for instance, which is relatively new (for FS100s, that is).

My System is currently running on FS 3.28.00

ted-miller:
> I've drafted a prerelease version of M+ libmicroros for FS100 here

Does this contain anything new/changed since the April tests?

> My System is currently running on FS 3.28.00

@Sanchitkedia I'll send a download link for new FS100 software to the email address in your github profile.

gavanderhoorn:
> > I've drafted a prerelease version of M+ libmicroros for FS100 here
> 
> Does this contain anything new/changed since the April tests?

not necessarily. It's `libmicroros` for the FS100, so not a MotoROS2 binary.

I did make sure to include all the changes made to `micro_ros_motoplus` since we migrated to the new distribution approach.

This is also the version that -- together with some additional changes on the MotoROS2 side -- results in significantly smaller binaries.

gavanderhoorn:
Rebased on top of `main` and added a temporary change to CI to get it to run with the draft M+ `libmicroros` release for FS100 mentioned in https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2187076126.

Sanchitkedia:
> @Sanchitkedia: have you been able to upgrade your FS100?

Not yet, but it should be done this week. I await the approval and go-ahead from others who use the robot.

Sanchitkedia:
@gavanderhoorn @ted-miller Upgrade to 4.00 was successful.

gavanderhoorn:
@ted-miller wrote:
> Other quick notes:
> 
> * `effort` data in joint_states is garbage data

and:

> In the error message about the trajectory being outside tolerance, the joint names were non-ascii characters.

just tested on FS100 again, and although I did not encounter the things you describe @ted-miller, I did notice some strings in the debug log contained non-ascii characters.

Specifically, in this line:

https://github.com/Yaskawa-Global/motoros2/blob/a1c9b5b7f9a9e0806dd67ba9f087701d4a821bd6/src/ServiceResetError.c#L126

`response->message` contained things like `ELP`, `E=@` and `` EA` ``. Or at least I think it did, as that's what showed up on the PC running the debug client.

gavanderhoorn:
@Sanchitkedia: have you had a chance to test the binary?

jfernandez37:
@gavanderhoorn Hello. I work with Sanchit and we are testing it today. We do not see the yaml file in the zip.

jfernandez37:
While trying to install, we could not find how to upgrade to the safety security level. Our only three options are editing, operation, and management.

jfernandez37:
Notes following the installation procedure for motoman:

1. We didn't find the safety security mode (Skipped this step)
2. When trying to load the user app file, got error 3640: limit of file number. We have three files in the file list: MotoFitFS.out, motoPlusTcpServer.out, and MotoRosFS100.out

ted-miller:
> We didn't find the safety security mode

I don't believe safety security exists on the FS100. You can just use MANAGEMENT.

> We have three files in the file list: MotoFitFS.out, motoPlusTcpServer.out, and MotoRosFS100.out

You will almost certainly need to remove MotoRosFS100.out. We have not done any compatiblity testing with MotoFit. I suspect that the motoPlusTcpServer.out is a custom application created by a colleague in your office.

I recommend testing MotoROS2 with all of the other .out file removed, to ensure there are no resource conflicts.

jfernandez37:
While installing, we found S2C1402 and we don't see S1D [https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2018801181]

ted-miller:
S2C1402 does exist in the controller, but I don't believe it is used by the FS100 software.

The S1D parameter is a protected parameter. It can only be accessed by a Yaskawa technician. (I should not have posted that on a public forum.)

For the YAML file, please use the standard one from https://github.com/Yaskawa-Global/motoros2/tree/main/config

You will need to load it using the "DX200 method" using the USB port on the main CPU board

gavanderhoorn:
@jfernandez37: could you please use a single comment to document your installation experience and tests in?

You can edit comments later to add new content.

Otherwise this PR thread will get *very long*.

jfernandez37:
Installation successful. Joint states look to be updating correctly.

gavanderhoorn:
> For it to work, we had to leave the USB drive in.

I'd be very interested in all the debug logs you've collected.

SRAM should work on FS100, but perhaps my test controller is special.

jfernandez37:
[20240711T170246_motoros2_debug_log.txt](https://github.com/user-attachments/files/16194759/20240711T170246_motoros2_debug_log.txt)
Here is the logs from one of the times we ran it. This is from the debug log client. We could not find PANELBOX.LOG file on the pendant.

gavanderhoorn:
Thanks, but this does not appear to be a log that contains the startup procedure.

Did you start the logger *before* powering on the FS100?

jfernandez37:
No, but I will start the logger before powering the FS100 when I am back in the lab next week.

gavanderhoorn:
thanks, but I was mostly interested in the log of the first run during/after installation. That's why I wrote (in https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2205991850):

> Before installation, please make sure to have the [debug log client](https://github.com/Yaskawa-Global/motoros2/blob/a1c9b5b7f9a9e0806dd67ba9f087701d4a821bd6/doc/troubleshooting.md#debug-log-client) running. This will let you (and us) capture all output MotoROS2 may produce while starting up and during execution and will help us diagnose any issues.

I'm still interested to see a log from a cold start though.

And one without the USB stick in the USB port.

jfernandez37:
I just powered on the FS100 with and without the drive in the USB port. It ran correctly both times. Attached is the cold start log.
[20240716T102554_motoros2_debug_log.txt](https://github.com/user-attachments/files/16250584/20240716T102554_motoros2_debug_log.txt)

gavanderhoorn:
Thanks. Looks like SRAM indeed works.

Please keep us updated on your experiences with `alpha60`.

jfernandez37:
Hello. I am trying to run the ros2 driver, but I am now getting an `eprosima::fastcdr::exception::NotEnoughMemoryException` error. Attached is the log.
[20240716T154504_motoros2_debug_log.txt](https://github.com/user-attachments/files/16255881/20240716T154504_motoros2_debug_log.txt)

gavanderhoorn:
Could you please confirm you are using a supported version of ROS 2? The `alpha60` binary I shared with you only works with ROS 2 Humble, and only the `amd64` version of ROS 2 Humble, either on Windows or on Linux, with the `microros/micro-ros-agent:humble` Docker image or the Agent built from source.

jfernandez37:
That was also my issue. I was accidentally on iron instead of humble. Thanks.

gavanderhoorn:
@jfernandez37 @Sanchitkedia: have you had any chance to test `alpha60` more?

jfernandez37:
Yes. We have been testing it and are able to move the robot using a follow joint trajectory action client.

jfernandez37:
I did not run into the same issues. The velocity seems to be correctly publishing correctly. I have attached a rosbag with /joint_states and /robot_status which was created while the robot was moving using a follow joint trajectory action client.
[motoros2_output.zip](https://github.com/user-attachments/files/16366108/motoros2_output.zip)

jfernandez37:
While trying to use moveit, I ran into an issue with getting the joint states. This is the error message I got:
```[motoman_robot_commander_node-1] [INFO] [1721939822.516113008] [moveit_ros.current_state_monitor]: Listening to joint states on topic 'joint_states'
[motoman_robot_commander_node-1] [INFO] [1721939833.546227356] [moveit_ros.current_state_monitor]: Didn't receive robot state (joint angles) with recent timestamp within 1.000000 seconds. Requested time 1721939832.546154, but latest received state has time 0.000000.
[motoman_robot_commander_node-1] Check clock synchronization if your are running ROS across multiple machines!
[motoman_robot_commander_node-1] [ERROR] [1721939833.546280750] [move_group_interface]: Failed to fetch current robot state
```
Any ideas on how to fix this?

ted-miller:
> Interesting you don't run into the issues with the efforts and velocities @ted-miller reported.
> @ted-miller: would you have any idea what could cause this difference?

No idea. I would need to reconnect my FS to test with.

gavanderhoorn:
@jfernandez37: from your bag file it would appear MotoROS2 is publishing on `joint_states`, so this could be a MoveIt configuration issue, or a QoS issue.

I'd make sure:

- [you've configured the QoS](https://github.com/Yaskawa-Global/motoros2/blob/f61a7238d78066aa565542b6fefd112bad42d0cb/config/motoros2_config.yaml#L225-L240) for `joint_states` to be compatible with MoveIt ([Readme: Default QoS settings](https://github.com/Yaskawa-Global/motoros2/blob/f61a7238d78066aa565542b6fefd112bad42d0cb/README.md#default-qos-settings))
- you haven't namespaced MoveIt (you'd probably need to remap the `joint_states` topic then)
- you haven't put MotoROS2 in a namespace (you'd probably need to remap the `joint_states` topic then)

gavanderhoorn:
@jfernandez37: have you been able to get things to work?

jfernandez37:
Yes. We got it to work. Our issue was with the moveit config. Thanks for your help.

gavanderhoorn:
Could you describe what you needed to change? This could be important information for (future) users.

jfernandez37:
Sure. The moveit config I was using was for iron, so the planning yaml file needed to be changed from this
```yaml
planning_plugin:
  ompl_interface/OMPLPlanner
response_adapters:
  - default_planning_response_adapters/AddTimeOptimalParameterization
```
To this:
```yaml
planning_plugin: ompl_interface/OMPLPlanner
request_adapters: >-
  default_planner_request_adapters/AddTimeOptimalParameterization
  default_planner_request_adapters/FixWorkspaceBounds
  default_planner_request_adapters/FixStartStateBounds
  default_planner_request_adapters/FixStartStateCollision
  default_planner_request_adapters/FixStartStatePathConstraints
```

gavanderhoorn:
Would you have a debug log of when this happens?

jfernandez37:
Thanks for your response. Attached is the log. 
[20240731T121704_motoros2_debug_log.txt](https://github.com/user-attachments/files/16444616/20240731T121704_motoros2_debug_log.txt)

gavanderhoorn:
Can you describe a bit more what you were attempting to do? Without more information this is going to be rather difficult to diagnose. All I can say at this point is MotoROS2 ran into an issue while executing your trajectory which caused it to crash. Hence the abrupt end of the debug log and the timeout on the Agent's side.

The trajectory doesn't appear to be overly long (6 points) so I doubt that could be the problem.

We've had a couple of cases / test reports where some memory corruption occurred on FS100, which could cause what see -- but without more debugging it's going to be hard to tell.

jfernandez37:
Sure. Right now, there is a moveit node which is supposed to move the end effector of the robot down by one centimeter and then in a separate trajectory, move back up to it's original position. The first movement is planned and executed correctly, but after it is executed, the agent times out. I have tried executing the trajectory directly through moveit and also through an action client directly to the `follow_joint_trajectory` action, but both had the same result. None of the trajectories have had many points, so I do not think that this would be the issue.

gavanderhoorn:
Could you post the trajectory / something representative here?

jfernandez37:
Here is the trajectory:

[trajectory_msgs.txt](https://github.com/user-attachments/files/16446313/trajectory_msgs.txt)

gavanderhoorn:
Somewhat off-topic, but: that trajectory is about 0.45 seconds long. Distance travelled is not that long, but I doubt an SIA20 would be able to actually execute this in that time.

gavanderhoorn:
@jfernandez37: this will take me some time to get to. I'm planning to spend some time on it beginning next week.

I'll probably have to try and replicate the symptom you're seeing.

Can you execute any trajectories generated by MoveIt, or does it never work? Have you been able to notice a pattern? Or does it just flat out refuse to execute everything?

jfernandez37:
I can execute one trajectory. After one trajectory is executed, the connection to motoros2 stops and nothing else can be run. Also, I was just trying to connect to the controller and I got this issue:
```
terminate called after throwing an instance of 'eprosima::fastcdr::exception::NotEnoughMemoryException'
  what():  Not enough memory in the buffer stream
[ros2run]: Aborted
```
I have reset the controller and my computer, but am still unable to connect

gavanderhoorn:
https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2233060567?

jfernandez37:
Humble was sourced before running. I was able to connect this morning, but not since and nothing has changed since then.

jfernandez37:
I've done some more testing with trajectory execution. It seems that moveit is not the problem, but any trajectory where a joint moves more than a certain amount, this causes the micro-ros agent to crash. For example, the micro-ros agent was fine when I moved the base by 0.2 radians, but when I move a joint by pi/2, the agent crashes.

ted-miller:
> Humble was sourced before running

Just to confirm, you're certain that you're using the humble version of the agent docker image? Is there any chance that more than one instance of an agent is running?

I just ran a few tests. My agent does not crash, regardless of the quantity or size of the motion steps.

One oddity that did occur, is that after opening a second terminal, I started getting this error after every ros2 command I issue from the CLI.
`2024-08-07 14:41:43.861 [RTPS_TRANSPORT_SHM Error] Failed init_port fastrtps_port7411: open_and_lock_file failed -> Function open_port_internal`
But it doesn't seem to affect the behavior.

----------------------------------

> For effort:
> I don't know what to think of the numbers being displayed

I still don't have a clue if the numbers are accurate. 

> The feedback position is 'laggy'.

> I tested a motion with the FJT script. The motion worked, but the final position was outside of tolerance on both axes that I moved by 0.012 rad.
> I'm wondering if this is related to the 'laggy' response I get when jogging.

Still true.  Every FJT action has ended with an error. 

> In the error message about the trajectory being outside tolerance, the joint names were non-ascii characters.

So now I'm getting a valid joint name. But the deviation amount is clearly incorrect. `control_msgs.action.FollowJointTrajectory_Result(error_code=-500301, error_string='Final position was outside tolerance. Check robot safety-limits that could be inhibiting motion. [group_1/joint_1: 1060489246 deviation]')`

-------------------------------------

I've tried running Jimmy's test app and I got some weird behavior regarding `stop_traj_mode` service. Sometimes it works fine, but I'm seeing frequent cases of an unknown error.

```
stop_traj_mode: attempting to stop trajectory mode
WARNING: Message processing not stopped before clearing queue
stop_traj_mode: StopMotion failed (unknown error)
```

But it seems to always work when manually invoked from the CLI. So I'm guessing there's just a timing situation.

jfernandez37:
We have done more testing and have isolated the issue. During execution of certain trajectories, we lose connection to the controller. We receive this error message from the micro-ros agent:
```
[1723150710.094950] info     | Root.cpp           | delete_client            | delete                 | client_key: 0xB52B2416
[1723150710.095008] info     | SessionManager.hpp | destroy_session          | session closed         | client_key: 0xB52B2416, address: 192.168.1.33:7387
[1723150710.095018] info     | Processor.cpp      | check_heartbeats         | Session destroyed due to liveliness timeout | client_key: 0xB52B2416, address: 192.168.1.33:7387
``` 
Currently, this error occurs on trajectories that modify the position of joint U as well as trajectories that modify joint S and T simultaneously. It's possible that other combinations cause this error, but these are the only combinations we've seen so far.  When the error occurs the motion is still executed, but we never receive a response from the action server and micro-ros can no longer connect to the robot without a restart of the controller.

We have captured the results of the debug listener for both properly executed trajectories and trajectories which fail. For both cases, we executed a single joint motion of 0.1 radians over 3 seconds. The debug files are attached.

[joint_0_success.txt](https://github.com/user-attachments/files/16553470/joint_0_success.txt)
[joint_3_fail.txt](https://github.com/user-attachments/files/16553471/joint_3_fail.txt)

gavanderhoorn:
@jfernandez37: I started work to try and reproduce what you've reported last week, but haven't been able so far. My FS100 isn't setup with physical hw, which may be why. I'll try some more ways to 'break' it.

gavanderhoorn:
@jfernandez37: I'm starting to suspect the sleep/timer implementation on FS100 platforms (not the base OS, but MotoROS2's wrapper/use of it), but haven't been able to pin-point anything yet.

Have you observed any more issues in the meantime?

And thank you for testing the alpha. This is valuable.

jfernandez37:
No, we haven't noticed any more issues with it. Thank you for your help.

gavanderhoorn:
#296 fixes at least one case of "strange chars" in the debug log of the FS100 build I'm testing.

gavanderhoorn:
Just to provide an update: @jfernandez37: due to the way my FS100 is configured I can't seem to reproduce the problem(s) you encountered.

I've tried to break things by executing many different trajectories all over the (combined) joint position ranges of all joints without running into any issues so far. I've identified a couple of other issues with the port which we'll be addressing later, but we're going to need a different setup to debug.

I believe @ted-miller may have an opportunity soon to assist.

gavanderhoorn:
@ted-miller wrote:
> I have not been able to reproduce this issue either. Using a small 6-axis arm, I'm able to execute any trajectories repeatedly.

did you encounter any other issues while executing those trajectories?

This was with actual HW right?

> I also tried reconfiguring as a 7-axis SIA (with a virtual arm). Still not able to cause this issue.
> 
> It seems the problem is with the micro-ros agent. Please ensure that the latest humble docker image is being used.

Just making sure (again? perhaps) @jfernandez37: you are running this on an `am64` desktop/laptop and are indeed using the `latest` Agent Docker image?

Could (probably is) something else, but just to rule things out.

One possible other cause of the error mentioned in https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2260735928 would be a buffer overflow (on the MR2 side) and/or memory corruption.

> Other oddities:
> 
> * I didn't initially follow the setup instructions properly and forgot to set some of the parameters. I got fatal alarm [10] (couldn't create publisher) until I set these and rebooted.
>   I don't understand why this alarm occurred.

Yes, I remember running into this myself. I'm also unsure what causes that. I've considered it a red herring for now.

> * When setup as a virtual arm, I got a whole lot of `Failed to get pulse position (command)`. But the next day, that went away. I don't _think_ I changed anything. But I can't explain this. (Probably a moot point, since I was configured for DUMMY mode.)

was this with `DUMMY_SERVO_MODE` on the MR2 side, or without?

> * `start_traj_mode` didn't work in DUMMY mode either. It wouldn't enable the servos automatically. I had to manually turn them on and then switch to REMOTE mode. (Again, moot since this is DUMMY)

Also ran into this. Had to use the same work-around.

thanks for testing.

---

At this point I'll see if can create a heavily instrumented version of MotoROS2 which logs just about everything around the trajectory processing, re-ordering and execution queues.

gavanderhoorn:
Apologies, that was supposed to read `amd64` (ie: `x86_64`). But thank you for confirming.

jfernandez37:
I just checked the laptop that we have been using, and it has an Intel cpu

gavanderhoorn:
> I just checked the laptop that we have been using, and it has an Intel cpu

that's ok. `amd64` is an architecture, not a brand. As long as it's not an ARM based processor, or something more exotic. The current version of MotoROS2 for FS100 is only compatible with `x86_64`.

ted-miller:
> did you encounter any other issues while executing those trajectories?

I got the "outside tolerance" failure that was mentioned previously. I think there's something "laggy" with the FS100. 

But the arm executed the motion just fine.


> This was with actual HW right?

The 6 axis tests were with a physical arm. The 7 axis tests were dummy-mode.


> > When setup as a virtual arm, I got a whole lot of Failed to get pulse position (command). But the next day, that went away. I don't think I changed anything. But I can't explain this. (Probably a moot point, since I was configured for DUMMY mode.)

> was this with DUMMY_SERVO_MODE on the MR2 side, or without?

Yes, DUMMY SERVO MODE was enabled when compiling MR2. `Ros_CtrlGroup_GetPulsePosCmd` doesn't look at the DUMMY flag because it's not trying to read feedback position.

gavanderhoorn:
I'm going to close this for now. As testing by @Sanchitkedia, @jfernandez37 and @ted-miller has shown, it needs some more work.

As I don't have sufficient time to work on it right now, I'll close it -- temporarily -- to take it off the review queue.

I'll re-open when I have something to update.

@Sanchitkedia @jfernandez37: I'll still try and get an instrumented build together for you, and if you'd still be interested in it I'd still appreciate your help testing it with your SIA20 setup. I would understand if you'd migrated to other solutions by that time though.

ted-miller:
Hi @ashwinsnambiar 
`/start_traj_mode` makes the robot accept a `FollowJointTrajectory` action which contains all points in the entire trajectory.

You want to use `start_point_queue_mode` to use the `queue_traj_point` service. 

Also, please use extra caution when moving the robot with the alpha binary. It has not undergone adequate testing and validation.

gavanderhoorn:
I might recommend actually to ask questions regarding functionality on the [Discussion forum](https://github.com/Yaskawa-Global/motoros2/discussions/categories/troubleshooting). As long as you clearly mention you're using the alpha for FS100, that should be ok.

re: "Waiting for an action server to become available...": the queue point service is a service, not an action server, so that message somewhat confuses me.

ashwinsnambiar:
> Hi @ashwinsnambiar `/start_traj_mode` makes the robot accept a `FollowJointTrajectory` action which contains all points in the entire trajectory.
> 
> You want to use `start_point_queue_mode` to use the `queue_traj_point` service.
> 
> Also, please use extra caution when moving the robot with the alpha binary. It has not undergone adequate testing and validation.

Sorry, that was my mistake. I copy pasted the wrong command here. I had indeed used the `FollowJointTrajectory` action. That resulted in the `waiting for an action server.... `.

When that didn't work out, I tried triggered the `start_point_queue_mode`and made the above `queue_traj_point` service call. That too did not pan out.

As @gavanderhoorn suggested, I'll move this to the discussions page.

ashwinsnambiar:
I just wanted to confirm that I am also facing the same issue as mentioned in https://github.com/Yaskawa-Global/motoros2/pull/227#issuecomment-2276664882. I have verified that I am running the latest 'humble' docker image of the micro-ros agent on a `amd64` platform. In addition to the test cases mentioned previously, I am facing this issue moving just the `s` joint, executing a trajectory of `1.65 -> 1.63 -> 1.64 -> 1.65` radians over a time interval of 5 secs each.

gavanderhoorn:
I can't guarantee I can find time to work on this, but if you can provide a reproducible test case, that might help.

ashwinsnambiar:
I have attached the [cpp code in txt format](https://github.com/user-attachments/files/19564043/client_traj_action.cpp.txt), which I used to send goal requests to the FollowJointTrajectory action server. It is based on the files from https://github.com/YaskawaEurope/ros2-starter-for-yaskawa-robots. At times, the micro-ros agent crashes immediately on running this action client node. Sometimes, the session closes only after running it 2-3 times. After that, the controller has to be restarted, to connect the micro-ros agent again. Also, I won't be able to test it on the robot until May again, so no worries about the time. If you could point me in the right direction, I could also take a look. Though I have to admit, I'm a novice at this.

gavanderhoorn:
Thanks for sharing your code.

You mentioned using MoveIt earlier.

Have you ever been able to trigger the disconnects / make things stop working when using MoveIt during your testing?

ashwinsnambiar:
Yes, I've been using Moveit too. It does almost always disconnect after every execution of small trajectories. I was using the motion planning with `kdl_kinematics_plugin` to plan the trajectories through the GUI.

lesurJ:
Commenting to follow this thread as i am also very much interested about a ROS2 port on the FS100 controller.

steelshot47:
We are getting this same issue here running on a YRC1000 where only on certain trajectories do we get a liveliness timeout. We have checked and confirmed that the agent and client ROS version is the same (Humble). Is there any update on whether this was recreated or has a solution?

Would updating to Jazzy help?

PR #229, opened 2024-03-27
Issue 203 auto switch cycle mode

yai-rosejo:
#203 
A continuation of the discussion and development of the cycle mode. These small changes allow Ros_MotionControl_StartMotionMode() to verify that the system is in a continuous cycle mode, and if it isn't it will try to set the cycle mode to continuous.


ted-miller:
Just realized that I missed this. I'll validate in the next day or so. But the code changes look OK with a quick glance.

gavanderhoorn:
I'm still wondering whether it would help raising an alarm if for some reason we can't change the cycle setting, in addition to the debug log statement.

(the service call will of course still fail with an appropriate message)

ted-miller:
I agree that an alarm is warranted if the set-operation fails _and_ the current cycle is `STEP`.

ted-miller (changes requested):
@yai-rosejo 
I just tested and it works great. 

Two things:
1) It would nice to implement that alarm.
2) I think it should do a final check to `Ros_Controller_IsContinuousCycle` to verify the setting was applied. This would require a call to `Ros_Controller_IoStatusUpdate` to fetch the value.

If you're tied up with Matt, let me know and I'll implement the changes.

ted-miller (approved):

src/MotionControl.c
```diff
@@ -1436,6 +1430,34 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
...
+        bzero(&sCycleData, sizeof(sCycleData));
+        bzero(&rData, sizeof(rData));
+        sCycleData.sCycle = 3;  // Auto
+        ret = mpSetCycle(&sCycleData, &rData);
+        if( (ret != 0) || (rData.err_no != 0) ) 
+        {
+            Ros_Debug_BroadcastMsg(
+                "Can't set cycle mode to continuous because: '%s' (0x%04X)",
+                Ros_ErrorHandling_ErrNo_ToString(rData.err_no), rData.err_no);
+            mpSetAlarm(ALARM_OPERATION_FAIL, "Set job-cycle to AUTO", SUBCODE_OPERATION_SET_CYCLE);
```
gavanderhoorn, line 1447: This message doesn't appear to correspond to what the documentation shows?

(same below)

ted-miller: Oops. I changed my mind on the alarm msg after writing the doc. It is now updated.

doc/ros_api.md
```diff
@@ -111,7 +111,7 @@ Note: errors and alarms which require physical operator intervention (e-stops, e
 
 Type: [motoros2_interfaces/srv/StartTrajMode](https://github.com/yaskawa-global/motoros2_interfaces/blob/d6805d32714df4430f7db3d8ddc736c340ddeba8/srv/StartTrajMode.srv)
 
-Attempt to enable servo drives and activate trajectory mode, allowing the action server (`follow_joint_trajectory`, see below) to execute incoming `FollowJointTrajectory` action goals.
+Attempts to set cycle mode, enable servo drives, and activate trajectory mode, allowing the action server (`follow_joint_trajectory`, see below) to execute incoming `FollowJointTrajectory` action goals.
```
gavanderhoorn, line 114: we're probably not going to do anything about it, but it's a bit 'mysterious' to have this state it's going to *set cycle mode* without any motivation. Enabling the servos and activating trajectory mode all makes sense, but changing cycle mode seems odd.

doc/troubleshooting.md
```diff
@@ -1117,3 +1117,24 @@ ALARM 8015
...
+*Example:*
+
+```text
+ALARM 8016
+ Set job-cycle to AUTO
+[0]
+```
+
+*Solution:*
+The job cycle is currently set to `STEP` and MotoROS2 was unable to automatically change it to `AUTO`.
```
gavanderhoorn, line 1132: should we perhaps add some clarification as to *why* this is needed?

doc/troubleshooting.md
```diff
@@ -1117,3 +1117,24 @@ ALARM 8015
...
+
+First, manually set the controller to use `AUTO`.
```
gavanderhoorn, line 1134: `AUTO` what?

doc/troubleshooting.md
```diff
@@ -1117,3 +1117,24 @@ ALARM 8015
...
+
+ 1. upgrade to *MANAGEMENT* security level by touching `[System Info]`→`[Security]` (default password is all `9`'s)
+ 1. touch `[Setup]`→`[Operate Cond.]`
+ 1. change `CYCLE SWITCH IN REMOTE MODE` to `AUTO`
+
+If the problem persists, verify that the `CIOPRG.LST` ladder program is not writing to `#40050 - #40052`.
```
gavanderhoorn, line 1140: I expect we'll receive support requests here, because this is rather cryptic.

src/MotionControl.c
```diff
@@ -1436,6 +1430,34 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
...
     }
 
+    // Check if in continous cycle mode
+    if (!Ros_Controller_IsContinuousCycle())
+    {
+        // set the cycle mode to auto if not currently
+        MP_CYCLE_SEND_DATA sCycleData;
+        bzero(&sCycleData, sizeof(sCycleData));
+        bzero(&rData, sizeof(rData));
+        sCycleData.sCycle = 3;  // Auto
```
gavanderhoorn, line 1440: M+ does not define any named constant(s) for this?

src/MotionControl.c
```diff
@@ -1436,6 +1430,34 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
...
+        ret = mpSetCycle(&sCycleData, &rData);
+        if( (ret != 0) || (rData.err_no != 0) ) 
+        {
+            Ros_Debug_BroadcastMsg(
+                "Can't set cycle mode to continuous because: '%s' (0x%04X)",
+                Ros_ErrorHandling_ErrNo_ToString(rData.err_no), rData.err_no);
```
gavanderhoorn, line 1446: is `rData.err_no` guaranteed to be non-zero in case `ret != 0`?

src/MotionControl.c
```diff
@@ -1436,6 +1430,34 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
...
+            mpSetAlarm(ALARM_OPERATION_FAIL, "Set job-cycle to AUTO", SUBCODE_OPERATION_SET_CYCLE);
+            return FALSE;
+        }
+
+        Ros_Controller_IoStatusUpdate(); //verify the cycle got set and wasn't forced back due to CIO logic
```
gavanderhoorn, line 1451: would this require a delay first?

gavanderhoorn:
Friendly ping?

ted-miller:
Apologies for the delay. I will look at this again later in the week.

doc/ros_api.md
ted-miller: Are you saying that you'd prefer we remove this text? 

I think it's implied that if this service is performing a particular action, then there's a good reason for it.

src/MotionControl.c
ted-miller: It does not

src/MotionControl.c
ted-miller: Yes

src/MotionControl.c
ted-miller: Yeah. good catch

doc/troubleshooting.md
ted-miller: I added text to contact yaskawa directly

doc/ros_api.md
gavanderhoorn: > Are you saying that you'd prefer we remove this text?

No, just that the part about this 'attempting to set cycle mode' is a bit cryptic. The other actions it performs 'make sense': can't move the robot without the servos being enabled or trajectory mode being activated.

But then it also *sets cycle mode*, which to me at least, is far from clear why.

But perhaps this is too much of a nit.

> I think it's implied that if this service is performing a particular action, then there's a good reason for it.

Of course. I just always like to understand *why* something happens.

gavanderhoorn:
I've added (a) named constant(s) in a461aeed7ecf9f5daf9e1806c05241ac5c4e4e0f and rebased on top of `main` as this was otherwise in conflict (due to move of header to the new `lib` sub dir).

ted-miller:
build-check was verified locally.

PR #231, opened 2024-04-02
Use separate directory for external dependencies (platform lib, CMOS parameter lib, etc)

gavanderhoorn:
As per title.

This clearly separates those dependencies from MotoROS2 proper and additionally, prevents GCC from treating them as part of MotoROS2 itself.

Conversely, it prevents GCC from treating MotoROS2 sources as 'external', and thus disabling many of its checks and warnings.


gavanderhoorn:
I was unsure as to whether this would work, but a test build (I've only tested the YRC1 configuration) seems to suggest it does.

@ted-miller: as you have more experience with this, could you please take a look?

I'm mostly trying to avoid the compiler marking the main `src` (project) directory as *system*, which then makes it ignore issues in that directory (which we don't want, obviously).

src/MotoROS2_AllControllers.vcxproj
```diff
@@ -126,109 +126,109 @@
     <NMakeBuildCommandLine>"$(MP_VS_Install)mpBuilder.exe" -c $(Configuration) -p "$(ProjectDir)\" -n "mr2_dx2_h" -b "$(SolutionDir)bin\$(Configuration)" -o build -i "$(MP_VS_Install)DX200\inc"</NMakeBuildCommandLine>
```
gavanderhoorn, line 126: I thought I'd had to update this as well to include the `$(SolutionDir)lib` directory as an include directory, but for some reason that's not needed.

ted-miller:
> I'm mostly trying to avoid the compiler marking the main src (project) directory as system, which then makes it ignore issues in that directory (which we don't want, obviously)

What type of issues have you seen the compiler ignore?

src/MotoROS2_AllControllers.vcxproj
ted-miller: The `lib` subfolder is being automatically passed to the compiler executable with a `isystem` tag. 

I'm sure it's due to the shenanigans that mpBuilder is doing to parse all the subfolders. But I don't fully understand it.

ted-miller:
> I'm mostly trying to avoid the compiler marking the main src (project) directory as system

I have [verified](https://learn.microsoft.com/en-us/sysinternals/downloads/procmon) that the main project folder is passed with `-I` and not `isystem`.

src/MotoROS2_AllControllers.vcxproj
ted-miller: Oh, I think it's pulling the path out of the vcxproj file

gavanderhoorn:
I'm doing some strange / unorthodox things locally which caused the `src` dir to be passed with `-isystem` instead of regular `-I`.

For normal use it doesn't really affect anything, but for the case(s) I describe having the external dependencies in a separate directory helps to avoid the `-isystem`.

gavanderhoorn:
@ted-miller: just making sure: ok with me merging this?

Issue #233, opened 2024-04-08
goal_tolerance parameter in follow_joint_trajectory action goal not properly supported


rsaintobert:
Anytime a trajectory is submitted I get the following error (with more or less joints incriminated), and the goal is aborted.

`Final position was outside tolerance. Check robot safety-limits that could be inhibiting motion. [gp50_joint_1_s: 0.0000 deviation] [gp50_joint_6_t: 0.0000 deviation]`

The log is done here but the 4-digits precision does not give enough info. Anyway the deviation is quite low here.
https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/ActionServer_FJT.c#L563-L580

The problem is more in the check of the deviation:
https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/ActionServer_FJT.c#L500-L520

Leaving the goal_tolerance parameter unset in the goal request should default to a tolerance of 0.01 which is not the case. Maybe the message buffer is badly initialized ?
https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/ActionServer_FJT.c#L511-L512

Anyway, any value set in the goal request leads to the same result, so the user input is not taken properly into account.

gavanderhoorn:
Could you please provide:

- the `motoros2_config.yaml` as loaded on the controller
- an example of a `FollowJointTrajectoryAction` goal which results in the problem you report (dumping the complete goal to `stdout` from your client program would be one way to do this)
- a [full debug log](https://github.com/Yaskawa-Global/motoros2/blob/b78ea310d6a5bb2b1c8f80ee0bf104c1986ca119/doc/troubleshooting.md#debug-log-client), from controller start up to and including a failed motion execution
- information on which version of MotoROS2 this is, version of ROS 2, how you installed it, etc
- information on your robot controller: type and robot configuration

(if easier: a minimal action client which submits a goal which results in the problem you report would also be ok -- and would make it easier to reproduce your problem)

gavanderhoorn:
I've edited the title of the issue to better reflect the symptoms observed.

We can't conclude yet what is the cause.

ted-miller:
> Leaving the goal_tolerance parameter unset in the goal request should default to a tolerance of 0.01 which is not the case

[...]

PR #236, opened 2024-04-11
Change `double` comparison to use epsilon instead of 0.0

ted-miller:
Regarding https://github.com/Yaskawa-Global/motoros2/issues/233#issuecomment-2045896777

I don't know if this will have an impact on #233. But it does seem like a potential problem.

### DIFF
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -503,7 +503,7 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
             diff = fabs(feedback_FollowJointTrajectory.feedback.desired.positions.data[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
 
             double posTolerance = g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[axis].position; //user-provided a goal_tolerance
-            if (posTolerance == 0.0) //user did NOT provide a tolerance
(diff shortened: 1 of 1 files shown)


src/ActionServer_FJT.c
```diff
@@ -503,7 +503,7 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
             diff = fabs(feedback_FollowJointTrajectory.feedback.desired.positions.data[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
 
             double posTolerance = g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[axis].position; //user-provided a goal_tolerance
```
gavanderhoorn, line 505: should we not check whether `g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data` is of the expected/required size? Otherwise we could be trying to read outside the array.

ted-miller: > We might already be doing that before we get here, I haven't checked

We're not.

You're correct in that this is another potential issue.

@yai-rosejo mentioned in passing that he's got a branch doing some more thorough checking. I'll let him comment on what he's done.

gavanderhoorn:
@ted-miller: I suggest we close this in favour of the more correct fix @yai-rosejo is preparing (https://github.com/Yaskawa-Global/motoros2/issues/233#issuecomment-2052605104)

ted-miller:
agreed

PR #241, opened 2024-04-16
Issue 233-outside goal tolerance

yai-rosejo:
This PR is to solve multiple issues dealing with goal tolerances referenced in #233 

1. Received Goal tolerances can now be received out of order, the name of the tolerance is used to match it with the expected joint.
2. The final goal position is used instead of the desired position to deal with a small out of sync issue.
3. The special notations of -1.0, to ignore tolerance, and 0.0, to use default values, is now supported.

This PR does not fix the issues with the Ros_Debug_BroadcastMsg function and doubles.

Testing:
[tolerance_issue_04162024.zip](https://github.com/Yaskawa-Global/motoros2/files/15001895/tolerance_issue_04162024.zip)

1. Upload the mr2_yrc1_h.out to the controller and then restart the controller
2. Run the microros client in a terminal : 
```sudo docker run -it --net=host -v /dev:/dev --privileged microros/micro-ros-agent:humble udp4 -p 8888```
3. Start the TrajMode:
```ros2 service call /start_traj_mode motoros2_interfaces/srv/StartTrajMode```
4. Run the fjt client:
```python3 fjt_client.py```
5. Optionally the debug_listener can also be ran to capture messages.

### DIFF
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -495,18 +495,59 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
         //check to see if each axis is in the desired location
         BOOL positionOk = TRUE;
         int maxAxes = MAX_CONTROLLABLE_GROUPS * MP_GRP_AXES_NUM;
-        double violators[MAX_CONTROLLABLE_GROUPS * MP_GRP_AXES_NUM];
 
+        double violators[MAX_CONTROLLABLE_GROUPS * MP_GRP_AXES_NUM];
         bzero(violators, sizeof(violators));
+
+        double posTolerance[maxAxes];
+        for (int i = 0; i < maxAxes; i++) {
+            posTolerance[i] = DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
+        }
+        for (size_t selected_tolerance_idx = 0; selected_tolerance_idx < g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.size; selected_tolerance_idx +=1)
+        {
+            rosidl_runtime_c__String selected_tolerance_name = g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[selected_tolerance_idx].name;
+            Ros_Debug_BroadcastMsg("FJT found position tolerance: %s, %f", selected_tolerance_name, g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[selected_tolerance_idx].position);
+
+            for (size_t selected_position_idx = 0; selected_position_idx < feedback_FollowJointTrajectory.feedback.joint_names.size; selected_position_idx += 1)
+            {
+                rosidl_runtime_c__String selected_position_name = feedback_FollowJointTrajectory.feedback.joint_names.data[selected_position_idx];
+                if (rosidl_runtime_c__String__are_equal(&selected_position_name,&selected_tolerance_name))
+                {
+                    Ros_Debug_BroadcastMsg("FJT using position tolerance: %s", selected_position_name);
+                    posTolerance[selected_position_idx] = g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[selected_tolerance_idx].position;
+                    break;
+                }
+            }
+        }
+
+        double finalPositions[maxAxes];
+        bzero(finalPositions, sizeof(finalPositions));
+        size_t trajectorySizeOffset = g_actionServer_FJT_SendGoal_Request.goal.trajectory.points.size - 1;
+        for (size_t selected_final_position_idx = 0; selected_final_position_idx < g_actionServer_FJT_SendGoal_Request.goal.trajectory.joint_names.size; selected_final_position_idx += 1)
+        {
+            rosidl_runtime_c__String selected_final_position_name = g_actionServer_FJT_SendGoal_Request.goal.trajectory.joint_names.data[selected_final_position_idx];
+            for (size_t selected_position_idx = 0; selected_position_idx < feedback_FollowJointTrajectory.feedback.joint_names.size; selected_position_idx += 1)
+            {
+                rosidl_runtime_c__String selected_position_name = feedback_FollowJointTrajectory.feedback.joint_names.data[selected_position_idx];
+                if (rosidl_runtime_c__String__are_equal(&selected_position_name, &selected_final_position_name))
+                {
+                    finalPositions[selected_position_idx] = g_actionServer_FJT_SendGoal_Request.goal.trajectory.points.data[trajectorySizeOffset].positions.data[selected_final_position_idx];
+                    Ros_Debug_BroadcastMsg("Final Position Goal: %s, %f", selected_position_name, finalPositions[selected_position_idx]);
+                    break;
+                }
+            }
+        }
+
         for (int axis = 0; axis < maxAxes; axis += 1)
         {
-            diff = fabs(feedback_FollowJointTrajectory.feedback.desired.positions.data[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
+            diff = fabs(finalPositions[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
+            // Ros_Debug_BroadcastMsg("desired - actual: %d, %d", (int)(1000 * finalPositions[axis]), (int)(1000 * feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]));
+            double chosen_posTolerance = posTolerance[axis];
+            if (chosen_posTolerance == 0.0) //user did NOT provide a tolerance
+                chosen_posTolerance = DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
 
-            double posTolerance = g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance.data[axis].position; //user-provided a goal_tolerance
-            if (posTolerance == 0.0) //user did NOT provide a tolerance
-                posTolerance = DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
-
-            if (diff > posTolerance)
+            // A negative tolerance means that tolerance doesn't matter for this axis
+            if ((chosen_posTolerance > 0.0) && (diff > chosen_posTolerance))
             {
                 positionOk = FALSE;
                 //record the deviation for use in the feedback message below
@@ -564,7 +605,7 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
                     if (violators[axis])
                     {
                         char formatBuffer[64] = { 0 };
-                        snprintf(formatBuffer, 64, " [%s: %.4f deviation]",
+                        snprintf(formatBuffer, 64, " [%s: %d deviation]",
                             feedback_FollowJointTrajectory.feedback.joint_names.data[axis],
                             violators[axis]);
                         strcat(msgBuffer, formatBuffer);
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -822,3 +822,7 @@ void Ros_ActionServer_FJT_ProcessResult()
         }
     }
 }
+
+
+//included here as this tests 'static' functions
+#include "Tests_ActionServer_FJT.c"
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -670,7 +670,8 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
         {
             double current_jstate = feedback_FollowJointTrajectory.feedback.actual.positions.data[axis];
             diff = fabs(lastTrajPtPositions[axis] - current_jstate);
-            Ros_Debug_BroadcastMsg("desired - actual: %d, %d", (int)(1000 * lastTrajPtPositions[axis]), (int)(1000 * current_jstate));
+            Ros_Debug_BroadcastMsg("desired - actual: %12.8f - %12.8f: %12.8f",
+                lastTrajPtPositions[axis], current_jstate, diff);
 
             double chosen_posTolerance = posTolerance[axis];
             if (chosen_posTolerance == 0.0) //user did NOT provide a tolerance
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -532,7 +532,8 @@ static STATUS Ros_ActionServer_FJT_Parse_GoalPosTolerances(
         Ros_Debug_BroadcastMsg("%s: parsing JointTolerance for '%s': pos: %f", __func__,
             selected_tolerance_name.data, goal_joint_tolerances->data[jtol_idx].position);
 
-        for (size_t ptol_idx = 0; ptol_idx < joint_names->size; ptol_idx += 1)
+        size_t ptol_idx = 0;
+        for (; ptol_idx < joint_names->size; ptol_idx += 1)
         {
             rosidl_runtime_c__String selected_position_name = joint_names->data[ptol_idx];
             if (rosidl_runtime_c__String__are_equal(&selected_position_name, &selected_tolerance_name))
@@ -542,6 +543,14 @@ static STATUS Ros_ActionServer_FJT_Parse_GoalPosTolerances(
                 posTolerances[ptol_idx] = goal_joint_tolerances->data[jtol_idx].position;
                 break;
             }
+        }
+
+        //couldn't find joint
+        if (ptol_idx == joint_names->size)
+        {
+            //TODO(gavanderhoorn): make this a fatal error?
+            Ros_Debug_BroadcastMsg("%s: WARNING: couldn't find '%s' in internal joint names, ignoring",
+                __func__, selected_tolerance_name.data);
         }
     }
 
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -642,15 +642,19 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
 
         //'parse' the JointTolerance elements from the goal. Map their 'name:tolerance'
         //to the 'joint_index:tolerance' we need
-        STATUS status = Ros_ActionServer_FJT_Parse_GoalPosTolerances(
+        STATUS statusParseGoalTolerance = Ros_ActionServer_FJT_Parse_GoalPosTolerances(
             &g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance,
             &feedback_FollowJointTrajectory.feedback.joint_names,
             posTolerance, numAxesToCheck);
 
-        if (status != OK)
+        BOOL bToleranceParseOk = TRUE;
+        if (statusParseGoalTolerance != OK)
         {
-            Ros_Debug_BroadcastMsg("%s: parsing 'goal_tolerance' field failed: %d", __func__, status);
-            //TODO(gavanderhoorn): what do we do here?
+            Ros_Debug_BroadcastMsg("%s: parsing 'goal_tolerance' field failed: %d", __func__, statusParseGoalTolerance);
+            //TODO(gavanderhoorn): implement tolerance parsing in Ros_ActionServer_FJT_Goal_Received(..) instead
+            bToleranceParseOk = FALSE;
+            Ros_Debug_BroadcastMsg("%s: skipping goal tolerance check", __func__);
+            goto goal_complete_skip_tolerance_comparison;
         }
 
         //retrieve the last traj pt from the goal traj and re-order position values such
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/ActionServer_FJT.c [hunk 1]
```diff
@@ -495,18 +495,59 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
...
+        size_t trajectorySizeOffset = g_actionServer_FJT_SendGoal_Request.goal.trajectory.points.size - 1;
+        for (size_t selected_final_position_idx = 0; selected_final_position_idx < g_actionServer_FJT_SendGoal_Request.goal.trajectory.joint_names.size; selected_final_position_idx += 1)
+        {
+            rosidl_runtime_c__String selected_final_position_name = g_actionServer_FJT_SendGoal_Request.goal.trajectory.joint_names.data[selected_final_position_idx];
+            for (size_t selected_position_idx = 0; selected_position_idx < feedback_FollowJointTrajectory.feedback.joint_names.size; selected_position_idx += 1)
+            {
+                rosidl_runtime_c__String selected_position_name = feedback_FollowJointTrajectory.feedback.joint_names.data[selected_position_idx];
+                if (rosidl_runtime_c__String__are_equal(&selected_position_name, &selected_final_position_name))
+                {
+                    finalPositions[selected_position_idx] = g_actionServer_FJT_SendGoal_Request.goal.trajectory.points.data[trajectorySizeOffset].positions.data[selected_position_idx];
```
gavanderhoorn, line 534: I haven't worked it out in detail, but should this be:

```suggestion
                    finalPositions[selected_position_idx] = g_actionServer_FJT_SendGoal_Request.goal.trajectory.points.data[trajectorySizeOffset].positions.data[selected_final_position_index];
```

?

yai-rosejo: You are correct, during my testing my final positions were all the same, so I didn't run into an issue with this line.

src/ActionServer_FJT.c
```diff
@@ -495,18 +495,59 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
...
+            }
+        }
+
+        for (int axis = 0; axis < maxAxes; axis += 1)
+        {
+            diff = fabs(finalPositions[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
+            // Ros_Debug_BroadcastMsg("desired - actual: %d, %d", (int)(1000 * finalPositions[axis]), (int)(1000 * feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]));
+            double chosen_posTolerance = posTolerance[axis];
+            if (chosen_posTolerance == 0.0) //user did NOT provide a tolerance
+                chosen_posTolerance = DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
```
gavanderhoorn, line 547: do we still need this is line 504 already sets a default for all entries in `posTolerance`?

src/MotionControl.h
```diff
@@ -8,7 +8,7 @@
 #ifndef MOTOROS2_MOTION_CONTROL_H
 #define MOTOROS2_MOTION_CONTROL_H
 
-#define START_MAX_PULSE_DEVIATION           30
+#define START_MAX_PULSE_DEVIATION           10
```
gavanderhoorn, line 11: I'm not sure this would work -- as @ted-miller wrote in #233, we might need a separate value for detecting the end of motion. There have been users who've had to increase this value (for larger / heavier robots fi) which would then break the end-of-motion detection.

src/ActionServer_FJT.c
yai-rosejo: This allows a client to reset back to the default position joint tolerance

src/MotionControl.h
yai-rosejo: Would we want to remove this from the PR and make a separate issue for dealing with detecting the end of motion?

gavanderhoorn: I'd suggest so.

This value actually already makes it impossible for my robot here (GP25) to start any motion.

yai-rosejo:
> @yai-rosejo: I've used your PR as a basis and restructured it a little. See [this](https://github.com/yai-rosejo/motoros2/compare/issue_233_outside_goal_tolerance...gavanderhoorn:motoros2:issue_233_outside_goal_tolerance_wip) for the diff. All WIP right now.
> 
> It's basically the same as your changes, but split into a couple of functions. My idea is we'd be able to write some tests for the new functionality this way.
> 
> I've tested it on my YRC with GP25 and it (still) works.
> 
> High-level question: did you intentionally iterate over the `maxAxes`? Would limiting it to just `g_Ros_Controller.totalAxesCount` (or `feedback_FollowJointTrajectory.feedback.joint_names.size`) not work as well?

I just did a cursory look through your solution; I think it looks much better than my current one.
I iterated over all of the maxAxes due to violators being the size of maxAxes and the 'for' loop checking the differences already using maxAxes. https://github.com/Yaskawa-Global/motoros2/blob/daa47bd7275dc7907f03ecae39ef159cbf87e032/src/ActionServer_FJT.c#L498-L504

yai-rosejo:
> Ok. Thanks.
> 
> I'll add some tests, and then push some commits to this PR if that would be OK with you.

That works for me!

gavanderhoorn:
Added some initial tests for `Ros_ActionServer_FJT_Parse_GoalPosTolerances(..)`:

<details><summary>Click to expand</summary>

```text
===
Performing unit tests
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: one or more inputs NULL, abort
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_null_args: goal_joint_tolerances NULL: PASS
Ros_ActionServer_FJT_Parse_GoalPosTolerances: one or more inputs NULL, abort
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_null_args: joint_names NULL: PASS
Ros_ActionServer_FJT_Parse_GoalPosTolerances: one or more inputs NULL, abort
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_null_args: posTolerances NULL: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_null_args: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: no joint tolerances specified, returning 6 defaults
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_empty_jtolerances: call: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_empty_jtolerances: all defaults: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_empty_jtolerances: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: pos tolerance output array too small: 1 < 2 (nr of joints)
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_out_array_too_small: out array too small: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_out_array_too_small: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint5, 1.000000
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_more_jtol_than_jnames: call: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_more_jtol_than_jnames: all defaults: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_more_jtol_than_jnames: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint0, 1.000000
Ros_ActionServer_FJT_Parse_GoalPosTolerances: using position tolerance: joint0
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot0: call: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot0: expected j0 pos tol: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot0: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint0, 1.000000
Ros_ActionServer_FJT_Parse_GoalPosTolerances: using position tolerance: joint0
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot2: call: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot2: j0 pos tol at [2]: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_jtol_in_slot2: PASS
~~~
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint1, 0.123000
Ros_ActionServer_FJT_Parse_GoalPosTolerances: using position tolerance: joint1
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint2, 0.234000
Ros_ActionServer_FJT_Parse_GoalPosTolerances: using position tolerance: joint2
Ros_ActionServer_FJT_Parse_GoalPosTolerances: found position tolerance: joint1, 0.345000
Ros_ActionServer_FJT_Parse_GoalPosTolerances: using position tolerance: joint1
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_setting_wins: call: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_setting_wins: j0 pos tol at [0]: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_setting_wins: j2 pos tol at [1]: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_setting_wins: j1 pos tol at [2]: PASS
Testing Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_setting_wins: PASS
~~~
Testing SUCCESSFUL
===
```

</details>

gavanderhoorn:
@yai-rosejo: I've added tests for the two new functions I introduced, refactored things a bit and cleaned up the Git history. See the [diff](https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:issue_233_outside_goal_tolerance_wip).

Your changes are now all squashed into the initial commit.

Could you take a look whether this would work for you? If you agree, I'll force-push these changes to your branch to update this PR.

gavanderhoorn:
@yai-rosejo wrote:
> @gavanderhoorn wrote:
> > High-level question: did you intentionally iterate over the `maxAxes`? Would limiting it to just `g_Ros_Controller.totalAxesCount` (or `feedback_FollowJointTrajectory.feedback.joint_names.size`) not work as well?
> 
> [..] I iterated over all of the maxAxes due to violators being the size of maxAxes and the 'for' loop checking the differences already using maxAxes.
> 
> https://github.com/Yaskawa-Global/motoros2/blob/daa47bd7275dc7907f03ecae39ef159cbf87e032/src/ActionServer_FJT.c#L498-L504

@ted-miller: would it be absolutely necessary to iterate over `maxAxes`, or would limiting those arrays and loops to just the configured nr of axes also work? I assume the latter, but wanted to make sure.

ted-miller:
> would it be absolutely necessary to iterate over maxAxes, or would limiting those arrays and loops to just the configured nr of axes also work? I assume the latter, but wanted to make sure.

Limiting the loops would be OK. I can't think of any justification for keeps maxaxes.

yai-rosejo:
> @yai-rosejo: I've added tests for the two new functions I introduced, refactored things a bit and cleaned up the Git history. See the [diff](https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:issue_233_outside_goal_tolerance_wip).
> 
> Your changes are now all squashed into the initial commit.
> 
> Could you take a look whether this would work for you? If you agree, I'll force-push these changes to your branch to update this PR.

The changes look good to me. You can go ahead with the force push.

gavanderhoorn:
Working on this a bit more and I'm beginning to wonder whether we shouldn't already parse the `JointTolerance` instances when first accepting the goal. If parsing fails -- due to unknown joints being referenced for instance -- we could refuse to execute the goal and provide a very clear error code and message to clients.

If we fail to parse at the end of an executed trajectory, we can't really do much anymore: failing the goal completely seems like a bit of a 'waste' of a good trajectory execution and failing so late for something almost 'unrelated' doesn't apply fail-early and least-surprise principles.

ted-miller:
So you're suggesting separating the logic of validating the tolerance input from validating the position is within that tolerance? That makes sense to me. 

We've got the `Ros_MotionControl_InitTrajectory` function which returns a custom `Init_Trajectory_Status` that could clearly identify an issue with the specified input.

My next question is how much of the goal message do we validate? So far, we've been ignoring much of the goal message, such as `path_tolerance` and the `component_x_tolerance`. Do those remain unused?

gavanderhoorn:
> So you're suggesting separating the logic of validating the tolerance input from validating the position is within that tolerance? That makes sense to me.

Indeed.

> We've got the `Ros_MotionControl_InitTrajectory` function which returns a custom `Init_Trajectory_Status` that could clearly identify an issue with the specified input.

True, although the `goal_tolerance` field is not part of the trajectory.

> My next question is how much of the goal message do we validate? So far, we've been ignoring much of the goal message, such as `path_tolerance` and the `component_x_tolerance`. Do those remain unused?

I would for now indeed suggest to fix & refactor the current implementation which only takes the `goal_tolerance` field into account, not extend it.

gavanderhoorn:
Coming back to this: unknown joints in `goal_tolerance` will not immediately cause problems, they will simply be ignored by the current implementation.

Adding goal validation -- which would include `goal_tolerance` validation -- would be nice, but wouldn't be needed to resolve the current issue (ie: #233).

I'll create an issue to log the enhancement request.

@yai-rosejo: I've pushed some more commits to [my version of your branch](https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:issue_233_outside_goal_tolerance_wip). If you could PTAL and let me know whether you'd still be ok with me force-pushing here? Thanks.

yai-rosejo:
> Coming back to this: unknown joints in `goal_tolerance` will not immediately cause problems, they will simply be ignored by the current implementation.
> 
> Adding goal validation -- which would include `goal_tolerance` validation -- would be nice, but wouldn't be needed to resolve the current issue (ie: #233).
> 
> I'll create an issue to log the enhancement request.
> 
> @yai-rosejo: I've pushed some more commits to [my version of your branch](https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:motoros2:issue_233_outside_goal_tolerance_wip). If you could PTAL and let me know whether you'd still be ok with me force-pushing here? Thanks.

I have taken a look and agree with pushing those onto this branch. This PR is focused on supporting goal functionalities, so I also agree that the goal validation should be a separate task.

src/ActionServer_FJT.c
```diff
@@ -494,19 +630,63 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
...
+        STATUS status = Ros_ActionServer_FJT_Parse_GoalPosTolerances(
+            &g_actionServer_FJT_SendGoal_Request.goal.goal_tolerance,
+            &feedback_FollowJointTrajectory.feedback.joint_names,
+            posTolerance, numAxesToCheck);
+
+        if (status != OK)
         {
-            diff = fabs(feedback_FollowJointTrajectory.feedback.desired.positions.data[axis] - feedback_FollowJointTrajectory.feedback.actual.positions.data[axis]);
+            Ros_Debug_BroadcastMsg("%s: parsing 'goal_tolerance' field failed: %d", __func__, status);
+            //TODO(gavanderhoorn): what do we do here?
```
gavanderhoorn, line 652: what should we do here? @yai-rosejo and @ted-miller any ideas?

ted-miller: IMO, `positionOk = FALSE;`
A trajectory should not 'pass' a tolerance-check if the tolerance cannot be properly evaluated.

yai-rosejo: For this line and also this one:
https://github.com/Yaskawa-Global/motoros2/blob/64c652409b88eab479940a9d61019ef99dfa6458/src/ActionServer_FJT.c#L662-L673
Could we create a bool called parsingOk that way the checks down here:
https://github.com/Yaskawa-Global/motoros2/blob/64c652409b88eab479940a9d61019ef99dfa6458/src/ActionServer_FJT.c#L720-L769
Could have a branch for parsingOk similar to the ones for positionOk

ted-miller: So add a new entry into the `Failed_Trajectory_Status` enum and give a unique error message indicating that the tolerance input was fubar?

Logically, that makes sense. But as mentioned above, the validity of the tolerances should be checked in `Ros_ActionServer_FJT_Goal_Received`. At this point, in `Ros_ActionServer_FJT_Goal_Received`, I think we should just fail with `Final position was outside tolerance`

gavanderhoorn: Ok, 'for now', I've implemented something in c552d58b248a3559c5cc5280ced8db22cdac32b1.

I don't like it too much, but it does result in a clear(er) error message than just "tolerance violation", and resolves the issue.

gavanderhoorn:
@yai-rosejo and/or @ted-miller: could I perhaps ask you to test this implementation on your hw?

ted-miller:
Definitely. I can't do it today (and probably not tomorrow), but I'll get it done. I realize this PR has lingered for quite a while now.

src/ActionServer_FJT.c
```diff
@@ -482,6 +482,142 @@ void Ros_ActionServer_FJT_ProcessFeedback()
...
+    //if caller hasn't passed any JointTolerances, set all entries to default
+    if (goal_joint_tolerances->size == 0)
+    {
+        Ros_Debug_BroadcastMsg("%s: no joint tolerances specified, returning %d defaults",
+            __func__, posTolerances_len);
+        return OK;
+    }
+
+    //it's OK if we are passed more JointTolerance instances than we have joints,
+    //as technically a client could send duplicate tolerance specs (ie: multiple
```
ted-miller, line 509: Is this common practice? It seems to me that this is a bad idea. If I accidently duplicate a joint, then I would be very confused when my intended tolerance isn't applied.

src/ActionServer_FJT.c
```diff
@@ -482,6 +482,142 @@ void Ros_ActionServer_FJT_ProcessFeedback()
     }
 }
 
+static STATUS Ros_ActionServer_FJT_Parse_GoalPosTolerances(
+    control_msgs__msg__JointTolerance__Sequence const* const goal_joint_tolerances /* in */,
+    rosidl_runtime_c__String__Sequence const* const joint_names /* in */,
+    double* posTolerances /* in/out */, size_t posTolerances_len /* in */)
```
ted-miller, line 488:
```suggestion
    double* posTolerances /* out */, size_t posTolerances_len /* in */)
```

src/ActionServer_FJT.c
```diff
@@ -482,6 +482,142 @@ void Ros_ActionServer_FJT_ProcessFeedback()
...
+                    __func__, selected_position_name.data, jtol_idx, ptol_idx);
+                posTolerances[ptol_idx] = goal_joint_tolerances->data[jtol_idx].position;
+                break;
+            }
+        }
+
+        //couldn't find joint
+        if (ptol_idx == joint_names->size)
+        {
+            //TODO(gavanderhoorn): make this a fatal error?
```
ted-miller, line 551: I assume (hope) that you mean fatal to the trajectory request (return not OK) and not a fatal assertion.

src/ActionServer_FJT.c
```diff
@@ -494,26 +630,79 @@ void Ros_ActionServer_FJT_Goal_Complete(GOAL_END_TYPE goal_end_type)
...
-                posTolerance = DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
 
-            if (diff > posTolerance)
+        //compare current state with target positions from last traj point
+        for (int axis = 0; axis < numAxesToCheck; axis += 1)
+        {
+            double current_jstate = feedback_FollowJointTrajectory.feedback.actual.positions.data[axis];
+            diff = fabs(lastTrajPtPositions[axis] - current_jstate);
+            Ros_Debug_BroadcastMsg("desired - actual: %12.8f - %12.8f: %12.8f",
+                lastTrajPtPositions[axis], current_jstate, diff);
```
ted-miller, line 690:
```suggestion
            Ros_Debug_BroadcastMsg("target pos (%12.8f) - actual pos (%12.8f) = (%12.8f)",
                lastTrajPtPositions[axis], current_jstate, diff);
```

ted-miller: Also, I'm getting zeroes again in the output (just for `lastTrajPtPositions[axis]`. I have no idea why

ted-miller: Actually, as I look closer, none of the values make sense.

```
[1715980487.96850944] [192.168.1.31:50399]: 2024-05-17 21:14:47.967896 desired - actual:   0.00000000 -   0.00046206:   0.00046206
[1715980487.96860313] [192.168.1.31:50399]: 2024-05-17 21:14:47.967896 desired - actual:   0.00000000 -  -0.00001342:   0.00001342
[1715980487.96866250] [192.168.1.31:50399]: 2024-05-17 21:14:47.967896 desired - actual:   0.00000000 -   0.00001227:   0.00001227
[1715980487.96870780] [192.168.1.31:50399]: 2024-05-17 21:14:47.968096 desired - actual:   0.00000000 -   0.00000000:   0.00000000
[1715980487.96891475] [192.168.1.31:50399]: 2024-05-17 21:14:47.968096 desired - actual:   0.00000000 -   0.00040952:   0.00040952
[1715980487.96918893] [192.168.1.31:50399]: 2024-05-17 21:14:47.968496 desired - actual:   0.00000000 -   0.00000000:   0.00000000

```

ted-miller:
Functionally, everything seems to be working well on hardware.

src/ActionServer_FJT.c
gavanderhoorn: not common practice per se. I just didn't feel like making it an issue.

We could upgrade it to one though, and return a specific error code if duplicated joint names are passed as part of the goal.

src/ActionServer_FJT.c
gavanderhoorn: Yes, even more 'local' perhaps: fatal as in: exit this function with an appropriate return code.

src/ActionServer_FJT.c
gavanderhoorn: > Also, I'm getting zeroes again in the output (just for `lastTrajPtPositions[axis]`. I have no idea why

me neither.

I was under the impression I'd 'fixed' that when referring to joint names, but apparently you still get them.

PS: you really need to update your `debug_listener.py` checkout.

ted-miller (approved):
I'm fine with merging this.

PR #244, opened 2024-04-26
Update documentation to point to new `micro_ros_motoplus` repository

gavanderhoorn:
Releases of M+ `libmicroros` will be hosted there now.

We'll need to update the wording in the release descriptions as well.


gavanderhoorn:
Releases have been updated as well.

PR #247, opened 2024-04-29
msbuild ci: match full paths in `mpBuilder` output

gavanderhoorn:
By itself this won't magically work, but with an updated mpBuilder this will allow Github to annotate the correct file at the correct line nr with the error or warning output by mpBuilder, facilitating PR reviews.

Example:

image

note the annotation is still not entirely complete as the full path is missing, but that's something we need to clear up on the mpBuilder side I believe.


ted-miller:
Which mpBuilder update are you referring to? I'm confused because there's an email eight minute before this that says it's not necessary.

gavanderhoorn:
E-mail was a bit unclear. No updates on the `mpBuilder` side are needed for this to work. The Github problem matcher does need to be updated, which is what this PR does.

ted-miller (approved):

PR #252, opened 2024-05-15
ci: test only specific upstream M+ uROS releases

gavanderhoorn:
As per title.

This lets us test specific versions instead of the full Cartesian product of all matrix entries (ie: `foxy AND galactic AND humble on dx2 AND yrc1 AND yrc1u`).

Upstream (ie: `micro_ros_motoplus`) may not support all combinations of all versions of micro-ROS on all controller platforms for specific releases, hence the need to be able to exhaustively list which specific releases should be included in the build.

This change obviously still tests `foxy AND galactic AND humble on dx2 AND yrc1 AND yrc1u`, but that's because we currently *do* have M+ uROS releases for all those combinations.

I expect that to not be true any more in the (near) future, hence this change.

### DIFF
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -57,7 +57,17 @@ jobs:
       fail-fast: false
       matrix:
         controller: [dx200, yrc1000, yrc1000u]
-        ros2_codename: [humble, galactic, foxy]
+        # officially matrix doesn't support non-scalar values, but it does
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Test build: [here](https://github.com/gavanderhoorn/motoros2/actions/runs/9094091839), which failed, but that's due to https://github.com/Yaskawa-Global/motoros2/pull/212#issuecomment-1963952062, not because of the CI config changes.

ted-miller (approved):

Issue #192, opened 2023-11-22
Incorrect error message for StartTrajMode


ted-miller:
I got the robot into an error state by pressing HOLD during an FJT routine.

```
[1700669032.41332889] [192.168.1.31:54399]: 2023-11-22 16:03:52.412220 mpExRcsIncrementMove returned -1 (Not executing WAIT instruction)
[1700669032.43213272] [192.168.1.31:54399]: 2023-11-22 16:03:52.431020 Alarm/error occurred while executing trajectory. Aborting.
[1700669032.43316460] [192.168.1.31:54399]: 2023-11-22 16:03:52.432220 Stopping all motion
[1700669032.46177649] [192.168.1.31:54399]: 2023-11-22 16:03:52.460620 FJT action complete
```

When I tried to call `/StartTrajMode`, I got:

```
ted@ted-Ubuntu:~$ ros2 service call /start_traj_mode motoros2_interfaces/srv/StartTrajMode
requester: making request: motoros2_interfaces.srv.StartTrajMode_Request()

response:
motoros2_interfaces.srv.StartTrajMode_Response(result_code=motoros2_interfaces.msg.MotionReadyEnum(value=106), message='Servo power is OFF. Please call start_traj_mode or start_point_queue_mode service')
```

This error message is incorrect and confusing.

MotoROS2 log:

```
[1700671198.53841758] [192.168.1.31:54399]: 2023-11-22 16:39:58.536889 Ros_MotionControl_StartMotionMode: enter
[1700671198.53868961] [192.168.1.31:54399]: 2023-11-22 16:39:58.537689 Controller is in a fault state. Please call /reset_error
[1700671198.53889847] [192.168.1.31:54399]: 2023-11-22 16:39:58.537889 Ros_ServiceStartTrajMode_Trigger: Servo power is OFF. Please call start_traj_mode or start_point_queue_mode service (106)
```

Issue #37, opened 2023-06-01
`start_traj_mode` fails if home position is not configured


gavanderhoorn:
If the home position is not set, or the home position needs to be checked, the controller refuses to start INFORM jobs.

For MotoROS2 specifically, this means that `start_traj_mode` & `start_point_queue_mode` fail with:

```shell
$ ros2 service call /start_traj_mode motoros2_interfaces/srv/StartTrajMode
requester: making request: motoros2_interfaces.srv.StartTrajMode_Request()

response:
motoros2_interfaces.srv.StartTrajMode_Response(result_code=motoros2_interfaces.msg.MotionReadyEnum(value=108), message='The INIT_ROS job has not started. Please call start_traj_mode service')
```

There is no indication on the pendant as to "why" the job failed to start. 

It is unclear whether this issue can be detected, or whether MotoROS2 can provide a better error message.

gavanderhoorn:
Note: this is also an issue with MotoROS1.

ted-miller:
We need to document this error in the readme.
https://github.com/Yaskawa-Global/motoros2/discussions/201#discussioncomment-8208576

They should be seeing a message on the pendant 
![image](https://github.com/Yaskawa-Global/motoros2/assets/4623811/ff8c1900-dbb7-400c-90a2-5ba080e47314)

When attempting to start a job (using the pendant), they should get error 0380 Check Position

gavanderhoorn:
Should we instead not have MR2 check whether the home pos is set and raise an appropriate alarm if it isn't?

ted-miller:
It looks like we already are
https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/MotionControl.c#L1514

So that's apparently broken.

gavanderhoorn:
You mean we should be seeing:

https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/ErrorHandling.c#L27

?

ted-miller:
Yeah. That's what I would have thought.

Though perhaps 0x3040 doesn't apply if the home position has been registered and just needs to be validated. If that's the case, then I don't know of any mechanism to detect that this scenario.

gavanderhoorn:
Just ran into this when testing something for #19:

```
[...] [...]: Ros_MotionControl_StartMotionMode: enter
[...]

PR #265, opened 2024-06-19
More accurate/descriptive error messages for start_traj_mode and start_point_queue_mode services

jimmy-mcelwain:
In reference to #37 and #192. Now the error message and error codes match the reasons that the motion start failed.


jimmy-mcelwain:
Dependent upon https://github.com/Yaskawa-Global/motoros2_interfaces/pull/17

gavanderhoorn:
High-level comment (about the proposed approach): the current implementation seems to use a different approach: error reporting uses the result of `Ros_Controller_GetNotReadySubcode()` (based on the result of `Ros_Controller_IsMotionReady()`) if `Ros_MotionControl_StartMotionMode(..)` (or anything else) failed.

The current approach is slightly racy (as in: noticing *something* was wrong and reporting about *what* was wrong are done at different times -- but so far that hasn't really been problematic), but the proposed approach starts mixing responsibilities. It also duplicates quite a bit of code.

Perhaps we should try and find a middle-ground?

gavanderhoorn:
And another (but only slightly related to this PR): seems the new error / pattern matcher in combination with the updated `mpBuilder` doesn't yet capture compiler errors correctly, resulting in the many incomplete annotations.

I'll look into this and fix it.

jimmy-mcelwain:
@gavanderhoorn This is sort of an awkward responsibility-mixing design. I was trying to solve the issue without any sort of big redesign of how `Ros_Controller_GetNotReadySubcode()` is used. 

I do think that the current way that errors are reported, by noticing that there is an issue, and then trying to identify that issue later based on the state of the controller, seems likely to cause to problems/makes it difficult to get useful error messages. I think that a more direct error-reporting approach may be more reliable/less 'racy', but yeah there would be more code duplication. 

More significant changes would need to be made to get rid of the responsibility-mixing aspect, or I could move closer to the original approach. but I don't know what the best option is. I pushed some minor changes that make it a bit cleaner, but didn't change the approach.

src/MotionControl.c
```diff
@@ -1355,23 +1355,25 @@ static STATUS Ros_Controller_DisableEcoMode()
...
-BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
+MotionNotReadyCode Ros_MotionControl_StartMotionMode(MOTION_MODE mode, rosidl_runtime_c__String *responseMessage)
 {
     int ret;
     MP_STD_RSP_DATA rData;
     MP_START_JOB_SEND_DATA sStartData;
     int checkCount;
     int grpNo;
+    MotionNotReadyCode motion_readiness_code = MOTION_NOT_READY_UNSPECIFIED;
+    char output[256] = { 0 };
```
ted-miller, line 1368: The `256` needs to be a named constant. It is used in multiple places below.

Additionally, I'm not sure that the `output` buffer is really needed. If things are checked in the correct order, then `Ros_ErrorHandling_MotionNotReadyCode_ToString` should contain sufficient information to the user _(I think)_. In that case, you shouldn't need to pass in the pointer to `responseMessage`.

Do you have some example strings from testing that you can post?

ted-miller (changes requested):
I'm also not a big fan of the repeated code.

What if in `Ros_MotionControl_StartMotionMode`, the first thing you do is call `Ros_Controller_GetNotReadySubcode`? If it returns a code that cannot be resolved by the actions of this function (like estop or play-mode), then return that code immediately. That should remove some of the duplication.

I expect that such an approach would also need to ensure that things are checked in a particular order within `Ros_Controller_GetNotReadySubcode`. (Unresolvable issues checked first)

PR #271, opened 2024-06-27
ci: use GitHub CLI to download M+ libmicroros

gavanderhoorn:
As per title.

This switches to the [GH CLI](https://cli.github.com/) for interaction with releases of `micro_ros_motoplus` from the currently used `dsaltares/fetch-gh-release-asset`.

This works just as well and also supports downloading assets from draft releases (which we'll need later).


gavanderhoorn:
Unfortunately the (misconfigured) `msbuild_filter` job prevented the `msbuild` job from running, but a build with this can be found [here](https://github.com/Yaskawa-Global/motoros2/actions/runs/9694483128). Be sure to choose one of the builds and expand the *Download M+ libmicroros (...)* step.

(note: that build failed, but not because of the changes in this PR, just the known issues with the `20221102` release of M+ `libmicroros`)

ted-miller (approved):

Issue #278, opened 2024-07-15
Add support for INFORM job CRUD


gavanderhoorn:
From [the roadmap](https://github.com/Yaskawa-Global/motoros2/blob/f61a7238d78066aa565542b6fefd112bad42d0cb/README.md#provisional-roadmap):

> The following items are on the MotoROS2 roadmap, and are listed here in no particular order:
>
> - [..]
> - CRUD of INFORM job files (ie: create, retrieve, update, delete)
> - [..]

In addition to the listed functionality, we'd also need to be able to *list* existing jobs, as the other operations will need names of jobs to operate on.

gavanderhoorn:
Current design of the services: [here](https://github.com/gavanderhoorn/motoros2_interfaces/tree/inform_crud) ([diff](https://github.com/Yaskawa-Global/motoros2_interfaces/compare/main...gavanderhoorn:motoros2_interfaces:inform_crud)).

gavanderhoorn:
Current WIP of everything except `inform_job/list`: https://github.com/Yaskawa-Global/motoros2/compare/inform_job_crud_list...gavanderhoorn:motoros2:inform_job_crud_the_rest.

Service definitions: https://github.com/Yaskawa-Global/motoros2_interfaces/compare/main...gavanderhoorn:motoros2_interfaces:inform_crud.

gavanderhoorn:
I've been trying to test my latest changes (adding 'the rest' of the CRUD operations), but for some reason none of the new services 'work'. The client just hangs. No output on the server side (in the debug log).

Almost as-if the executor isn't processing the events. Removing the `list` service from the executor results in the same behaviour (of that service).

I've compared the code (https://github.com/Yaskawa-Global/motoros2/compare/inform_job_crud_list...gavanderhoorn:motoros2:inform_job_crud_the_rest) to other service additions, but can't really find anything incorrect.

I'm probably missing something. @ted-miller: could you maybe take a look and see whether you can spot anything?

ted-miller:
It seems that the buffer for the incoming `name` field needs to be allocated in `Ros_ServiceGetInformJob_Initialize` (and other init functions)

```
    rosidl_runtime_c__String__init(&g_messages_GetInformJob.request.name);
    rosidl_runtime_c__String__assign(&g_messages_GetInformJob.request.name, "this is a string to allocate a buffer");
```

I'm not really sure why `assign` was also needed. But it worked after I added that.

[...]

PR #279, opened 2024-07-15
Add INFORM job `list` service

gavanderhoorn:
Work towards #278.

This implements a basic `job/list` function which returns the names of all jobs (`*.JBI`) on the controller.

I've submitted this as a small PR to avoid opening something for all 5 services at once. It is fully functional though, hence this is not a *draft*.

There are a couple of `TODO`s which I'd like to discuss.

Note: this depends on new messages in `motoros2_interfaces`: Yaskawa-Global/motoros2_interfaces#21.


gavanderhoorn:
Note: CI will fail as there is no build of `micro_ros_motoplus` with the changes in https://github.com/Yaskawa-Global/motoros2_interfaces/compare/main...gavanderhoorn:motoros2_interfaces:inform_crud.

src/RosApiNameConstants.h
```diff
@@ -28,6 +28,7 @@
 #define SERVICE_NAME_STOP_TRAJ_MODE "stop_traj_mode"
 #define SERVICE_NAME_QUEUE_TRAJ_POINT "queue_traj_point"
 #define SERVICE_NAME_SELECT_MOTION_TOOL "select_motion_tool"
+#define SERVICE_NAME_LIST_JOBS "job/list"
```
gavanderhoorn, line 31: I've opted to create a `job` namespace here, as the full implementation of #278 will add 4 more services. The namespace groups them all using a single `job` prefix.

src/ServiceJobList.c
```diff
@@ -0,0 +1,143 @@
...
+#include "MotoROS.h"
+
+rcl_service_t g_serviceListJobs;
+
+ServiceListJobs_Messages g_messages_ListJobs;
+
+typedef motoros2_interfaces__srv__ListJobs_Response ListJobsResponse;
+
+
+static const size_t RESP_NAMES_LIST_INITIAL_SIZE = 64;
```
gavanderhoorn, line 17: I don't expect MotoROS2 users to have too many regular INFORM jobs, so this is probably as good a default/starting value as any, but we could obviously go with a higher number.

src/ServiceJobList.c
```diff
@@ -0,0 +1,143 @@
...
+
+    rcl_ret_t ret = rclc_service_init_default(
+        &g_serviceListJobs, &g_microRosNodeInfo.node, type_support, SERVICE_NAME_LIST_JOBS);
+    motoRosAssert_withMsg(ret == RCL_RET_OK,
+        SUBCODE_FAIL_INIT_SERVICE_LIST_JOBS, "Failed to init service (%d)", (int)ret);
+
+    rosidl_runtime_c__String__init(&g_messages_ListJobs.response.message);
+    //NOTE: if/when needed, this will be resized
+    rosidl_runtime_c__String__Sequence__init(
+        &g_messages_ListJobs.response.names, RESP_NAMES_LIST_INITIAL_SIZE);
```
gavanderhoorn, line 35: The idea is to initialise the response with the default/initial list size and resize later in the service callback if/when needed.

src/ServiceJobList.c
```diff
@@ -0,0 +1,143 @@
...
+
+    //make sure we don't accidentally return old results (in case of errors fi)
+    rosidl_runtime_c__String__assign(&response->message, "");
+    response->result_code = 0;
+    response->names.size = 0;
+
+    //retrieve list of jobs (".JBI")
+    //TODO(gavanderhoorn): should this also look for other files (JBR)?
+    Ros_Debug_BroadcastMsg("%s: updating job file list", __func__);
+    LONG status = mpRefreshFileList(MP_EXT_ID_JBI);
```
gavanderhoorn, line 71: @ted-miller: would it make sense to include `JBR`s as well? `mpRefreshFileList(..)` has a constant for those, so we could do that.

src/ServiceJobList.c
```diff
@@ -0,0 +1,143 @@
...
+        goto DONE;
+    }
+    Ros_Debug_BroadcastMsg("%s: found %d jobs", __func__, job_count);
+    if (job_count > response->names.capacity)
+    {
+        Ros_Debug_BroadcastMsg("%s: resize needed: current: %d; need: %d",
+            __func__, job_count, response->names.capacity);
+        rosidl_runtime_c__String__Sequence__fini(&response->names);
+        //TODO(gavanderhoorn): handle errors
+        rosidl_runtime_c__String__Sequence__init(&response->names, job_count);
```
gavanderhoorn, line 98: this obviously (re)allocates, so there is a (small) chance of running out of memory. But it would depend on how many jobs there actually are.

@ted-miller: would we perhaps need to define an upper-limit? With 64 elements in `names`, the service allocated about 2.5 KB at service init (based on the memory tracing logs).

Max job name length is about 40 bytes, so with 10k jobs (fi), this would use approximately 400 KB of memory.

src/ServiceJobList.c
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    //copy all jobs on the list to the response
+    //32 characters for the name, plus 4 for extension, plus terminating null
+    //TODO(gavanderhoorn): is this really necessary? Copied from inform generator
+    //TODO(gavanderhoorn): confirm whether extension is included or not
```
gavanderhoorn, line 104: `mpGetFileName(..)` does appear to include the job extension.

@ted-miller: if we're only going to support `JBI` management, could we perhaps remove the extension?

src/RosApiNameConstants.h
ted-miller: Is `job` too generic? Would `inform_job` be better, or too verbose/unnecessary?

src/ServiceJobList.c
ted-miller: Personally, I think it's better to init to 1, then resize based on the value returned by `mpGetFileCount`. I don't see benefit to preallocating 64.

src/ServiceJobList.c
ted-miller: TBH, I don't even know what a JBR is. I can't find any relevant documentation and have never heard of anyone using them.

I'm ok with just allowing JBI

src/ServiceJobList.c
ted-miller: > I assume the maximum nr of jobs on a single controller is actually quite high?

Yeah.  The spec is for 200,000 motion steps and 10,000 non-motion instructions. I don't imagine someone have a 100,000 two-step jobs... but the possibility exists.
(I have seen people run out of job space before.)

src/ServiceJobList.c
ted-miller: > would we perhaps need to define an upper-limit? With 64 elements in names, the service allocated about 2.5 KB at service init (based on the memory tracing logs).

I agree there should be a limit. 

I have seen scenarios where people max out job space. Though I don't know the actual qty of jobs.

In my mind, it's reasonable to limit to something like 500-1000. ROS2 is not intended to be a job manager. We have other products for that.

src/ServiceJobList.c
ted-miller: Sure

src/RosApiNameConstants.h
gavanderhoorn: We could go for `inform_job`, but are there other types of jobs? If not, I'd suggest `job` would be sufficient.

I've also considered `program` and other more generic names (`robot_program`, `native_program`, `oem_program`, etc) -- so as to not tie this to Yaskawa terminology immediately -- but couldn't come up with something that really stuck.

src/ErrorHandling.h
```diff
@@ -169,6 +169,8 @@ typedef enum
     SUBCODE_CONFIGURATION_FAIL_MP_NICDATA1,
     SUBCODE_FAIL_MP_NICDATA_INIT1,
     SUBCODE_FAIL_INVALID_BASE_TRACK_MOTION_TYPE,
+    SUBCODE_FAIL_INIT_SERVICE_LIST_JOBS,
+    SUBCODE_FAIL_ADD_SERVICE_LIST_JOBS,
```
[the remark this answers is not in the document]
gavanderhoorn: I need to check why CI didn't flag these as undocumented.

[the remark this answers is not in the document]
gavanderhoorn: This isn't getting flagged as there is a *catch-all* section for `8011` ([here](https://github.com/Yaskawa-Global/motoros2/blob/f61a7238d78066aa565542b6fefd112bad42d0cb/doc/troubleshooting.md#alarm-8011xx)), and CI isn't setup to complain about those (as otherwise it would always complain, seeing as we're 'relying' on these catch-alls).

These `SUBCODE_FAIL_INIT_*` and `SUBCODE_FAIL_ADD_*` subcodes are all in `8011`, hence no errors raised.

Adding `--no-catch-alls` to the linter's configuration results in:

```
src/ErrorHandling.h:94:5: error: no documentation for '8010[0]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:95:5: error: no documentation for '8010[1]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:97:5: error: no documentation for '8010[3]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:102:5: error: no documentation for '8011[0]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:103:5: error: no documentation for '8011[1]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:104:5: error: no documentation for '8011[2]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:105:5: error: no documentation for '8011[3]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:106:5: error: no documentation for '8011[4]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:107:5: error: no documentation for '8011[5]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:108:5: error: no documentation for '8011[6]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:109:5: error: no documentation for '8011[7]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:110:5: error: no documentation for '8011[8]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:111:5: error: no documentation for '8011[9]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:113:5: error: no documentation for '8011[10]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:114:5: error: no documentation for '8011[11]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:115:5: error: no documentation for '8011[12]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:116:5: error: no documentation for '8011[13]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:117:5: error: no documentation for '8011[14]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:121:5: error: no documentation for '8011[18]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:172:5: error: no documentation for '8011[64]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:173:5: error: no documentation for '8011[65]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:179:5: error: no documentation for '8012[0]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:180:5: error: no documentation for '8012[1]' in 'doc/troubleshooting.md'
src/ErrorHandling.h:181:5: error: no documentation for '8012[2]' in 'doc/troubleshooting.md'
```

src/RosApiNameConstants.h
gavanderhoorn: Thinking about it a bit more, I believe I do actually prefer `inform_job` instead of just `job`. It's indeed much clearer from a ROS API perspective.

I'll update the PR.

src/ServiceJobList.c
gavanderhoorn: > Personally, I think it's better to init to 1, then resize based on the value returned by `mpGetFileCount`. I don't see benefit to preallocating 64.

less runtime allocations?

> > I assume the maximum nr of jobs on a single controller is actually quite high?
> 
> Yeah. The spec is for 200,000 motion steps and 10,000 non-motion instructions. I don't imagine someone have a 100,000 two-step jobs... but the possibility exists. (I have seen people run out of job space before.)

So the limits are for steps in a job, not job files themselves? Or am I misinterpreting this?

src/ServiceJobList.c
gavanderhoorn: Ok, I'll add an (artificial) limit.

We should of course document it clearly then.

Probably even return a non-success `result_code` in case a response only contains the first `whatever_the_limit_is_going_to_be` names.

src/ServiceJobList.c
ted-miller: > So the limits are for steps in a job, not job files themselves

Correct. 

There may be another limit for qty of files, due to the header overhead. But I can't find it anywhere.

gavanderhoorn:
I've addressed most of the discussed changes in functionality. Tested it here as well and it still works.

Just need to test behaviour when there are more jobs than allowed, define proper `result_codes` and document it all.

src/ErrorHandling.h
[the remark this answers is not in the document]
gavanderhoorn: So technically, these new subcodes are already documented. Just like many others in this category, by the `8011[xx]` catch-all.

Would that be sufficient for you @ted-miller?

doc/ros_api.md
```diff
@@ -73,6 +73,17 @@ Note: this topic is only namespaced if a namespace is configured *and* `namespac
 
 ## Services
 
+### inform_job/list
+
+Type: [motoros2_interfaces/srv/ListJobs](https://github.com/gavanderhoorn/motoros2_interfaces/blob/a7cac24d5a22ff756aeb3f3a53ea9cb2258c3e5b/srv/ListJobs.srv)
```
gavanderhoorn, line 78: note: this points to my fork for now.

src/ErrorHandling.h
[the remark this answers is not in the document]
ted-miller: Yeah, I forgot about the [xx]

doc/ros_api.md
ted-miller: Just curious why it's on your fork instead of a pending branch of y-global?

ted-miller (commented):
I don't see any issues. But I'm struggling to build right now. 

So, If you're happy and have tested it, then feel free to merge. Otherwise, I'll get my stuff fixed on monday and actually run it.

doc/ros_api.md
gavanderhoorn: No particular reason.

Seeing as we should merge a PR to `motoros2_interfaces` adding this definition before this PR anyway, it shouldn't really matter, as I'd update this PR with the correct location.

gavanderhoorn:
I have tested, but I'd like some independent confirmation I'm not making (implicit) assumptions about how things work based on my controller configuration.

I'll make a build of `micro_ros_motoplus` available to you with the needed interfaces.

ted-miller:
I have fixed my build.

ted-miller (dismissed):
Works well. I also verified that it lists macro jobs and system jobs.

Japanese names come across as gibberish. But this is to be expected. (https://github.com/Yaskawa-Global/motoros2_interfaces/pull/21#discussion_r1687116762)

gavanderhoorn:
@ted-miller: I believe this is the final version of this PR.

Switched to named result codes and strings, renamed files to match the fact it's now specifically about *INFORM* jobs and updated all references.

If you could take a last look?

To make it easier: [20240723_mr2_yrc1_h_inform_list_rc2.7z.zip](https://github.com/user-attachments/files/16350802/20240723_mr2_yrc1_h_inform_list_rc2.7z.zip).

This should not require any updates on the `motoros2_interfaces` side, but just in case: I updated Yaskawa-Global/motoros2_interfaces#21.

ted-miller:
I'll give this a run first-thing tomorrow

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+
+ServiceListInformJobs_Messages g_messages_ListInformJobs;
+
+typedef motoros2_interfaces__srv__ListInformJobs_Response ListInformJobsResponse;
+
+
+static const size_t RESP_NAMES_LIST_INITIAL_SIZE = 1;
+static const size_t MAX_NUMBER_OF_JOBS_TO_REPORT = 1024;
+
+#define MSG_RC(x) (motoros2_interfaces__msg__InformJobCrudResultCodes__RC_ ## x)
```
ted-miller, line 20: You're giving me flashbacks to MFC programming w/ these macros.

ted-miller (dismissed):
Looks good. Tested on YRC. But since you're using vanilla M+ API's, I'm not worried about other controllers.

src/ServiceInformJobList.c
gavanderhoorn: :)

ted-miller:
@gavanderhoorn Do you want to merge this, or pull `inform_job_crud_the_rest` into it first?

jimmy-mcelwain:
Rebased because I'd like to get this merged and there were some conflicts. I'll take a look at it

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+    {
+        Ros_Debug_BroadcastMsg("%s: error refreshing job list", __func__);
+        rosidl_runtime_c__String__assign(&response->message, MSG_RC_STR(ERR_REFRESHING_JOB_LIST));
+        response->result_code = MSG_RC(ERR_REFRESHING_JOB_LIST);
+        goto DONE;
+    }
+
+    //see how many jobs there are, and if necessary, resize list of names in service response
+    size_t job_count = mpGetFileCount();
+    if (job_count < 0)
```
jimmy-mcelwain, line 88: `size_t` is a `long unsigned int`, so this will never evaluate to true

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+
+    Ros_Debug_BroadcastMsg("%s: maximum nr of jobs supported: %d",
+        __func__, MAX_NUMBER_OF_JOBS_TO_REPORT);
+
+    const rosidl_service_type_support_t* type_support =
+        ROSIDL_GET_SRV_TYPE_SUPPORT(motoros2_interfaces, srv, ListInformJobs);
+
+    rcl_ret_t ret = rclc_service_init_default(
+        &g_serviceListInformJobs, &g_microRosNodeInfo.node, type_support, SERVICE_NAME_LIST_INFORM_JOBS);
+    motoRosAssert_withMsg(ret == RCL_RET_OK,
```
jimmy-mcelwain, line 36: Can be changed to `motoRos_RCLAssertOK_withMsg()`

src/CommunicationExecutor.c
```diff
@@ -422,6 +422,12 @@ void Ros_Communication_StartExecutors(SEM_ID semCommunicationExecutorStatus)
         &g_messages_ReadWriteIO.resp_mreg_write, Ros_ServiceWriteMRegister_Trigger);
     motoRos_RCLAssertOK_withMsg(rc, SUBCODE_FAIL_ADD_SERVICE_WRITE_M_REG, "Failed adding service (%d)", (int)rc);
 
+    //INFORM job CRUD services
+    rc = rclc_executor_add_service(
+        &executor_io_control, &g_serviceListInformJobs, &g_messages_ListInformJobs.request,
+        &g_messages_ListInformJobs.response, Ros_ServiceListInformJobs_Trigger);
+    motoRosAssert_withMsg(rc == RCL_RET_OK, SUBCODE_FAIL_ADD_SERVICE_LIST_INFORM_JOBS, "Failed adding service (%d)", (int)rc);
```
jimmy-mcelwain, line 429: Can be changed to `motoRos_RCLAssertOK_withMsg()`

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+            __func__, response->names.capacity, job_count);
+        rosidl_runtime_c__String__Sequence__fini(&response->names);
+        //TODO(gavanderhoorn): handle errors
+        rosidl_runtime_c__String__Sequence__init(&response->names, job_count);
+    }
+
+    //copy all jobs on the list to the response
+    //name, plus extension, plus terminating null
+#define JOB_EXT_LEN (4)
+#define MAX_JOB_NAME_LEN_WITH_EXT (MAX_JOB_NAME_LEN + JOB_EXT_LEN + 1)
```
jimmy-mcelwain, line 119: Already defined here
https://github.com/Yaskawa-Global/motoros2/blob/2007fd696811b6967ec1a17574c30ae2da17f0f1/src/InformCheckerAndGenerator.c#L11
Maybe best to bring to shared header

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
+//ServiceJobList.c
```
jimmy-mcelwain, line 1: Stale file name

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+ServiceListInformJobs_Messages g_messages_ListInformJobs;
+
+typedef motoros2_interfaces__srv__ListInformJobs_Response ListInformJobsResponse;
+
+
+static const size_t RESP_NAMES_LIST_INITIAL_SIZE = 1;
+static const size_t MAX_NUMBER_OF_JOBS_TO_REPORT = 1024;
+
+#define MSG_RC(x) (motoros2_interfaces__msg__InformJobCrudResultCodes__RC_ ## x)
+#define MSG_RC_STR(x) (motoros2_interfaces__msg__InformJobCrudResultCodes__STR_ ## x)
```
jimmy-mcelwain, line 21: We'll want these in a shared header eventually since other inform CRUD services will use them.

src/ServiceInformJobList.c
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    //resize
+    if (job_count > response->names.capacity)
+    {
+        Ros_Debug_BroadcastMsg("%s: resize needed: current: %d; need: %d",
+            __func__, response->names.capacity, job_count);
+        rosidl_runtime_c__String__Sequence__fini(&response->names);
+        //TODO(gavanderhoorn): handle errors
+        rosidl_runtime_c__String__Sequence__init(&response->names, job_count);
```
jimmy-mcelwain, line 113: Would we want to have the error handling be like what we have here?
https://github.com/Yaskawa-Global/motoros2/blob/2007fd696811b6967ec1a17574c30ae2da17f0f1/src/PositionMonitor.c#L188-L189
Or we could add a new service response "Failed to allocate memory"

jimmy-mcelwain (changes requested):
Notably, this uses the same executor as the I/O services. Maybe that is okay, but maybe it would be good to create a new executor (e.g. `inform_job_executor`) since I think that these service calls are probably much more CPU intensive than I/O service calls and may end up blocking important calls, especially once we add all of the other INFORM CRUD operations. 

I am not set on that, but I think it is worth discussing.

jimmy-mcelwain:
I've spent some time trying to figure out why the DX200 builds have failed. I am able to build them successfully locally using the same libmicroros and headers at the CI is using. My best guess is that it is using an outdated MotoPlus SDK and there are some headers missing from the DX200 version that are present in the YRC versions.

src/ServiceInformJobList.c
gavanderhoorn: You mean something like [this](https://github.com/Yaskawa-Global/motoros2/compare/inform_job_crud_list...gavanderhoorn:motoros2:inform_job_crud_the_rest#diff-b67b28dabcbe475a52035fc93166664afd746015906e4a8f200a8b356bd3ce0c)?

---

Edit: that [diff](https://github.com/Yaskawa-Global/motoros2/compare/inform_job_crud_list...gavanderhoorn:motoros2:inform_job_crud_the_rest) is a bit strange. I guess because of the rebase here.

gavanderhoorn:
@jimmy-mcelwain: the problem is a step earlier:

```
7-Zip 26.02 (x64) : Copyright (c) 1999-2026 Igor Pavlov : 2026-06-25

Scanning the drive for archives:
1 file, 0 bytes

Extracting archive: C:\build\mpsdk_ci.7z
ERROR: C:\build\mpsdk_ci.7z
Cannot open the file as archive


Can't open as archive: 1
Files: 0
Size:       0
Compressed: 0
```

looks like it's failing to download the SDK, which of course means the build fails immediately.

Don't know why that would only affect the DX200 builds though.

jimmy-mcelwain:
> Don't know why that would only affect the DX200 builds though.

@ted-miller figured it out. We think is not actually only affecting the DX200 builds, that was just a coincidence. The selection of jobs that it fails is seemingly random. We think it is a firewall thing where we are hosting the SDK and are trying to figure it out with IT now. 

--------------------------------
Edit see [run 1](https://github.com/Yaskawa-Global/motoros2/actions/runs/28965243299) and [run 2](https://github.com/Yaskawa-Global/motoros2/actions/runs/28960976066)

gavanderhoorn:
> We think it is a firewall thing where we are hosting the SDK and are trying to figure it out with IT now.

that was going to be my next comment as well.

I remembered it's hosted at YAI and I suspected a changed network policy / server configuration.

12 almost concurrent download requests for the same file might trigger some rate limiting somewhere.

---

Edit: restored SDK download URL and that seems to have turned CI green again (for now).

src/ServiceInformJobList.c
jimmy-mcelwain: Yes. I forgot about that branch.

src/ServiceInformJobList.c
gavanderhoorn: No, I would not assert in a service call. Not being able to allocate memory is serious, but in this case not serious enough to warrant shutting down.

It could be there just isn't enough free memory available to allocate sufficient entries in the `String` sequence for `job_count` jobs. That should make the service fail, not the application (as it's recoverable, we just won't be able to fulfil the service's contract).

jimmy-mcelwain: Sounds good. Nothing in [the result code enum](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/ff22d758205cec27ed4d93454ba47f8a2748a2f2/msg/InformJobCrudResultCodes.msg#L9-L10) matches particularly well. Maybe this should be added

src/ServiceInformJobList.c
gavanderhoorn: I'm not sure who was 'first'. I had that code lying around for a long time before I committed it.

But agreed.

src/ServiceInformJobList.c
gavanderhoorn: The enum will need to be extended anyway. https://github.com/Yaskawa-Global/motoros2/compare/inform_job_crud_list...gavanderhoorn:motoros2:inform_job_crud_the_rest uses many magic nrs right now.

PR #280, opened 2024-07-17
ci: complain about catch-alls documenting alarm (sub) codes

gavanderhoorn:
This won't mark the job as failed, but it will result in annotations being added to a PR or CI run.

Example:

image

---

Edit: and on a PR we'd get:

image

But only the first 10 warnings.

### DIFF
--- a/.ci/alarm_subcode_doc_linter.py
+++ b/.ci/alarm_subcode_doc_linter.py
@@ -132,7 +132,7 @@ def main():
         )
         if doc_range:
             if doc_range.is_catch_all and args.warn_catch_alls:
-                print(f"{sdef.loc_str}: warning: documented by catch-all '{doc_range}'")
+                print(f"{sdef.loc_str}: warning: '{sdef}' documented by catch-all '{doc_range}'")
         else:
             ret = 1
             print(f"{sdef.loc_str}: error: no documentation for '{sdef}' in '{args.md_file}'")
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Context: #279 (specifically: https://github.com/Yaskawa-Global/motoros2/pull/279#discussion_r1678323164).

@ted-miller: turns out I'd already added the possibility to emit warnings about catch-all ranges in the documentation, we just hadn't enabled it during the CI run.

This enables it and slightly improves the warning text to also include the specific alarm (sub) code which is covered by a catch-all.

Note: unfortunately we have quite a few alarms documented by catch-alls, and the run summary only shows up to 10.

gavanderhoorn:
Friendly ping @ted-miller

ted-miller:
> unfortunately we [have quite a few alarms documented by catch-alls](https://github.com/Yaskawa-Global/motoros2/actions/runs/9975098198/job/27564025972?pr=280#step:6:20), and the run summary only shows up to 10.

Are you sure we want this then? If there are other warnings, would they get smothered by this message?

gavanderhoorn:
> Are you sure we want this then? If there are other warnings, would they get smothered by this message?

so, yes, I would want this, but perhaps the script that checks these things should be extended to allow for ignoring certain alarm ranges/IDs.

ted-miller:
Ok. It's fine with me then. 

> but perhaps the script that checks these things should be extended to allow for ignoring certain alarm ranges/IDs

Not sure why you switched back to a Draft PR. Are you making such a change?

gavanderhoorn:
> Are you making such a change?

yeah, I'll add it.

gavanderhoorn:
I'm going to close this, as I won't have time to address the outstanding issue/enhancement (ie: add ignore argument to script).

we can pick this up at a later time if needed/desired.

PR #289, opened 2024-08-06
Bump version to 0.1.4pre

gavanderhoorn:
As per title.

Facilitate distinguishing between from-source builds and regular released binaries.


gavanderhoorn:
We might end up with `0.2.0` as a release, but we can at least already bump this away from `0.1.3`.

ted-miller (approved):

PR #296, opened 2024-08-28
reset_error: print correct char field, not struct

gavanderhoorn:
A `rosidl_runtime_c__String` is a `struct`, not a plain `char*`.

On some architectures/controller platforms it's OK to pass a pointer to the `struct` (apparently), as the first member is the `char*`, but at least on FS100 this doesn't work and results in garbage output to the debug log.


gavanderhoorn:
There might be more places where we do this, but I happened to notice this one and it made me distrust the FS1 build enough to fix it.

ted-miller (approved):

PR #297, opened 2024-08-29
Restructuring error reporting in start_traj_mode and start_point_queue_mode 

jimmy-mcelwain:
Changes start_traj_mode and start_point_queue_mode services so that they properly report the reasons for failure in case motion cannot be started. This addresses many of the concerns brought up in the discussion of PR #265. I closed that PR and opened this to simplify bringing it up to date with main and to make the history less confusing.

### DIFF
--- a/src/ControllerStatusIO.h
+++ b/src/ControllerStatusIO.h
@@ -129,7 +129,7 @@ extern BOOL Ros_Controller_IsPflActive();
 extern BOOL Ros_Controller_IsMpIncMoveErrorActive();
 extern BOOL Ros_Controller_IsAnyFaultActive();
 extern BOOL Ros_Controller_MasterTaskIsJobName(const char* const jobName);
-extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode(bool attemptingToStartMotion);
+extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode(bool ignoreTractableProblems);
 
 //reset internal flag indicating whether PFL became active during a move
 extern void Ros_Controller_Reset_PflDuringRosMove();
--- a/src/ControllerStatusIO.h
+++ b/src/ControllerStatusIO.h
@@ -129,7 +129,7 @@ extern BOOL Ros_Controller_IsPflActive();
 extern BOOL Ros_Controller_IsMpIncMoveErrorActive();
 extern BOOL Ros_Controller_IsAnyFaultActive();
 extern BOOL Ros_Controller_MasterTaskIsJobName(const char* const jobName);
-extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode();
+extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode(bool attemptingToStartMotion);
 
 //reset internal flag indicating whether PFL became active during a move
 extern void Ros_Controller_Reset_PflDuringRosMove();
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -1364,7 +1364,7 @@ MotionNotReadyCode Ros_MotionControl_StartMotionMode(MOTION_MODE mode, rosidl_ru
     MP_START_JOB_SEND_DATA sStartData;
     int checkCount, grpNo, alarmcode;
     MotionNotReadyCode motion_readiness_code;
-    char output[MOTION_START_ERROR_MESSSAGE_LENGTH] = { 0 };
+    char output[MOTION_START_ERROR_MESSAGE_LENGTH] = { 0 };
 
     Ros_Debug_BroadcastMsg("%s: enter", __func__);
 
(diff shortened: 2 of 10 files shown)


jimmy-mcelwain:
Also in reference to #37 and #192

src/ControllerStatusIO.h
```diff
@@ -128,7 +128,7 @@ extern BOOL Ros_Controller_IsPflActive();
 extern BOOL Ros_Controller_IsMpIncMoveErrorActive();
 extern BOOL Ros_Controller_IsAnyFaultActive();
 extern BOOL Ros_Controller_MasterTaskIsJobName(const char* const jobName);
-extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode();
+extern MotionNotReadyCode Ros_Controller_GetNotReadySubcode(bool attemptingToStartMotion);
```
ted-miller, line 131: The argument name doesn't match what's in the .c file

src/MotionControl.h
```diff
@@ -12,6 +12,7 @@
 
 #define MOTION_START_TIMEOUT                5000  // in milliseconds
 #define MOTION_START_CHECK_PERIOD           50  // in millisecond
+#define MOTION_START_ERROR_MESSSAGE_LENGTH  256
```
ted-miller, line 15:
```suggestion
#define MOTION_START_ERROR_MESSAGE_LENGTH  256
```

src/MotionControl.c
```diff
@@ -1355,79 +1355,76 @@ static STATUS Ros_Controller_DisableEcoMode()
...
+        case MOTION_NOT_READY_HOLD:
+            return motion_readiness_code;
+        case MOTION_NOT_READY_ERROR:
+            alarmcode = Ros_Controller_GetAlarmCode();
+            snprintf(output, MOTION_START_ERROR_MESSSAGE_LENGTH, "%s: '%s' (0x%04X)",
+                Ros_ErrorHandling_MotionNotReadyCode_ToString(MOTION_NOT_READY_ERROR),
+                Ros_ErrorHandling_ErrNo_ToString(alarmcode),
+                alarmcode);
+            Ros_Debug_BroadcastMsg(output);
+            rosidl_runtime_c__String__assign(responseMessage, output);
```
ted-miller, line 1402: Is it intentional to fall through and not return here?

src/MotionControl.c
```diff
@@ -1355,79 +1355,76 @@ static STATUS Ros_Controller_DisableEcoMode()
...
-        if (!Ros_Controller_MasterTaskIsJobName(g_nodeConfigSettings.inform_job_name))
-        {
+        case MOTION_NOT_READY_ESTOP:
+        case MOTION_NOT_READY_NOT_PLAY:
+        case MOTION_NOT_READY_NOT_REMOTE:
+        case MOTION_NOT_READY_HOLD:
+            return motion_readiness_code;
+        case MOTION_NOT_READY_ERROR:
+            alarmcode = Ros_Controller_GetAlarmCode();
+            snprintf(output, MOTION_START_ERROR_MESSSAGE_LENGTH, "%s: '%s' (0x%04X)",
```
ted-miller, line 1397: I don't understand why this is the only case where the return string gets populated

src/MotionControl.c
```diff
@@ -1557,30 +1567,26 @@ BOOL Ros_MotionControl_StartMotionMode(MOTION_MODE mode)
...
 
-updateStatus:
-    // Update status
     Ros_Controller_IoStatusUpdate();
 
     Ros_Debug_BroadcastMsg("%s: exit", __func__);
 
     //Required to allow motion api to work (Potential race condition)
     Ros_Sleep(200);
+    motion_readiness_code = Ros_Controller_GetNotReadySubcode(false);
```
ted-miller, line 1576: What's the purpose of the second call to `Ros_Controller_GetNotReadySubcode`?

ted-miller (changes requested):

src/MotionControl.c
jimmy-mcelwain: Yes it is intentional. I can get rid of the fall through though and just duplicate the 2 lines if it's more clear that way.

src/ControllerStatusIO.h
jimmy-mcelwain: I'll fix that. The name in the .c file is what I intended to use

src/MotionControl.c
jimmy-mcelwain: The first call ignored "tractable problems" which is what I'm calling `MotionNotReady` codes that can be resolved within the `Ros_MotionControl_StartMotionMode()` function. This second call makes sure that it actually meets all of the conditions for motion. It is logically the same as a `Ros_Controller_IsMotionReady()` call, like the calls before it, but I just used `Ros_Controller_GetNotReadySubcode` instead just in case it failed so that the proper `motion_readiness_code` can be returned

ted-miller: Ok. that makes sense. Can you add some comment to that effect?

gavanderhoorn:
@jimmy-mcelwain: could you perhaps describe your approach at a high-level?

You write:

> Changes start_traj_mode and start_point_queue_mode services so that they properly report the reasons for failure in case motion cannot be started.

what did they not do correctly before these changes, and which changes (to control flow perhaps) address that specifically?

jimmy-mcelwain:
What I wrote there doesn't describe what I changed since the last PR that I closed. I rebased on main and that comment sums up the purpose of this PR as a whole, not just the differences from the last PR.

This approach changes `Ros_Controller_GetNotReadySubcode()` to work better with `start_traj_mode` and `start_point_queue_mode` such that much less code is duplicated. Within `Ros_Controller_GetNotReadySubcode()` I check for intractable problems that require the user to take some action, e.g. like switching to remote mode or calling `/reset_error`. If I am attempting to start some motion mode, I will ignore tractable errors from `Ros_Controller_GetNotReadySubcode()` that are potentially addressed by starting the motion, e.g. turning on servos, starting the job. 

I also significantly simplified the control flow of the program. Previously, there were several calls to `goto updatestatus` which are no longer needed. The jumps to `[updatestatus](https://github.com/Yaskawa-Global/motoros2/blob/4133298e68a7dcf76ed6f636c79d922ddcdfda2d/src/MotionControl.c#L1560-L1583)` always followed some intractable issue. These `goto` calls didn't make sense to me, since they are checking if motion is ready almost immediately after they already determined that motion was not ready. I should have gotten rid of them in the previous PR. To me it makes a lot more sense to simply return with a `MotionNotReadyCode` if it is determined that the motion cannot start and nothing can be done about it.

src/MotionControl.c
ted-miller: Ok. Just add some comment that states the message is being populated here because that this location has the most information about the error. Also, that other cases will populate the message upstream.

ted-miller (approved):
testing worked well. Conflicts resolved using web editor

Issue #181, opened 2023-10-16
Unclear message in `ResetError` response for major alarms


gavanderhoorn:
While working on #14, I ran into a situation where I had at least one major alarm active and as part of testing tried to use the `reset_error` service to reset the controller.

As major alarms can't be reset (that way), this didn't work -- as expected.

But the service result returned by MotoROS2 wasn't very informative:

```
motoros2_interfaces.srv.ResetError_Response(
  result_code=motoros2_interfaces.msg.MotionReadyEnum(value=101),
  message='Robot has an active ALARM'
)
```

while technically true -- there is still, at least, one active alarm -- we might perhaps try to provide a better message in case someone tries to reset major alarms.

The PP shows "can't reset major alarm" in the message area whenever you try to reset those. Not sure whether we can pick up on that, or whether we could perhaps check the *type* of alarm(s) that remain(s) and return a different `motoros2_interfaces.msg.MotionReadyEnum` value in cases where some alarms can't be reset.

PR #298, opened 2024-09-04
Understandable message for /reset_error if major alarm is active

jimmy-mcelwain:
In reference to #181. Now if a major alarm is active upon calling `/reset_error`, motoros2 will not make any effort to reset alarms, and it will give a better response message. I also added a `MotionNotReadyCode `for major alarms, separating it from minor alarms.

### DIFF
--- a/src/ControllerStatusIO.c
+++ b/src/ControllerStatusIO.c
@@ -315,6 +315,11 @@ BOOL Ros_Controller_IsAlarm()
         || (g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_USER]!=0) );
 }
 
+BOOL Ros_Controller_IsMajorAlarm()
+{
+    return ((g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_MAJOR] != 0));
+}
+
 BOOL Ros_Controller_IsError()
 {
     return ((g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ERROR]!=0));
--- a/src/ServiceResetError.c
+++ b/src/ServiceResetError.c
@@ -83,7 +83,15 @@ void Ros_ServiceResetError_Trigger(const void* request_msg, void* response_msg)
         }
     }
 
-    // Check for condition that can be fixed remotely
+    // Major alarms cannot be reset remotely
+    if (Ros_Controller_IsMajorAlarm())
+    {
+        rosidl_runtime_c__String__assign(&response->message, "Major alarm active. Cannot be reset. Check teach pendant");
+        response->result_code.value = MOTION_NOT_READY_MAJOR_ALARM;
+        goto DONE;
+    }
+
+    // Other alarm types can be reset remotely
     if (Ros_Controller_IsAlarm())
     {
         // Reset alarm
--- a/src/ControllerStatusIO.h
+++ b/src/ControllerStatusIO.h
@@ -111,6 +111,7 @@ extern void Ros_Controller_StatusInit();
 extern BOOL Ros_Controller_StatusRead(USHORT ioStatus[IO_ROBOTSTATUS_MAX]);
 extern BOOL Ros_Controller_IoStatusUpdate();
 extern BOOL Ros_Controller_IsAlarm();
+extern BOOL Ros_Controller_IsMajorAlarm();
 extern BOOL Ros_Controller_IsError();
 extern BOOL Ros_Controller_IsPlay();
 extern BOOL Ros_Controller_IsTeach();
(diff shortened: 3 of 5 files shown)


jimmy-mcelwain:
Here is the necessary [motoros2_interfaces pr](https://github.com/Yaskawa-Global/motoros2_interfaces/pull/23)

src/ServiceResetError.c
```diff
@@ -84,7 +84,15 @@ void Ros_ServiceResetError_Trigger(const void* request_msg, void* response_msg)
...
     // Check for condition that can be fixed remotely
-    if (Ros_Controller_IsAlarm())
+    if (g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_MAJOR] != 0)
+    {
+        rosidl_runtime_c__String__assign(&response->message, "Major alarm active. Cannot be reset. Check teach pendant");
+        response->result_code.value = MOTION_NOT_READY_MAJOR_ALARM;
+        goto DONE;
+    } 
+    else if ((g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_MINOR] != 0)
+        ||  (g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_SYSTEM] != 0)
```
ted-miller, line 94: @jimmy-mcelwain, were you able to confirm that `IO_ROBOTSTATUS_ALARM_SYSTEM` is resettable?

jimmy-mcelwain: Yes, system alarms are in the 9XXX range and can be reset. What we were talking about before are Off-line alarms, which don't have a dedicated I/O bit, but looking at documentation it looks like Off-line alarms are a subset of major alarms/the major alarm bit should be 1 if there is an offline alarm.

ted-miller:
@jimmy-mcelwain I just restructured a bit for consistency with the rest of the code.

ted-miller (approved):
@jimmy-mcelwain Let me know if you have any issue with my changes. Otherwise, I think it's good

src/ControllerStatusIO.c
```diff
@@ -315,6 +315,11 @@ BOOL Ros_Controller_IsAlarm()
         || (g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_USER]!=0) );
 }
 
+BOOL Ros_Controller_IsMajorAlarm()
+{
+    return (g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_MAJOR] != 0);
```
gavanderhoorn, line 320: I know it's not needed, but let's keep style consistent:

```suggestion
    return ((g_Ros_Controller.ioStatus[IO_ROBOTSTATUS_ALARM_MAJOR] != 0));
```

src/ServiceResetError.c
```diff
@@ -84,7 +84,13 @@ void Ros_ServiceResetError_Trigger(const void* request_msg, void* response_msg)
     }
 
     // Check for condition that can be fixed remotely
-    if (Ros_Controller_IsAlarm())
+    if (Ros_Controller_IsMajorAlarm())
+    {
+        rosidl_runtime_c__String__assign(&response->message, "Major alarm active. Cannot be reset. Check teach pendant");
```
gavanderhoorn, line 89: Ideally we'd use ` motoros2_interfaces__msg__MotionReadyEnum__NOT_READY_MAJOR_ALARM_STR` instead of a custom string. But seeing the rest of the function, perhaps that would be out-of-scope for this PR.

src/ServiceResetError.c
```diff
@@ -84,7 +84,13 @@ void Ros_ServiceResetError_Trigger(const void* request_msg, void* response_msg)
...
+        response->result_code.value = MOTION_NOT_READY_MAJOR_ALARM;
+        goto DONE;
+    } 
+    else if (Ros_Controller_IsAlarm())
```
gavanderhoorn, line 93: Because of the `goto DONE`, this is not needed.

I'd simplify by just making this an `if`.

gavanderhoorn (changes requested):

src/ServiceResetError.c
jimmy-mcelwain: Yeah I would prefer to use `motoros2_interfaces__msg__MotionReadyEnum__NOT_READY_MAJOR_ALARM_STR` but I decided to keep it consistent with the rest of the function. I can make another PR that changes the rest of the function to match in the future.

gavanderhoorn (approved):

gavanderhoorn:
as a final request: @jimmy-mcelwain: could you maybe squash your commits (and @ted-miller's fixup)? This change doesn't seem to need 4 commits :)

gavanderhoorn:
Thanks @jimmy-mcelwain :+1:

gavanderhoorn:
We'll need to fix CI after https://github.com/Yaskawa-Global/motoros2_interfaces/pull/23 is merged.

PR #277, opened 2024-07-10
Update changelog (0.1.3)

gavanderhoorn:
Preparation for the `0.1.3` release.

See discussion in Yaskawa-Global/motoros2_interfaces#20 on whether we should include #265 (and related Yaskawa-Global/motoros2_interfaces#17).

PR #299, opened 2024-09-04
Allow the user to insert USB with config file after bootup

ted-miller:
While testing #277, I inserted the USB after bootup. It didn't recognize my config file until I rebooted.

### DIFF
--- a/src/ConfigFile.c
+++ b/src/ConfigFile.c
@@ -767,14 +767,14 @@ void Ros_ConfigFile_Parse()
 
     Ros_ConfigFile_SetAllDefaultValues();
 
-#if defined (YRC1000) || defined (YRC1000u)
-    //config file always resides on USB for DX200/FS100, so only check
-    //on YRC1000 and micro
-    Ros_ConfigFile_CheckUsbForNewConfigFile();
-#endif
-
     do
     {
+#if defined (FS100) ||defined (YRC1000) || defined (YRC1000u)
+        //config file always resides on USB for DX200, so only check
+        //on YRC1000 and micro
+        Ros_ConfigFile_CheckUsbForNewConfigFile();
+#endif
+
         if (!bOkToInit)
             Ros_Sleep(3000);
 


src/ConfigFile.c
```diff
@@ -767,14 +767,14 @@ void Ros_ConfigFile_Parse()
...
 
-#if defined (YRC1000) || defined (YRC1000u)
-    //config file always resides on USB for DX200/FS100, so only check
-    //on YRC1000 and micro
-    Ros_ConfigFile_CheckUsbForNewConfigFile();
-#endif
-
     do
     {
+#if defined (FS100) ||defined (YRC1000) || defined (YRC1000u)
```
gavanderhoorn, line 772:
```suggestion
#if defined (YRC1000) || defined (YRC1000u)
```

gavanderhoorn (changes requested):
This is also a bit of a behaviour change. Don't we state in the documentation that config changes need a controller restart?

ted-miller:
> This is also a bit of a behaviour change. Don't we state in the documentation that config changes need a controller restart

Not really. We just tell people to insert the USB while the controller is powered down. That's primarily for safety reasons of opening the cabinet while powered up. 

We already continuously poll the SRAM for a file. I don't see any reason why we shouldn't poll USB too.

src/ConfigFile.c
ted-miller: I added FS100 because it's already called out a few lines down, when setting the `storageDrive` path.

gavanderhoorn (approved):

src/ConfigFile.c
gavanderhoorn, line 772:
```suggestion
#if defined (FS100) || defined (YRC1000) || defined (YRC1000u)
```

gavanderhoorn (changes requested):
Noticed it just now.

gavanderhoorn (approved):

PR #300, opened 2024-09-06
ci: disable mpBuilder problem matcher

gavanderhoorn:
As per title.

At least for now.

It needs more work and the many -- repeated -- annotations distract from actual issues.

Additionally: mpBuilder itself needs to be changed a bit.


ted-miller:
I don't understand what this does. 

Does this remove those awful warnings in the `Files changed` view?

gavanderhoorn:
> Does this remove those awful warnings in the `Files changed` view?

For now, yes.

They will return though, but in a more meaningful way.

ted-miller (approved):

Issue #52, opened 2023-06-06
Support Iron Irwini


gavanderhoorn:
Tasks:

 - [x] update VS project with Iron targets for all supported controllers (https://github.com/Yaskawa-Global/motoros2/compare/main...iron_wip, @ted-miller)
 - [ ] update M+ `libmicroros` build infra with Iron support
   - updates to `micro_ros_motoplus`: Yaskawa-Global/micro_ros_motoplus#3
   - updates to `_buildscripts`
 - [ ] find solution for broken CLI argument parsing (MotoROS2 does not have a CLI) due to upstream micro-ROS / `rcl` changes
    - work-around for now: https://github.com/Yaskawa-Global/motoros2/compare/main...gavanderhoorn:iron_no_remap_rules
 - [x] update documentation: (https://github.com/Yaskawa-Global/motoros2/pull/303)
   - [x] General requirements
   - [ ] Extracting the files
   - [x] Agent
   - [x] Only Fast-DDS is supported (still)

gavanderhoorn:
If I'm not mistaken, Iron support has just been released upstream: [micro-ROS Iron: 4.1.0](https://github.com/micro-ROS/micro_ros_setup/releases/tag/4.1.0).

gavanderhoorn:
@ted-miller: I believe you suggested adding this to the `0.2.0` milestone in #120?

gavanderhoorn:
Something to be aware of: https://github.com/ros2/rmw_dds_common/pull/68.

Almost starts to pay off to implement some sort of CI that checks message compatibility between different ROS 2 distributions.

ted-miller:
Doesn't build yet. But I'm working on this:
https://github.com/Yaskawa-Global/motoros2/tree/iron_wip

ted-miller:
> update VS project with Iron targets for all supported controllers

This kinda sucks to do and the current method does not scale well. I'm assuming the `cmake-passthrough` mode of M+SDK would make compilation easier. But I'm not sure how intellisense would behave without the stuff that was manually added.

ted-miller:
Very quick/preliminary testing. But I wanted to write it down before I leave.

AFAICT, libmicroros seems OK. (More testing needed)

Other things I noticed using the Iron version
- During setup, I had to `set S2C1402=3 and reboot` twice. This was probably a mistake on my part, but I really thought I set it right on the first attempt.
[...]

PR #303, opened 2024-09-06
Iron Irwini support

ted-miller:
- Add project infrastructure
- Remove remap rules
- Update documentation


gavanderhoorn:
Honestly, I'm hesitant to approve this as it basically breaks existing functionality.

I agree with your [comment ](https://github.com/Yaskawa-Global/motoros2/issues/52#issuecomment-2334647081) that we should support Iron already, but remaps do see use.

I'm not sure whether having Iron supported without remapping is a good idea.

But I won't block it if you really want to merge it in.

(I would recommend rebasing though, instead of merging in `main`)

ted-miller:
This passes all my testing.

ted-miller:
Since Iron Irwini is EOL in a month, we will skip this release and focus on Jazzy, which is LTS.

ted-miller:
NOTE: The work in `iron_wip` will be carried forward into the jazzy release

gavanderhoorn:
@jimmy-mcelwain @ted-miller: the `strtoll(..)` implementation -- if really needed -- should ideally go into the posix shim. Not in MotoROS2 itself.

Issue #128, opened 2023-08-16
Provide human-readable definitions/alarms for RCL(C) initialisation errors


ted-miller:
@ted-miller: we might want to consider reporting some of these RCL(C) initialisation errors, or defining some subcodes for them.

Especially the `NODE`, `PUBLISHER`, `SUBSCRIPTION`, etc categories seem to be caused by things which could often be solved by users without needing to post on the tracker here.

_Originally posted by @gavanderhoorn in https://github.com/Yaskawa-Global/motoros2/discussions/125#discussioncomment-6743301_

gavanderhoorn:
Looking into this a bit: this wouldn't be too much work to implement, but I'm trying to figure out a good place for these kinds of "RCL(C) initialisation errors/alarms" in `ErrorHandling.h`.

We don't really have a main alarm category for those I believe:

https://github.com/Yaskawa-Global/motoros2/blob/c9a9670d46298fceb4d4197e345c0515f66ca946/src/ErrorHandling.h#L71-L82

The issue title already refers to these kinds of problems as "initialisation errors", so it doesn't seem like they would fit under `ALARM_CONFIGURATION_FAIL` or `ALARM_TASK_CREATE_FAIL`. And we only use `ALARM_ASSERTION_FAIL` for actual assertions, which don't get much text on the PP.

do we still have 'space' for a category dedicated to RCLC_INIT errors/alarms?

We could then raise a regular alarm and provide some descriptive text as part of it.

gavanderhoorn:
As to subcodes: initially I thought we could perhaps opt to just 'forward' the various `RCL_RET_*` defined codes.

However, there is no guarantee those values will never change and that would mean we'd have to keep track of that and update our documentation if that happens.

Two approaches I've come up with:

1. define a single subcode, reuse one of the alarm categories, and include the RCL(C) error code as part of the message text

1. define (something like) an `ALARM_RCL_RCLC_FAIL` category, then (re)define relevant (perhaps all?) `RCL_RET_*` as subcodes (in a new `typedef enum`). RCL(C) API failures could then be posted to the PP:

   ```c
   mpSetAlarm(ALARM_RCL_RCLC_FAIL, "Invalid node name",
       SUBCODE_RCL_RCLC_FAIL_NODE_INVALID_NAME);
   ```

the former would be trivial to implement (although we need to figure out which category to use).

The latter would be a little more work, but would make RCL(C) errors more 'explicit' and we could document each of them individually.

[...]

Issue #44, opened 2023-06-01
VS intelli-sense has difficulty parsing headers


ted-miller:
There's a LOT of red squiggles indicating code errors. However, this builds just fine.

I suspect that either the order of folders in the intellisense-search-path is incorrect, or it's caused by the monolithic `MotoROS2.h` that includes everything in one place.

gavanderhoorn:
I'm not sure whether this just affects IntelliSense.

To appease a static analysis tool I'm trialling, I sorted all `#include`s alphabetically. That by itself caused the build to fail. Reverting the change made things work again.

We might want to switch to an IWYU style, which should help avoid this.

Might also help IntelliSense make sense (hah) of everything again (but I'm not sure about that).

gavanderhoorn:
Looking into this a bit: I believe this (ie: intellisense complaining about certain headers) is caused by the fact it's by default incapable of processing non-MSVC targeting headers.

In contrast to VS Code, VS is difficult to configure for cross-compilation / non-MSVC toolchains. That is, with `msbuild` / native projects.

Where in VS Code it's almost trivial to teach the C++ plugin how to talk to a non-MSVC toolchain, VS doesn't really seem to support it, unless you use the *Open Directory* and/or *CMake project* features.

Here is some output from the Intellisense debug log when it's parsing headers for `CommunicationExecutor.c`:

<details>
<summary>Click to expand</summary>

```
<8476> [IntelliSense] UpdateRange: (84) with pos range (0, 1000) - C:\Users\user\dev\motoros2\src\CommunicationExecutor.c
<8476> [IntelliSense] Translation Unit: C:\Users\user\dev\motoros2\src\CommunicationExecutor.c
<8476> [IntelliSense] Project Name: C:\USERS\USER\DEV\MOTOROS2\SRC\MOTOROS2_ALLCONTROLLERS.VCXPROJ
<8476> [IntelliSense] Project GUID: {64C31524-A3C4-49D4-AD04-955D44202226}
<8476> [IntelliSense] Priority: Normal
<8476> [IntelliSense] Configuration Name: YRC1000_humble|Win32
<8476> [IntelliSense] Toolset IntelliSense Identifier: < EMPTY >
<8476> [IntelliSense] command line options:
/c
/W1
/WX-
/DMOTOROS2_MEM_TRACE_ENABLE
/DCPU=_VX_ATOM
/DTOOL_FAMILY=gnu
/DTOOL=gnu
/DYRC1000
[...]

PR #306, opened 2024-09-10
Report rcl/rclc errors

jimmy-mcelwain:
Fix for #128. RCL(C) API errors get reported to the user now. It just notifies the user that an RCL(C) API error occurred and gives the integer value of the `rcl_ret_t` value associated with that error that the user has to look up themselves. 

`motoRos_RCLAssertOK_withMsg()` checks that `the rcl_ret_t` value that is passed in is OK, and if it is not, it not only sets an alarm indicating that, but it also sets an alarm indicating where it went wrong, like `motoRos_Assert_withMsg()` does. It's a bit ugly, something like what I described in the last paragraph of [this comment](https://github.com/Yaskawa-Global/motoros2/issues/128#issuecomment-2326789648) may be prettier/do a better job of separating functionality/not duplicating code, but that would increase complexity a bit. It would require either a global "problems" struct, or several local "problems" structs in different files that get passed around that the hypothetical `RaiseAlarms()` function could reference. 

This is a draft for now because I would like advice on what to do with `motoRos_RCLAssertOK_withMsg()`. I can squash the commits afterwards and rebase on main.


gavanderhoorn:
> It just notifies the user that an RCL(C) API error occurred and gives the integer value of the rcl_ret_t value associated with that error that the user has to look up themselves.

I would phrase it slightly more positive: users would lookup alarms raised by MotoROS2 in the troubleshooting documentation already, so this is just one more to look up.

It's just that in this case, it's as-if there's an additional sub code (a *sub-sub code*?), which happens to correspond to the `rcl_ret_t` value.

jimmy-mcelwain:
Yeah that's a good point

jimmy-mcelwain:
I was updating the troubleshooting guide to add information for the new error code, but the suggestion [here](https://github.com/Yaskawa-Global/motoros2/issues/128#issuecomment-2333759564) says to include information on user-serviceable errors. I am not sure which errors are user serviceable and which ones are not. For now I just have a link to [the types.h file referenced in our most recent build](https://github.com/Yaskawa-Global/micro_ros_rcl/blob/humble-motoplus1-candidate-20230624/rcl/include/rcl/types.h) but I know that isn't sufficient. 

Also, while I agree that the previous approach with the "shadow codes" would require maintenance in case there are changes to `types.h` in the future, won't we need to make changes in the troubleshooting guide if `types.h` changes anyways? 

Also, unrelated, but I was having trouble figuring out why the alarm linter was upset. Then I figured out that it didn't like me using `rcl_ret_t` as a parameter type in the function prototypes without it knowing about that typedef, so despite the fact that it was compiling and working correctly, the parser for the alarm linter was failing. I just changed it to `int` (`rcl_ret_t`'s real type), but I don't know if that is the preferred way of resolving that problem. It makes it a little less clear what you're supposed to pass in to the function.

gavanderhoorn:
> Also, while I agree that the previous approach with the "shadow codes" would require maintenance in case there are changes to types.h in the future, won't we need to make changes in the troubleshooting guide if types.h changes anyways?

it's true we'd have to keep an eye on it in any case. The hope is we've reduced the amount of maintenance to a minimum by not also introducing a new enum on our side (whose sole purpose is to shadow an already existing enum in the RCL(C) code base).

> Also, unrelated, but I was having trouble figuring out why the alarm linter was upset.

it looks like it's unhappy because the file it was parsing ([ErrorHandling.h](https://github.com/Yaskawa-Global/motoros2/blob/c61fb86e2cf31ff894189efd8a36a2f1fc5d1161/src/ErrorHandling.h)) referenced an undefined type before your edit in 59f95820 (this [version](https://github.com/Yaskawa-Global/motoros2/blob/e6cee2fe9e503628cf600e45041649bb5b7e6260/src/ErrorHandling.h#L234-L235)): `rcl_ret_t` is not part of C proper, but a custom / derived type. But `ErrorHandling.h` does not `#include` anything, so that type is not defined. 

> despite the fact that it was compiling and working correctly

that's because all `.c` files include the top-level `MotoROS.h`, which *does* include the required other `.h`s which then define the `rcl_ret_t` type.

I've wanted/tried to switch to an IWYU style before, but the MotoPlus SDK VS integration isn't happy if we do that (see #44).

ted-miller:
> It's a bit ugly, something like what I described in the last paragraph of https://github.com/Yaskawa-Global/motoros2/issues/128#issuecomment-2326789648 may be prettier/do a better job of separating functionality/not duplicating code, but that would increase complexity a bit

I think this is ok for the sake of simplicity.

ted-miller (approved):

PR #309, opened 2024-09-12
Add debug broadcast enable/disable option and debug broadcast port option to config file

jimmy-mcelwain:
I added 2 extra settings to the `motoros2_config.yaml` file, `userlan_debug_broadcast_enabled` and `userlan_debug_broadcast_port`. By default, it broadcasts to all enabled ports, but the user can choose a specific port if they only want it to broadcast to one port, or the user can disable it.


jimmy-mcelwain:
I have tested pretty thoroughly on a YRC1000 and I did not run into any issues.

gavanderhoorn:
CI failures of the `msbuild` tasks is expected. Any other failures would have to be addressed.

src/ConfigFile.c
```diff
@@ -327,7 +333,7 @@ void Ros_ConfigFile_CheckYamlEvent(yaml_event_t* event)
                 case Value_UserLanPort:
 #if defined (FS100) || defined (DX200)
                     // single port, override whatever was configured
-                    *(Ros_UserLan_Port_Setting*)activeItem->valueToSet = CFG_ROS_USER_LAN1;
+                    * (Ros_UserLan_Port_Setting*)activeItem->valueToSet = CFG_ROS_USER_LAN1;
                     Ros_Debug_BroadcastMsg("DX200 or FS100: override to 'USER_LAN1'");
 
 #elif defined (YRC1000) || defined (YRC1000u)
```
ted-miller, line 339: This is not accurate. YRCu only has one LAN port.

config/motoros2_config.yaml
```diff
@@ -284,3 +284,35 @@ publisher_qos:
...
+# Which network port should MotoROS2 broadcast the debug messages to, if 
+# 'userlan_debug_broadcast_enabled' is 'true'?
+#
+# If not specified and 'userlan_debug_broadcast_enabled' is true, MotoROS2 will 
+# send the broadcast to all network ports which are enabled on the controller. 
+#
+# To choose a specific port to broadcast debug messages, uncomment 
+# 'userlan_debug_broadcast_port' below and set it to the desired port.
+#
+# NOTE 1:   this setting only applies to YRC1000 and YRC1000u controllers.
```
ted-miller, line 312: YRC1000u only has one LAN port

doc/troubleshooting.md
```diff
@@ -1024,6 +1053,28 @@ In case the alarm is still raised after calibration was performed, TF broadcasti
...
+
+```text
+ALARM 8013
+ Bad UserLan debug port in cfg
+[17]
+```
+
+*Solution:*
+The `userlan_debug_broadcast_port` key in the `motoros2_config.yaml` configuration file is set to an invalid value.
+Debug broadcasting will be disabled for this session.
```
ted-miller, line 1068: > Debug broadcasting will be disabled for this session

I don't like this behavior. I definitely think broadcasting should be enabled if there's an error in the configuration.

src/Debug.c
```diff
@@ -14,8 +14,14 @@
...
-int ros_DebugSocket = -1;
-struct sockaddr_in ros_debug_destAddr1;
+typedef struct
+{
+    UINT8 enabledPortCount;
+    int debugSocket[MAX_NETWORK_PORTS];
+    struct sockaddr_in destAddr[MAX_NETWORK_PORTS];
+} userLanInfo;
+
+userLanInfo debugPorts = {0};
```
ted-miller, line 24:
```suggestion
userLanInfo ros_debugPorts = {0};
```

This all exists in global kernel space. Easy to conflict with other global names.

src/Debug.c
```diff
@@ -24,31 +30,91 @@ void Ros_Debug_Init()
...
+    STATUS status;
+    int count = 0;
+    int* socket = debugPorts.debugSocket;
+    struct sockaddr_in* sin = debugPorts.destAddr;
+
+    bzero(&debugPorts, sizeof(debugPorts));
+
+    for (int i = 1; i <= MAX_NETWORK_PORTS; i++)
+    {
+        socket[count] = mpSocket(AF_INET, SOCK_DGRAM, 0);
```
ted-miller, line 42: This should be inside of the `if (status == OK)` block

src/Debug.c
```diff
@@ -24,31 +30,91 @@ void Ros_Debug_Init()
...
+            sin[count].sin_addr.s_addr = ip_be | (~subnetmask_be);
+            sin[count].sin_family = AF_INET;
+            sin[count].sin_port = mpHtons(DEBUG_UDP_PORT_NUMBER);
+            count++;
+        }
+    }
+    debugPorts.enabledPortCount = count;
+    if (count < 1)
+    {
+        mpSetAlarm(ALARM_ASSERTION_FAIL, "Must enable ETHERNET function", SUBCODE_DEBUG_INIT_FAIL_MP_NICDATA_ALL);
```
ted-miller, line 58: This is already in `Ros_ConfigFile_SetAllDefaultValues`

ted-miller (changes requested):

doc/troubleshooting.md
jimmy-mcelwain: I can change that. So `userlan_monitoring_enabled` should be set to disabled if there is an error in the config, but broadcasting should still be enabled? I only ask because they're in the same case for the config switch case.

config/motoros2_config.yaml
jimmy-mcelwain: There is a lot of code outside of this PR assuming that the YRCu has 2 LAN ports. I'll fix it for this code, but there should probably be a second PR that fixes that up elsewhere.

doc/troubleshooting.md
ted-miller: I don't have an opinion on the behavior of `userlan_monitoring_enabled` in this case

config/motoros2_config.yaml
gavanderhoorn: Tracking in #310.

doc/troubleshooting.md
gavanderhoorn: UserLan monitoring is extra functionality: it makes detecting disconnections faster, but isn't absolutely necessary (other places will also detect disconnects, they'll just take longer).

There's no harm in disabling it if the configuration is incorrect.

But debug broadcasts should indeed not be disabled in cases of incorrect configuration. Especially since we often use debug logs to diagnose misconfiguration.

src/ConfigFile.c
gavanderhoorn: We'll fix the `YRC1000u` case when addressing #310.

src/ConfigFile.c
```diff
@@ -690,7 +696,7 @@ void Ros_ConfigFile_ValidateNonCriticalSettings()
             if (g_nodeConfigSettings.userlan_monitor_port != CFG_ROS_USER_LAN1)
 #endif
             {
-                mpSetAlarm(ALARM_CONFIGURATION_FAIL, "Invalid UserLan port in cfg",
+                mpSetAlarm(ALARM_CONFIGURATION_FAIL, "Bad UserLan monitor port in cfg",
```
gavanderhoorn, line 699: Did you have a particular reason changing the language here?

src/ConfigFile.c
```diff
@@ -862,6 +890,12 @@ void Ros_ConfigFile_Parse()
 
     Ros_ConfigFile_ValidateCriticalSettings();
     Ros_ConfigFile_ValidateNonCriticalSettings();
+#if defined(YRC1000) || defined(YRC1000u)
+    if(g_nodeConfigSettings.userlan_debug_broadcast_enabled &&
+        (g_nodeConfigSettings.userlan_debug_broadcast_port == CFG_ROS_USER_LAN1 ||
+        g_nodeConfigSettings.userlan_debug_broadcast_port == CFG_ROS_USER_LAN2))
+        Ros_Debug_SetFromConfig();
```
gavanderhoorn, line 897:
```suggestion
    if(g_nodeConfigSettings.userlan_debug_broadcast_enabled &&
        (g_nodeConfigSettings.userlan_debug_broadcast_port == CFG_ROS_USER_LAN1 ||
        g_nodeConfigSettings.userlan_debug_broadcast_port == CFG_ROS_USER_LAN2))
    {
        Ros_Debug_SetFromConfig();
    }
```

please always enclose bodies of conditionals in brackets. Just to make my internal parser happy perhaps, but clearer in any case.

src/ConfigFile.h
```diff
@@ -88,17 +88,20 @@ typedef enum
 
 typedef enum
 {
-    CFG_ROS_USER_LAN_DISABLED = -2,  //sentinel
+    CFG_ROS_USER_LAN_MALFORMED = -3,  //sentinel
+    CFG_ROS_USER_LAN_ALL = -2,  //sentinel
     CFG_ROS_USER_LAN_AUTO = -1,  //sentinel
     CFG_ROS_USER_LAN1 = ROS_USER_LAN1,
     CFG_ROS_USER_LAN2 = ROS_USER_LAN2,
 } Ros_UserLan_Port_Setting;
```
gavanderhoorn, line 96: I understand why you opted to reuse the `Ros_UserLan_Port_Setting` type, but it does make the parsing code a little harder to read.

src/ConfigFile.c
jimmy-mcelwain: There are now 2 different UserLan ports that the user can specify (monitor and debug broadcast), so I added the word "monitor" to clarify which one this condition checked. But then that put it above 32 characters (the alarm message limit), so I chose to shorten "Invalid" to "Bad" to fit in the character limit. I think that "invalid" is a better word for that, but it's more important to be specific about which config setting needs to be changed.

src/ConfigFile.h
jimmy-mcelwain: Yeah I tried doing it both ways, and this had much less duplicated code and was overall cleaner in my opinion. I don't especially like it either though.

src/ConfigFile.c
gavanderhoorn: So should we instead separate parsing and error reporting for monitoring and debug log port?

Would lead to some mild code duplication, but would also make parsing clearer and avoids this situation.

config/motoros2_config.yaml
```diff
@@ -284,3 +284,35 @@ publisher_qos:
 #
 # DEFAULT: false
 #ignore_missing_calib_data: false
+
+#-----------------------------------------------------------------------------
+# Should MotoROS2 broadcast debug messages?
+#
+# If enabled, this will broadcast log messages on the network on port UDP 21789.
+# the user can use the debug script provided in the distribution of MotoROS2 to
```
gavanderhoorn, line 292:
```suggestion
# The user can use the debug script provided in the distribution of MotoROS2 to
```

unfortunately, we actually *don't* include the logger script in release `.zip`s.

config/motoros2_config.yaml
```diff
@@ -284,3 +284,35 @@ publisher_qos:
...
+# monitor the state of the robot, identify problems, and debug their code.
+#
+# The debug script is available under the Yaskawa-Global/motoros2 repository in
+# the tools directory
+# https://github.com/Yaskawa-Global/motoros2/tree/main/tools
+#
+# DEFAULT: true
+#userlan_debug_broadcast_enabled: true
```
gavanderhoorn, line 300: I might have a slight preference for removing the `userlan_` prefix here.

From a user perspective, there are no other LAN ports. The distinction between the port used for the PP and the 'other one(s)' is only made by the OS. Even from M+ we can't really use any other ports.

It would also avoid the (implied) relation between the LAN monitoring and the debug broadcast -- which doesn't exist.

config/motoros2_config.yaml
```diff
@@ -284,3 +284,35 @@ publisher_qos:
...
+
+#-----------------------------------------------------------------------------
+# Which network port should MotoROS2 broadcast the debug messages to, if
+# 'userlan_debug_broadcast_enabled' is 'true'?
+#
+# If not specified and 'userlan_debug_broadcast_enabled' is true, MotoROS2 will
+# send the broadcast to all network ports which are enabled on the controller.
```
gavanderhoorn, line 307:
```suggestion
# send messages over all network ports which are active on the controller.
```

config/motoros2_config.yaml
```diff
@@ -284,3 +284,35 @@ publisher_qos:
...
+# To choose a specific port to broadcast debug messages, uncomment
+# 'userlan_debug_broadcast_port' below and set it to the desired port.
+#
+# NOTE 1:   this setting only applies to YRC1000 and YRC1000u controllers.
+#         DX200 and FS100 controllers only have a single ethernet port, and will
+#         always default to USER_LAN1
+#
+# OPTIONS: USER_LAN1, USER_LAN2
+# DEFAULT: (all available network ports)
+#userlan_debug_broadcast_port: USER_LAN1
```
gavanderhoorn, line 318: what would be the value (ie: string) for 'all available network ports'? Because the documentation states that's the default, but I can't set this to `'all available network ports'` I believe.

config/motoros2_config.yaml
jimmy-mcelwain: Should it be removed from the `userlan_monitoring_` config options as well, or just the `userlan_debug_broadcast_` options?

src/Debug.c
```diff
@@ -82,7 +141,9 @@ void Ros_Debug_BroadcastMsg(char* fmt, ...)
         memcpy(str, timestamp, timestamp_length);         
     }
 
-    mpSendTo(ros_DebugSocket, str, strlen(str), 0, (struct sockaddr*) &ros_debug_destAddr1, sizeof(struct sockaddr_in));
+    for (int i = 0; i < ros_debugPorts.enabledPortCount; i++)
+        mpSendTo(ros_debugPorts.debugSocket[i], str, strlen(str), 0, (struct sockaddr*)&(ros_debugPorts.destAddr[i]), sizeof(struct sockaddr_in));
```
gavanderhoorn, line 145:
```suggestion
    for (int i = 0; i < ros_debugPorts.enabledPortCount; i++)
    {
        mpSendTo(ros_debugPorts.debugSocket[i], str, strlen(str), 0, (struct sockaddr*)&(ros_debugPorts.destAddr[i]), sizeof(struct sockaddr_in));
    }
```

src/ErrorHandling.h
```diff
@@ -170,6 +170,7 @@ typedef enum
     SUBCODE_CONFIGURATION_FAIL_MP_NICDATA1,
     SUBCODE_FAIL_MP_NICDATA_INIT1,
     SUBCODE_FAIL_INVALID_BASE_TRACK_MOTION_TYPE,
+    SUBCODE_DEBUG_INIT_FAIL_MP_NICDATA_SPECIFIC,
```
gavanderhoorn, line 173: what's *specific* about it?

config/motoros2_config.yaml
gavanderhoorn: I'd consider everything not related to the new functionality you're introducing as out-of-scope for this PR.

config/motoros2_config.yaml
jimmy-mcelwain: So there is not a dedicated string for 'all available ports'. I was planning on having that, but part of the weirdness with it having the same type as `userlan_monitor_port` (`Ros_UserLan_Port_Setting`) is that even if they have different options with different behavior (e.g. `userlan_monitor_port` has 'auto detection' and `userlan_debug_broadcast_port` has 'all available ports'), they are parsed the same. So I had the behavior match that of `userlan_monitor_port`, which doesn't have a specific string for 'auto detection'. That being said, unlike `userlan_monitor_port`, if the user types in an invalid string, it will default to 'all available ports' rather than 'disabled'

src/ErrorHandling.h
jimmy-mcelwain: A previous commit had `SUBCODE_DEBUG_INIT_FAIL_MP_NICDATA_ALL` as a code, but now that has been removed so I could get rid of the `_SPECIFIC`

src/ConfigFile.c
jimmy-mcelwain: I don't believe that this specific message is affected by combined parsing/error reporting for the two port config options. My preference would be to keep the parsing as combined, but I would be fine with changing it.

doc/troubleshooting.md
jimmy-mcelwain: I changed the behavior and this documentation so that it broadcasts to all ports if the specific port isn't active.

gavanderhoorn:
I'll test this on my YRC1 and FS1 next week.

gavanderhoorn:
I've not tested this extensively yet (I will try to do that now), but as a high-level initial comment I might suggest adding some more comments.

From a quick read it's not immediately clear why only `socket[0]` is initialised in `Ros_Debug_SetFromConfig()` fi. The interaction between `Ros_Debug_SetFromConfig()` and `Ros_Debug_Init()` could also use some comments.

src/Debug.c
```diff
@@ -24,31 +30,92 @@ void Ros_Debug_Init()
...
 {
     char str[MAX_DEBUG_MESSAGE_SIZE];
     va_list va;
 
+    if (g_nodeConfigSettings.debug_broadcast_enabled && ros_debugPorts.enabledPortCount == 0)
+    {
+        Ros_Debug_Init();
+    }
+
+    if (!g_nodeConfigSettings.debug_broadcast_enabled && !g_nodeConfigSettings.log_to_stdout)
```
gavanderhoorn, line 108: this might also be a nice place to explain why this will allow control flow to continue here if `g_nodeConfigSettings.log_to_stdout == TRUE`?

(because we can reuse all the machinery after this to format the message and as `ros_debugPorts.enabledPortCount` is most likely `0`, nothing will be sent out over any of the LAN ports, but it will be printed to the controller's console. IIUC).

gavanderhoorn: @jimmy-mcelwain: I'm wondering whether this works as intended (specifically this line).

Could it be this prevents logs from getting broadcast until at least the config file has been parsed? Both `debug_broadcast_enabled` and `log_to_stdout` will be `FALSE` in that case I believe.

Looking at the debug log output of my YRC1, it seems to be missing at least these printouts:

https://github.com/Yaskawa-Global/motoros2/blob/c61fb86e2cf31ff894189efd8a36a2f1fc5d1161/src/main.c#L60-L64

jimmy-mcelwain: I think that you're right. Maybe if we call `Ros_ConfigFile_SetAllDefaultValues()` to set `debug_broadcast_enabled` to `TRUE` in `main` before any calls to `Ros_Debug_BroadcastMsg()`, then we would be able to broadcast those messages. I would need to think it through more though, that might require more changes.

jimmy-mcelwain:
I am now calling `Ros_ConfigFile_SetAllDefaultValues()` in `main` prior to any calls to `Ros_Debug_BroadcastMsg()`. I have only tested a little bit, but broadcast messages now print successfully, even if the call is before the config file is parsed. I will test more tomorrow.

jimmy-mcelwain:
It now prints before the config file is parsed and I added more comments

ted-miller:
I'm having a hard time following the logic. But I don't think the selective-enabling is working as it's supposed to.

``` yaml
#-----------------------------------------------------------------------------
# Should MotoROS2 broadcast debug messages?
#
# If enabled, this will broadcast log messages on the network on port UDP 21789.
# The user can use the debug script to monitor the state of the robot, identify
# problems, and debug their code.
#
# The debug script is available under the Yaskawa-Global/motoros2 repository in
# the tools directory
# https://github.com/Yaskawa-Global/motoros2/tree/main/tools
#
# DEFAULT: true
debug_broadcast_enabled: true

#-----------------------------------------------------------------------------
# Which network port(s) should MotoROS2 broadcast debug messages on, if
# 'debug_broadcast_enabled' is 'true'?
#
# If not specified and 'debug_broadcast_enabled' is true, MotoROS2 will
# send messages over all network ports which are active on the controller.
#
# To choose a specific port to broadcast debug messages, uncomment
# 'debug_broadcast_port' below and set it to the desired port.
#
# NOTE 1:   this setting only applies to YRC1000 and YRC1000u controllers.
#         DX200 and FS100 controllers only have a single ethernet port, and will
#         always default to USER_LAN1
#
# OPTIONS: USER_LAN1, USER_LAN2
# DEFAULT: (all available network ports)
debug_broadcast_port: USER_LAN2
```

In my case, only USER_LAN1 (LAN 2 on CPU board) is enabled. So, I didn't expect to see the messages in my terminal. 

But, I am seeing messages. And there's a line, `Config: debug_broadcast_port = -2`, which is the default value. So I'm not sure that any assignment is ever happening. 

In fact, I'm pretty sure it's not happening, because I never see this line: https://github.com/Yaskawa-Global/motoros2/blob/f035d498a722c00e306da4ca1bd02b73d3324988/src/ConfigFile.c#L362

I wonder if we need quotes around the value in the yaml?

ted-miller:
Quotes fixed it

ted-miller (approved):

jimmy-mcelwain:
@ted-miller I haven't been using quotes and it's been reading the correct `debug_broadcast_port` without any problems. That being said it also works for me with quotes.

ted-miller:
Nevermind. I must have accidentally loaded a config file w/o the last two configuration entries. Or perhaps I put it in the pendant but forgot to load.

It works fine w/o quotes.

ted-miller (approved):

PR #314, opened 2024-10-02
Fix condition so tool number 63 can be selected properly

jimmy-mcelwain:
Right now tool 63 cannot be selected because this condition evaluates as true and exits the function. With this change, tool 63 should be selectable.


gavanderhoorn (approved):
Thanks.

Initially `MAX_VALID_TOOL_INDEX` was actually something like `MAX_NUMBER_OF_TOOLS`. That's why the condition was `>=`.

Apparently we forgot to update the conditional when it changed to maximum index (zero-based).

gavanderhoorn:
I'd like to see green CI before merging though.

If you could review #315, we could rebase this one and then merge after/with green CI.

gavanderhoorn:
Thanks @jimmy-mcelwain :+1:

Issue #35, opened 2023-06-01
Possible memory leak (after dis-, then reconnect of Agent)


gavanderhoorn:
As mentioned in the *Known issues & limitations* section of the README ([here](https://github.com/Yaskawa-Global/motoros2/tree/08329e5a1646315bffb5effdb40857eaea343523#memory-leak)), MotoROS2 currently appears to suffer from a small memory leak.

Testing shows the following values for used and free memory after several consecutive connect and disconnect cycles with a Galactic micro-ROS Agent:

```
Initialization complete. Memory available: (962992) bytes. Memory in use: (85584) bytes
Initialization complete. Memory available: (962992) bytes. Memory in use: (85584) bytes
Initialization complete. Memory available: (962552) bytes. Memory in use: (86024) bytes
Initialization complete. Memory available: (962280) bytes. Memory in use: (86296) bytes
Initialization complete. Memory available: (962280) bytes. Memory in use: (86296) bytes
Initialization complete. Memory available: (962280) bytes. Memory in use: (86296) bytes
Initialization complete. Memory available: (962256) bytes. Memory in use: (86320) bytes
Initialization complete. Memory available: (962224) bytes. Memory in use: (86352) bytes
```

As a table:

|**#**|**Free**|**Delta**|**In use**|**Delta**|
|----:|-----:|----:|-----:|---:|
|    0|962992|     | 85584|    |
|    1|962992|    0| 85584|   0|
|    2|962552| -440| 86024| 440|
|    3|962280| -272| 86296| 272|
|    4|962280|    0| 86296|   0|
|    5|962280|    0| 86296|   0|
|    6|962256|  -24| 86320|  24|
|    7|962224|  -32| 86352|  32|

No changes to the config file between ~~reboots~~ dis-and-re-connects.

It's unclear what causes this at this point.

jimmy-mcelwain:
Just as an update to this, I am running iron, which includes the previously mentioned PR (https://github.com/micro-ROS/rmw_microxrcedds/pull/280). But it seems to still have the same memory issues. 

Here's a table showing the memory free and delta on my YRC1000 with Iron

|**#**|**Free**|**Delta**|
|----:|-----:|----:|
|    0|797368|     |
|    1|796848| -520|
|    2|796800|  -48|
|    3|796712|  -88|
|    4|796712|    0|
|    5|796752|  +40|
|    6|796752|    0|
|    7|796800|  +48|
|    8|796568| -232|
[...]

PR #325, opened 2024-11-13
Partial memory leak fix

jimmy-mcelwain:
A partial fix for #35. 

There was a member of the QueueTrajPoint response that was getting initialized twice.
```
g_messages_QueueTrajPoint.response = motoros2_interfaces__srv__QueueTrajPoint_Response__create();
rosidl_runtime_c__String__init(&g_messages_QueueTrajPoint.response->message);
```
The response creation call already initialized `g_messages_QueueTrajPoint.response->message`, so the init call that followed caused a memory leak. 

This will not close out the issue, as there is still a trickier memory leak as mentioned [in this comment](https://github.com/Yaskawa-Global/motoros2/issues/35#issuecomment-2474370293). But from my testing, this does seem to recoup somewhere in the range of 24-48 bytes each connect/disconnect cycle.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -144,8 +144,7 @@ void RosInitTask()
         if (tid == ERROR)
             mpSetAlarm(ALARM_TASK_CREATE_FAIL, APPLICATION_NAME " FAILED TO CREATE TASK", SUBCODE_EXECUTOR);
 
-        Ros_Debug_BroadcastMsg("Initialization complete. Memory available: (%d) bytes. Memory in use: (%d) bytes",
-                       mpNumBytesFree(), MP_MEM_PART_SIZE - mpNumBytesFree());
+        Ros_Debug_BroadcastMsg("Initialization complete.");
 
         //==================================
         ULONG tickBefore = 0;
--- a/src/main.c
+++ b/src/main.c
@@ -105,6 +105,7 @@ void RosInitTask()
     //==================================
     FOREVER
     {
+        MOTOROS2_MEM_TRACE_START(full_connection_cycle);
         Ros_Controller_StatusInit();
 
         Ros_Allocation_Initialize(&g_motoros2_Allocator);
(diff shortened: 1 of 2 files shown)


ted-miller:
Are you sure that one line is the only 'relevant' fix? Your other branch had a lot more changes.

I just did 7 disconnects in a row and got these results. It seems to conflict with what you [stated during your tests](https://github.com/Yaskawa-Global/motoros2/issues/35#issuecomment-2474370293).

```
Initialization complete. Memory available: (992672) bytes. Memory in use: (55904) bytes
Initialization complete. Memory available: (992632) bytes. Memory in use: (55944) bytes
Initialization complete. Memory available: (992616) bytes. Memory in use: (55960) bytes
Initialization complete. Memory available: (992592) bytes. Memory in use: (55984) bytes
Initialization complete. Memory available: (992560) bytes. Memory in use: (56016) bytes
Initialization complete. Memory available: (992512) bytes. Memory in use: (56064) bytes
Initialization complete. Memory available: (992536) bytes. Memory in use: (56040) bytes
```

jimmy-mcelwain:
Ignore the `memory available`/`memory in use` printout. That is printed out in the middle of a connect/disconnect cycle. I don't know exactly why it varies (it does on my system too), but I'd guess that it has to do with the still-present memory leak (could be messing with struct byte alignment or something like that). The relevant statistic is the `full_connection_cycle` delta, which is printed out at the actual end of the connect/disconnect cycle. The other changes that I took out were almost entirely logging statements to nail down the second memory leak.

src/main.c
```diff
@@ -143,8 +144,7 @@ void RosInitTask()
         if (tid == ERROR)
             mpSetAlarm(ALARM_TASK_CREATE_FAIL, APPLICATION_NAME " FAILED TO CREATE TASK", SUBCODE_EXECUTOR);
 
-        Ros_Debug_BroadcastMsg("Initialization complete. Memory available: (%d) bytes. Memory in use: (%d) bytes",
-                       mpNumBytesFree(), MP_MEM_PART_SIZE - mpNumBytesFree());
+        Ros_Debug_BroadcastMsg("Initialization complete.");
```
gavanderhoorn, line 147: could we not move this to a place where initialisation was really complete?

Issue #199, opened 2023-12-19
Use `mpGetFBPulsePosEx` instead of `mpGetFBPulsePos`


ted-miller:
`mpGetFBPulsePos` returns raw feedback position of the encoders. This can be different than logical position of the joint. This is due to things such as [cross-axis couplings](https://github.com/Yaskawa-Global/motoros2/blob/9026a1d0b50b483f7c334bfd5ebbac8b328bfdb6/src/CtrlGroup.c#L345-L358) or gravity-compensation.

We already implement compensation for the cross-axis thing. And I'm told that gravity-compensation is a very specialized feature that's only available on one or two arms.

Now I'm hearing that there is a new feature on the next-gen controller that could also be affected by this.

The simple fix is to use mpGetFBPulsePos**Ex**, which should apply any required corrections to the output.

gavanderhoorn:
So would that allow us to remove the code we currently have for things like that cross-axis coupling you link to?

ted-miller:
> So would that allow us to remove the code we currently have for things like that cross-axis coupling you link to?

Yes, we could remove that chunk of code.

> Is there also an mpGetServoSpeedEx(..)?

Not that I'm aware of.

-------------------------

I'm seeing some contradictory information on `mpGetFBPulsePosEx` regarding the `absolute accuracy` feature of the next-gen controller. I don't yet have the ability to test myself.

**One report**:

_It looks that the axes correction is applied just after calling mpGetFBPulsePos(), but if other correcting features are enabled, the differences will be appered. For example, gravitiy compensation (this is for special customer) or absolute accuracy (this is only YNX1000 currently). 
I suggest to use mpGetFBPulsePosEx() instead of mpGetFBPulsePos(),_ 

**Another**:
(apologies for the screenshot)

![image](https://github.com/Yaskawa-Global/motoros2/assets/4623811/960069b2-749f-42ee-9d66-54059e2b8979)

gavanderhoorn:
Hm. So in any case this reads as-if we'd have to introduce another batch of controller-specific `#ifdef`s. Correct?

ted-miller:
Yeah, sounds that way

ted-miller:
At this time, we don't intend to build MotoROS2 for the YNX1000.

gavanderhoorn:
Do the other benefits not make this interesting in any case, also for other controllers?

This fi:

> > So would that allow us to remove the code we currently have for things like that cross-axis coupling you link to?
> 
[...]

PR #338, opened 2024-11-26
Add note about `mpGetFBPulsePosEx`

ted-miller:
Clarify about why we may want to consider it for future usage.

In response to https://github.com/Yaskawa-Global/motoros2/issues/199#issuecomment-2498879600


gavanderhoorn (approved):

gavanderhoorn:
Thanks @ted-miller :+1:

Issue #74, opened 2023-06-21
Make example MoveIt 2 configuration package(s) available


gavanderhoorn:
MoveIt 2 is non-trivial to configure.

Especially when *not* using `ros2_control`, as almost all available documentation and examples assume `ros2_control` is being used.

We give some hint(s) in the `README` ([Usage - With MoveIt](https://github.com/Yaskawa-Global/motoros2/tree/2d2b62df3655078b93cfe77c1c133ff5df8670fd#with-moveit)), but that's really rather minimal, and assumes an already working MoveIt 2 configuration package has been created, and just needs to be modified for use with MotoROS2.

Providing an example, known good configuration package would help users set up something for their own / other robots.

ted-miller:
@aaron-levan, did you keep any of your previous work on this topic?

Aaron-Levan:
No, I had everything in a VM that is, unfortunately, no longer with us...

Aaron-Levan:
> @Aaron-Levan, did you keep any of your previous work on this topic?

No, I had everything in a VM that is, unfortunately, no longer with us...

gavanderhoorn:
I keep putting this on my TODO, but it's too long already ..

I have working package for my GP25 + TSL1000, but it's for / from a very old version of MoveIt and not really suited for distribution.

charleswilmot:
At my company, we are currently deciding whether we should order Yaskawa MOTOMAN HC30PL arms or Fanuc CRX20IA arms.

Being able to use MoveIt is a necessary condition for us to purchase the Yaskawa robots.

Is there someone that we could contact before we take a decision? We would need to have one robot ready to use with MoveIt before October 2023.

gavanderhoorn:
@charleswilmot: could I ask you to please repost your question on the Discussion forum [in the Q&A section](https://github.com/Yaskawa-Global/motoros2/discussions/categories/q-a)? Your comment is related to this issue, but we're using this to track the work we need to do, and not for general MoveIt support related enquiries.

drolleigh:
Hello,
I have a simple moveit 2 config package for a Motoman gp25-12. I have motoROS2 successfully installed on the YRC1000 controller, and it is publishing the joint states that match the prepended group name. And it is broadcasting TFs with R1 as the namespace. The TFs appear to be accurate to the physical state of the robot. 

How do I make the moveit 2 launch files talk to motoROS2 on the robot controller? 

Any help is appreciated,
-D

[...]

PR #340, opened 2024-12-06
Change default joint_state names for single-group

jimmy-mcelwain:
As discussed in https://github.com/Yaskawa-Global/motoros2/discussions/331. Omit the "group_N" part of the message if it is a single group.

### DIFF
--- a/README.md
+++ b/README.md
@@ -641,12 +641,12 @@ follow_joint_trajectory:
   type: FollowJointTrajectory
   default: true
   joints:
-   - group_1/joint_1
-   - group_1/joint_2
-   - group_1/joint_3
-   - group_1/joint_4
-   - group_1/joint_5
-   - group_1/joint_6
+   - joint_1
+   - joint_2
+   - joint_3
+   - joint_4
+   - joint_5
+   - joint_6
 
 ...
 ```
--- a/config/motoros2_config.yaml
+++ b/config/motoros2_config.yaml
@@ -177,9 +177,11 @@ namespace_tf: true
 # When using a 7 axis robot arm with an elbow joint (E) in the middle of the
 # arm, the elbow axis should be listed last (SLURBTE).
 #
-# DEFAULT: "group_x/joint_y"
+# DEFAULT:
+# - Single-group: "joint_y"
+# - Multi-group: "group_x/joint_y"
 #joint_names:
-#- [group_1/joint_1, group_1/joint_2, group_1/joint_3, group_1/joint_4, group_1/joint_5, group_1/joint_6]
+#- [joint_1, joint_2, joint_3, joint_4, joint_5, joint_6]
 
 #-----------------------------------------------------------------------------
 # Logging settings
--- a/config/motoros2_config.yaml
+++ b/config/motoros2_config.yaml
@@ -178,8 +178,8 @@ namespace_tf: true
 # arm, the elbow axis should be listed last (SLURBTE).
 #
 # DEFAULT:
-# - Single-group: "joint_y"
-# - Multi-group: "group_x/joint_y"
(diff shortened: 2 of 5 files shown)


config/motoros2_config.yaml
```diff
@@ -177,9 +177,11 @@ namespace_tf: true
 # When using a 7 axis robot arm with an elbow joint (E) in the middle of the
 # arm, the elbow axis should be listed last (SLURBTE).
 #
-# DEFAULT: "group_x/joint_y"
+# DEFAULT:
+# - Single-group: "joint_y"
+# - Multi-group: "group_x/joint_y"
```
gavanderhoorn, line 182:
```suggestion
# - Single-group: "joint_n"
# - Multi-group: "group_n/joint_m"
```

let's switch to more common notation for this sort of thing. It would also align better with the FAQ entry.

README.md
```diff
@@ -641,12 +641,12 @@ follow_joint_trajectory:
...
-   - group_1/joint_3
-   - group_1/joint_4
-   - group_1/joint_5
-   - group_1/joint_6
+   - joint_1
+   - joint_2
+   - joint_3
+   - joint_4
+   - joint_5
+   - joint_6
```
gavanderhoorn, line 649: should we perhaps include an example for multi-group setups as well?

Once/if you know, it's trivial, but that's always the case with documentation.

src/ControllerStatusIO.c
```diff
@@ -99,7 +99,14 @@ BOOL Ros_Controller_Initialize()
...
                     {
-                        sprintf(g_nodeConfigSettings.joint_names[(groupIndex * MP_GRP_AXES_NUM) + jointIndex], DEFAULT_JOINT_NAME_FMT, groupIndex + 1, jointIndex + 1);
+                        if (g_Ros_Controller.numGroup == 1)
+                        {
+                            sprintf(g_nodeConfigSettings.joint_names[(groupIndex * MP_GRP_AXES_NUM) + jointIndex], DEFAULT_JOINT_NAME_FMT_SINGLE, jointIndex + 1);
+                        }
+                        else
+                        {
+                            sprintf(g_nodeConfigSettings.joint_names[(groupIndex * MP_GRP_AXES_NUM) + jointIndex], DEFAULT_JOINT_NAME_FMT_MULTI, groupIndex + 1, jointIndex + 1);
+                        }
```
gavanderhoorn, line 109: what about using the `if` to select the appropriate format constant? Would result in less duplicated code.

jimmy-mcelwain: The format constants have different number of format specifiers/`sprintf`s have different numbers of arguments (joint vs group + joint). So shouldn't it stay separate?

config/motoros2_config.yaml
jimmy-mcelwain: Good point

src/ControllerStatusIO.c
gavanderhoorn: Ah, yes, you're right. That's unfortunate.

*Technically*, we could just always pass `groupIndex` and `jointIndex` to `sprintf(..)`, even if one of the formats only uses the first. According to the C standard, excess arguments should be ignored.

But we probably don't want to / can't rely on that.

README.md
jimmy-mcelwain: I have actually not yet set up MoveIt on a multi-group setup, so I am not totally sure what this should look like. If you would like to provide an example, feel free. Either way is okay with me.

ted-miller: I've made a note in https://github.com/Yaskawa-Global/motoros2/issues/74#issuecomment-2554085013

ted-miller (approved):

Issue #336, opened 2024-11-22
Improve status/error reporting of point queue service (and related functionality)


gavanderhoorn:
Context: #230 & #335.

[QueueTrajPoint](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/31028bbcea463e99f74e2fd6a4a467998ffc1e6f/srv/QueueTrajPoint.srv) returns a `result_code`, with one of the values defined in [QueueResultEnum.msg](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/31028bbcea463e99f74e2fd6a4a467998ffc1e6f/msg/QueueResultEnum.msg).

As evidenced by the two references discussions however, these result codes are insufficient when trying to diagnose failures that might occur when enqueuing (new) trajectory points.

We should improve status and error reporting of that service, such that usage of the debug logger should not be required in order to understand what's going on.

PR #341, opened 2024-12-09
Forward Init_Trajectory_Status error message to queue_traj_point

jimmy-mcelwain:
#336 

All FAILURE error codes/messages from Ros_MotionControl_InitPointQueue (and further Ros_MotionControl_Init) previously got squashed into success/failure. This commit maintains the error meaning

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -607,7 +607,7 @@ BOOL Ros_MotionControl_AddPulseIncPointToQ(CtrlGroup* ctrlGroup, Incremental_dat
     return TRUE;
 }
 
-UINT8 Ros_MotionControl_ProcessQueuedTrajectoryPoint(motoros2_interfaces__srv__QueueTrajPoint_Request* request)
+UINT16 Ros_MotionControl_ProcessQueuedTrajectoryPoint(motoros2_interfaces__srv__QueueTrajPoint_Request* request)
 {
     if (Ros_MotionControl_MustInitializePointQueue)
     {
@@ -617,9 +617,13 @@ UINT8 Ros_MotionControl_ProcessQueuedTrajectoryPoint(motoros2_interfaces__srv__Q
         status = Ros_MotionControl_InitPointQueue(request);
 
         if (status == INIT_TRAJ_OK)
+        {
             return motoros2_interfaces__msg__QueueResultEnum__SUCCESS;
+        }
         else
-            return motoros2_interfaces__msg__QueueResultEnum__INIT_FAILURE;
+        {
+            return status;
+        }
     }
 
     //------------------------------------------------------------
--- a/src/ServiceQueueTrajPoint.c
+++ b/src/ServiceQueueTrajPoint.c
@@ -85,6 +85,7 @@ void Ros_ServiceQueueTrajPoint_Trigger(const void* request_msg, void* response_m
     }
     else
     {
+        const char* initMsg;
         response->result_code.value = Ros_MotionControl_ProcessQueuedTrajectoryPoint(request);
 
         switch (response->result_code.value)
@@ -92,11 +93,6 @@ void Ros_ServiceQueueTrajPoint_Trigger(const void* request_msg, void* response_m
         case motoros2_interfaces__msg__QueueResultEnum__SUCCESS: 
             rosidl_runtime_c__String__assign(&response->message,
                 motoros2_interfaces__msg__QueueResultEnum__SUCCESS_STR);
-            break;
-
-        case motoros2_interfaces__msg__QueueResultEnum__INIT_FAILURE:
-            rosidl_runtime_c__String__assign(&response->message, 
-                motoros2_interfaces__msg__QueueResultEnum__INIT_FAILURE_STR);
             break;
 
         case motoros2_interfaces__msg__QueueResultEnum__BUSY:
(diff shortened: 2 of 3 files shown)


jimmy-mcelwain:
I am going to make changes to this for sure but I wanted feedback. I did a pretty straightforward forwarding of errors here. I think that it works fine (in my limited testing) and it's an improvement over what it was before, but there are other changes that should be made to improve clarity and durability. 

The other changes that I think should happen are:
1.) At the very least, if we go with the forwarding approach I took here, in motoros2_interfaces, [`QueueResultEnum.msg` uses `UINT8` as the `value`](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/952007d44910e9027b3b4c2261cde03406b98f9d/msg/QueueResultEnum.msg#L26
) and [`InitTrajEnum.msg` uses `UINT16` as the `value`](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/952007d44910e9027b3b4c2261cde03406b98f9d/msg/InitTrajEnum.msg#L65). Those should be the same. 
2.) This is more of an organization complaint. But I don't really like that sending a `/follow_joint_trajectory` goal (indirectly) returns an `InitTrajEnum` value that gets reported without modification. However, submitting a point using `queue_traj_point` returns either a `QueueResultEnum` value or an `InitTrajEnum` value which gets either squashed (current state of main) or forwarded (behavior in this PR). The fact that `queue_traj_point` can send a message that is defined in a different `msg` file for a different action seems messy. Especially since the messages reference "trajectory" and `/start_traj_mode`. I don't know what the solution would be (duplicating the relevant codes between the 2 messages, having both `queue_traj_point` and `follow_joint_trajectory` share the same `msg` file, or having both messages inherit some of the same constants (not sure how this would be implemented)). It should be okay to keep it the same, it just seems a bit messy/hard to follow.

ted-miller:
> At the very least, if we go with the forwarding approach I took here, in motoros2_interfaces, [QueueResultEnum.msg uses UINT8 as the value](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/952007d44910e9027b3b4c2261cde03406b98f9d/msg/QueueResultEnum.msg#L26) and [InitTrajEnum.msg uses UINT16 as the value](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/952007d44910e9027b3b4c2261cde03406b98f9d/msg/InitTrajEnum.msg#L65). Those should be the same.

Agreed. My only concern is breaking compatibility of messages. But I'm not sure that can be helped.


> This is more of an organization complaint. But I don't really like that sending a /follow_joint_trajectory goal (indirectly) returns an InitTrajEnum value that gets reported without modification. However, submitting a point using queue_traj_point returns either a QueueResultEnum value or an InitTrajEnum value

Personally, I'm ok with this. Ultimately, both the service and the action are initializing a trajectory. It's just a matter of "when" the points get submitted.

To me, I don't mind it. But if you have some suggestion of how to reorganize, we can take a look. Maybe some superset message for "motion init result"?

jimmy-mcelwain:
I think that this is probably good for now. I can't really think of a better way of organizing the messages. I'll prepare a PR changing `UINT8->UINT16` in [`QueueResultEnum.msg`](https://github.com/Yaskawa-Global/motoros2_interfaces/blob/952007d44910e9027b3b4c2261cde03406b98f9d/msg/QueueResultEnum.msg#L26) I can get that PR ready there.

ted-miller:
Depends on https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30

src/MotionControl.c
```diff
@@ -630,9 +630,13 @@ UINT8 Ros_MotionControl_ProcessQueuedTrajectoryPoint(motoros2_interfaces__srv__Q
         status = Ros_MotionControl_InitPointQueue(request);
 
         if (status == INIT_TRAJ_OK)
+        {
             return motoros2_interfaces__msg__QueueResultEnum__SUCCESS;
+        }
         else
-            return motoros2_interfaces__msg__QueueResultEnum__INIT_FAILURE;
+        {
+            return (UINT8) status;
```
ted-miller, line 638: This function needs to return uint16

src/ServiceQueueTrajPoint.c
```diff
@@ -110,8 +106,16 @@ void Ros_ServiceQueueTrajPoint_Trigger(const void* request_msg, void* response_m
...
+            initMsg =  Ros_ErrorHandling_Init_Trajectory_Status_ToString((Init_Trajectory_Status)response->result_code.value);
+            if (strcmp(initMsg, motoros2_interfaces__msg__InitTrajEnum__INIT_TRAJ_UNSPECIFIED_STR) == 0) {
+                rosidl_runtime_c__String__assign(&response->message,
+                    "Unspecified error occurred while adding the trajectory point to the queue.");
+            }
+            else
+            {
+                rosidl_runtime_c__String__assign(&response->message,
+                    initMsg);
+            }
```
ted-miller, line 118: Can you add some comment regarding what's going on here?

ted-miller (changes requested):

jimmy-mcelwain:
I made the requested changes (amended) then rebased. I did a little bit of testing, and I think that it is alright.

ted-miller:
bumping milestone because we need updated motoros2_interfaces inside micro_ros_motoplus

ted-miller:
Pulling the milestone back again. See https://github.com/Yaskawa-Global/motoros2/pull/364#issue-2828533968

We need to make new builds of libmicroros.

ted-miller:
libmicroros has been updated to point to motoros2_interfaces v0.2.0. 

See #366

Issue #347, opened 2024-12-18
Unnecessary debug message for `prevPulsePos` 


jimmy-mcelwain:
Within `Ros_MotionControl_Init`, the same value is written to both `curPos` and `prevPulsePos`, and then they are both broadcasted out as if they are meant to be different values (current position and previous position). 

We could either
1.) Get rid of the `curPos` variable and print-out altogether and keep the setting/printing of `prevPulsePos` (or keep curPos and just copy its contents to prevPulsePos later to keep the naming scheme more accurate)
2.) Wait until close to the end of the function, or at least until the print-outs to set `prevPulsePos` to the current position. 

This is in `Ros_MotionControl_Init`, so I don't think that it makes a lot of sense to print out the previous pulse data, so I'd prefer 1.

https://github.com/Yaskawa-Global/motoros2/blob/f7cd860bc9b60572a1ba53d9711164f47bfab1e7/src/MotionControl.c#L189-L212

ted-miller:
I suspect this is from the copy/paste of MotoROS1 code. 

I'd be in favor of removing the `curPos` variable entirely. But, if you think the name makes more sense semantically, then we can just change it to a `long*` and point it to the `prevPulsePos` array.

PR #348, opened 2024-12-19
Get rid of duplicate position values and clean up names in `Ros_MotionControl_Init`

jimmy-mcelwain:
Get rid of duplication and error message from #347

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -29,8 +29,8 @@ BOOL Ros_MotionControl_MustInitializePointQueue = TRUE; //first point of streami
 
 Init_Trajectory_Status Ros_MotionControl_Init(rosidl_runtime_c__String__Sequence* sequenceGoalJointNames, trajectory_msgs__msg__JointTrajectoryPoint__Sequence* sequenceOfPoints)
 {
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Language-wise it's a bit strange to have a check with a comment stating it's going to check whether the start of the trajectory matches the *current command position*, but the variable it checks is then called `prevPulsePos`.

jimmy-mcelwain:
I could add `curPos` back as a pointer that aliases `prevPulsePos` as suggested in https://github.com/Yaskawa-Global/motoros2/issues/347#issuecomment-2554132165, or I could keep curPos as is and just copy its contents to `prevPulsePos` later on, but I just don't know what the cleanest option is.

ted-miller:
Add curpos back in for the sake of comparison and printing. Then copy it into prevPulsePos before setting `ctrlGroup->hasDataToProcess = TRUE`. And add some comment to the effect that the starting point is the previous-point in the motion algorithm.

jimmy-mcelwain:
Rebased and added a few more changes. Functionally nothing should change. I did a quick test and saw no changes, but I will test further

ted-miller (approved):
I have tested. The changes look good to me.

Issue #268, opened 2024-06-24
Extend `motoRosAssert(..)` to output expected value


gavanderhoorn:
[Current implementation](https://github.com/Yaskawa-Global/motoros2/blob/d8b75ddcfe763db44f6f13ebc66c28469d154acd/src/ErrorHandling.c#L79-L82) posts an alarm and enters an infinite loop printing `motoRosAssert: MotoROS2: Fatal Error (subcode: N)`.

While that works, it would be more helpful to us (and users) if it printed the expected value (whatever that would mean for specific `motoRosAssert(..)` invocations).

Example:

https://github.com/Yaskawa-Global/motoros2/blob/d8b75ddcfe763db44f6f13ebc66c28469d154acd/src/PositionMonitor.c#L55-L62

an extended version of `motoRosAssert(..)` would output the values of `ret` and `RCL_RET_OK` as part of the message.

gavanderhoorn:
Context: #266 (among others).

jimmy-mcelwain:
What would the scope of this issue be? The conditions tested by `motoRosAssert` have a wide variety of operators. An initial search shows ==, !=, >=, <=, non-NULL, etc. are used. It could also cover several different types. Would we want a function for each of these checks? The implementation wouldn't be difficult, but I don't know where the best place to stop would be.

gavanderhoorn:
We should be able to use [macro stringification](https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html) to avoid having to create `N` different versions of `motoRosAssert(..)`.

There are various C / C++ unit testing frameworks which use this approach to print more informative result/error messages.

In the end `motoRosAssert(1 < x, ...)` would result in the `1 < x` being available as a string itself, in addition to being a conditional. That string-ified version could then used as part of the message.

jimmy-mcelwain:
I have some work related to this [here](https://github.com/Yaskawa-Global/motoros2/tree/OutputExpectedValueOnAssertion). I am using macro stringification to print out variable names and values. The macros make it a bit ugly, but it adds a lot of information to the debug log without making calls to the assertion functions more difficult. I believe that I could get rid of the `motoRos_complete_ASSERT_XX_XXX` functions and just define everything with the macros, but I am not very familiar with the convention there. There's some code duplication that I could try to cut down on too. I haven't tested it very much, I just ran it once and it seemed to work. But is this the sort of thing that you are looking for @gavanderhoorn?

[...]

PR #350, opened 2024-12-30
Output expected value on assertion

jimmy-mcelwain:
In reference to #268. 

I am going to make this a draft because I definitely want feedback and expect that I will have to make changes. 

The second most recent commit (as of now), 6668a4d0b7b0506ce0427013e578b72d66789bb6 had a function for each assertion, but there was a lot of code duplication. 84e5e927fc473a191779ad1e411cfb017052a26d changed that by using more macros.


src/ErrorHandling.h
```diff
@@ -235,30 +238,292 @@ typedef enum
...
+        { \
+            char motoRos_error_handling_assert_msg[ERROR_MSG_MAX_SIZE] = {0}; \
+            snprintf(motoRos_error_handling_assert_msg, ERROR_MSG_MAX_SIZE, msgFmtOnFailure, ##__VA_ARGS__); \
+            Ros_Controller_SetIOState(IO_FEEDBACK_FAILURE, TRUE); \
+            Ros_Controller_SetIOState(IO_FEEDBACK_INITIALIZATION_DONE, FALSE); \
+            mpSetAlarm(ALARM_ASSERTION_FAIL, motoRos_error_handling_assert_msg, subCodeOnFailure); \
+            FOREVER \
+            { \
+                Ros_Debug_BroadcastMsg("MotoROS2 Assertion Failure: %s (subcode: %d)", motoRos_error_handling_assert_msg, subCodeOnFailure); \
+                Ros_Debug_BroadcastMsg("Actual: %s == %d", actualName, (actual)); \
```
gavanderhoorn, line 254: do we really need two specialised versions for this?

I'm sort-of hoping/expecting we can use token pasting / concatenation and just build the expected vs actual string part of the msg using the pre-processor. I don't know whether that's true though, I expect you've tried.

src/ErrorHandling.h
```diff
@@ -235,30 +238,292 @@ typedef enum
...
+
+#define motoRos_ASSERT_GENERIC_COMPARISON_INT_MESSAGE(actual, benchmark, subCodeOnFailure, actualName, benchmarkName, op, msgFmtOnFailure, ...) \
+    do { \
+        if (!((actual) op (benchmark))) \
+        { \
+            char motoRos_error_handling_assert_msg[ERROR_MSG_MAX_SIZE] = {0}; \
+            snprintf(motoRos_error_handling_assert_msg, ERROR_MSG_MAX_SIZE, msgFmtOnFailure, ##__VA_ARGS__); \
+            Ros_Controller_SetIOState(IO_FEEDBACK_FAILURE, TRUE); \
+            Ros_Controller_SetIOState(IO_FEEDBACK_INITIALIZATION_DONE, FALSE); \
+            mpSetAlarm(ALARM_ASSERTION_FAIL, motoRos_error_handling_assert_msg, subCodeOnFailure); \
```
gavanderhoorn, line 250: I'm wondering whether this is what we'd want to do. With the current implementation, we'd be adding both `mpSetAlarm(..)` and IO state management code to all call-sites of this macro.

It'd be invisible -- because it's part of the macro body -- but it would still be there. At the very least it increases binary size, but it also seems like we should be able to call a function to which we delegate this responsibility -- didn't we have that before?

src/ErrorHandling.h
```diff
@@ -235,30 +238,292 @@ typedef enum
     SUBCODE_RCL_RCLC_API_ERROR,
 } ALARM_RCL_RCLC_FAIL_SUBCODE; //8017
 
+
+#define motoRos_ASSERT_GENERIC_COMPARISON_INT_MESSAGE(actual, benchmark, subCodeOnFailure, actualName, benchmarkName, op, msgFmtOnFailure, ...) \
+    do { \
+        if (!((actual) op (benchmark))) \
```
gavanderhoorn, line 244: This is a nice use of token pasting / concatenation, but I might suggest keeping the actual 'ops' in the macros that call this one.

That would both make those macros easier to read (as we just see them doing something like `if (!(x > y)) assert(my_msg)`) and it would reduce the responsibilities of this macro (as it now only needs to log the message and raise an alarm (but see my other comment about this)).

Less responsibilities would simplify the code (less nesting), which is always good I feel.

src/ErrorHandling.h
jimmy-mcelwain: I can move some of the macro stuff into a function to reduce the binary size.

src/ErrorHandling.h
jimmy-mcelwain: The problem is the format specifier in the print statement. In this case, I could probably get rid of one, because I only have specialized versions for pointers and integers and it wouldn't be disastrous to print out a pointer with say `%d` (or an integer with `%p`). But if we add asserts for other non-integer types like floats that wouldn't print properly with `%d` then we would definitely need specialized versions.

src/ErrorHandling.h
jimmy-mcelwain: Yeah, I think that the 2 layers are sort of needed if we want to 
1. Reduce code duplication
2. Print out really detailed messages
I can get rid of the second macro call if we give up on one of these two.

jimmy-mcelwain:
I made some small changes that should at least reduce binary size a little bit, nothing big. I'll have to test it more, but I'm mostly looking for design suggestions. I am open to changes, but this has also gotten a bit more complicated than I was hoping and I'm no macro expert.

PR #337, opened 2024-11-25
Jazzy Jalisco

jimmy-mcelwain:
See #333. This updates MotoROS2 to make it compatible with Jazzy Jalisco, the new long term support release of ROS2.

PR #351, opened 2025-01-12
ci: bump M+ libmicroros version

gavanderhoorn:
As per title.

Should fix CI.

---

Edit: test run: https://github.com/Yaskawa-Global/motoros2/actions/runs/12735310867 (I should really fix CI to not skip the `msbuild` job if the workflow itself gets edited).


gavanderhoorn:
We should rebase #322, #337, #341 and #350 after we merge this one.

Issue #346, opened 2024-12-18
`Ros_MotionControl_ClearQ_All()` always returns `TRUE`


jimmy-mcelwain:
From https://github.com/Yaskawa-Global/motoros2/issues/76#issuecomment-2032477683, 

> bRet is the return value of ROS_MotionControl_ClearQ_ALL(), this value is always TRUE.

https://github.com/Yaskawa-Global/motoros2/blob/f7cd860bc9b60572a1ba53d9711164f47bfab1e7/src/MotionControl.c#L1275-L1299

Every other instance of
`if (mpSemTake(q->q_lock, Q_LOCK_TIMEOUT) == OK)`
has some sort of error mechanism/reporting if `q_lock` cannot be acquired in time. If the semaphore cannot be acquired within `Ros_MotionControl_ClearQ_All`, there should be a debug broadcast indicating that, and more importantly `bRet` should be set to `FALSE`.

PR #352, opened 2025-01-13
Fix stop_traj_mode errors

### LINKED ISSUE #76: `stop_traj_mode`: "Couldn't stop trajectory mode: unknown error"
While testing the node described in #66, I ran into the following problem.

Invoking the `stop_traj_mode` service returned:

```console
user@host:~$ ros2 service call /yrc1000/stop_traj_mode std_srvs/srv/Trigger 
requester: making request: std_srvs.srv.Trigger_Request()

response:
std_srvs.srv.Trigger_Response(success=False, message="Couldn't stop trajectory mode: unknown error")
```

debug log:

```
[2023-06-26 17:23:33.703843] [192.168.255.1:58950]: stop_traj_mode: attempting to stop trajectory mode
[2023-06-26 17:23:33.777264] [192.168.255.1:58950]: WARNING: Message processing not stopped before clearing queue
[2023-06-26 17:23:33.777611] [192.168.255.1:58950]: stop_traj_mode: StopMotion failed (unknown error)
[2023-06-26 17:23:33.783029] [192.168.255.1:58950]: Stopping point-queue motion mode. Please call 'start_point_queue_mode' to start a new queue.
```

This was after queuing points for a couple of trajectories using MoveIt.

---

Edit: `INIT_ROS` was terminated, and servos eventually also de-energised. Not sure what the error cause really was.

jimmy-mcelwain:
I believe that this should fix #76. I also included a fix for #346 because it made solving #76 a little bit less confusing. 

The problem is that made it so `stop_traj_mode` sometimes failed is that  `hasDataToProcess` is never set to `FALSE` during point queue mode, even after `bStopMotion` is set to `true`. That is because `ctrlGroup->trajectoryIterator->valid` is set to `FALSE` after the first point is queued. Because of this, the first condition in the code snippet below fails, and it never reaches the `if(bStopMotion)` block. 

https://github.com/Yaskawa-Global/motoros2/blob/b0eff07cf9a962aa134d96db2b542ae02421d468/src/MotionControl.c#L370-L378 

Slightly related, but `hasDataToProcess` a bit misleading. In point queue mode, there can be no data sitting in the queue/no data to process, but `hasDataToProcess` is true. I do believe that it is an accurate name for trajectory mode, though.

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -360,14 +360,8 @@ void Ros_MotionControl_AddToIncQueueProcess(CtrlGroup* ctrlGroup)
         if (Ros_MotionControl_AllGroupsInitComplete)
         {
             // if there is no message to process, delay and try again
-            if (ctrlGroup->hasDataToProcess && ctrlGroup->trajectoryIterator != NULL && ctrlGroup->trajectoryIterator->valid)
+            if (!g_Ros_Controller.bStopMotion && ctrlGroup->hasDataToProcess && ctrlGroup->trajectoryIterator != NULL && ctrlGroup->trajectoryIterator->valid)
             {
-                if (g_Ros_Controller.bStopMotion)
-                {
-                    bzero(ctrlGroup->trajectoryToProcess, sizeof(ctrlGroup->trajectoryToProcess));
-                    ctrlGroup->hasDataToProcess = FALSE;
-                    continue;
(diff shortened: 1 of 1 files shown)


ted-miller (approved):
Only a small functional change. It tested OK.

Issue #295, opened 2024-08-21
Document symptoms pointing to faulty / flaky networks


gavanderhoorn:
Context: #292.

Issue #288, opened 2024-08-05
Document potential firewall configuration issues


gavanderhoorn:
Context: #283.

Turned out to be `ufw` blocking incoming UDP traffic on the port the Agent was configured to listen (https://github.com/Yaskawa-Global/motoros2/discussions/283#discussioncomment-10240909).

Checking/configuring a/the firewall should probably be added as a step in the general installation instructions.

Issue #242, opened 2024-04-19
Document potential issue with DHCP network configuration


gavanderhoorn:
As noted by @ted-miller in https://github.com/Yaskawa-Global/motoros2/discussions/238#discussioncomment-9159344:

> Although DHCP is setting the controller IP to `10.1.1.x`, MotoROS2 is reading the values that were input in Maintenance mode.
> 
> ![image](https://private-user-images.githubusercontent.com/4623811/323748323-d2aa4377-82ab-4e83-a205-bb2244da1f9b.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTM1MTM2NzEsIm5iZiI6MTcxMzUxMzM3MSwicGF0aCI6Ii80NjIzODExLzMyMzc0ODMyMy1kMmFhNDM3Ny04MmFiLTRlODMtYTIwNS1iYjIyNDRkYTFmOWIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDQxOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA0MTlUMDc1NjExWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ODQzMzI0OTA2MTAyZDJmM2ViZjliODIzMzc3MTc3OWRlMGUxMzFjMTYwNTE1ZWU2ZWRkZGRhMGEyMzIxZDI5MiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.kPb2bqjtlr89aw4x5efR5Cnc4KQ8qdSK-3lMn895Hvo)
> 
> If you change these values in maintenance mode to be on the `10.1.1.x` subnet, it will get rid of the alarm.
> 
> We generally don't recommend DHCP for robotic systems since the robot needs to definitively know the address of the agent PC.

Issue #260, opened 2024-06-11
Readme: Explanation of Agent IP and Port


ted-miller:
The readme has a section on [configuration](https://github.com/Yaskawa-Global/motoros2?tab=readme-ov-file#configuration), which tells you to input the address of the uros agent.

But the agent isn't explained [until later](https://github.com/Yaskawa-Global/motoros2?tab=readme-ov-file#the-micro-ros-agent) in the readme.

For someone unfamiliar with uros, they may not know what address to put into the config file.

@jimmy-mcelwain, can you suggest some changes to the setup procedure to make it more clear?

PR #356, opened 2025-01-22
Document potential network issues

jimmy-mcelwain:
There are a couple of related issues about documenting problems with network configuration. #242, #288, and #295. I've added some information to the readme and troubleshooting guide about avoiding/resolving these network issues.


README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
  1. move to `MotoPlus FUNC.`, make sure it is set to `USED`. If it isn't, set it to `USED`
  1. move cursor down to `MOTOMAN DRIVER` and make sure it is set to `USED`. If it isn't, set it to `USED`
 
+#### Checking network configuration
```
ted-miller, line 262:
```suggestion
### Checking network configuration
```

ted-miller: Please also add this to the TOC

README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
...
+
+While still in *Maintenance* mode and *MANAGEMENT* security level, check the following network settings:
+
+ 1. touch `[System Info]`→`[Setup]` and select `OPTION FUNCTION`
+ 1. move to `LAN INTERFACE SETTING`, make sure `IP ADDRESS SETTING(LAN[X])` for the NIC you are using is set to `MANUAL SETTING`. If it isn't, set it to `MANUAL SETTING`
```
ted-miller, line 267: This is different on DX200. I'm not sure about the YRCmicro

README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
...
+ 1. Before leaving this screen, make sure that the robot controller's IP address is on the subnet of the PC running the micro-ROS agent application, and that the PC is on the subnet of the robot controller. To do this, choose one of the following options:
```
ted-miller, line 268: > Before leaving this screen, make sure that the robot controller's IP address is on the subnet of the PC running the micro-ROS agent application, and that the PC is on the subnet of the robot controller.

Is that saying the same thing twice? Or am I misunderstanding?

ted-miller, line 268:
```suggestion
 1. Before leaving this screen, make sure that the robot controller's IP address is on the subnet of the PC running the micro-ROS agent application, and that the PC is on the subnet of the robot controller.
To do this, choose one of the following options:
```

README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
...
+    - Modify the `agent_ip_address` key in the `motoros2_config.yaml` file and specify an IP address that is on the robot's subnet. Then [propagate the changes to the Yaskawa controller](#updating-the-configuration). You will need to ensure that the PC running the micro-ROS agent application uses this static IP address on the network port connected to the robot controller.
```
ted-miller, line 269: > Modify the `agent_ip_address` key in the `motoros2_config.yaml` file

This comes later in the instructions.

README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
...
+    - Modify the robot controller's IP and subnet mask so it is on the subnet of the PC running the micro-ROS agent.
```
ted-miller, line 270: I suggest that we _only_ give this instruction

README.md
```diff
@@ -259,6 +259,17 @@ Use the following steps to verify MotoPlus has been correctly configured for Mot
...
+    - Modify the robot controller's network settings to add a gateway which can reach the IP address of the subnet of the PC running the micro-ROS agent.
```
ted-miller, line 271: I don't think we've tested using the agent through a gateway. It _should_ work... but then we have to worry about port-forwarding and stuff.

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
 When you encounter an issue using MotoROS2, please start the debug client script and keep it running in the background while you reproduce the issue.
 Attach the log it produces and a copy of the `PANELBOX.LOG` from the robot's teach pendant to any support tickets you open on the [Issue tracker](https://github.com/yaskawa-global/motoros2/issues).
 
+## Network issues
```
ted-miller, line 27: For all of these, please put each sentence on a new line. The markdown will recombobulate it into a single line. (two line breaks are required for a new line)

I don't actually know why this is a thing. But it maintains consistency throughout the doc

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+If you are unable to connect the PC running the micro-ROS agent application and the controller, the network configuration may be the issue. Network problems can sometimes be the root of seemingly unrelated problems, such as rcl/rclc errors.
+
+To troubleshoot network issues, first ensure that the robot controller's IP address is `MANUAL SETTING`. It is important that it is not using DHCP. [See here](../README.md#checking-network-configuration).
+
+If that is not the problem, check that the `agent_ip_address` matches the IP address of the PC. Also make sure that the robot controller IP address is on the same subnet as the PC.
```
ted-miller, line 33:
```suggestion
If that is not the problem, check that the `agent_ip_address` in the yaml configuration matches the IP address of the PC.
Also make sure that the robot controller IP address is on the same subnet as the PC.
```

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+If there is a mismatch somewhere, choose one of the following options:
+
+1. Modify the `agent_ip_address` key in the `motoros2_config.yaml` file and specify an IP address that is on the robot's subnet. Then [propagate the changes to the Yaskawa controller](../README.md#updating-the-configuration). You will need to ensure that the PC running the micro-ROS agent application uses this static IP address on the network port connected to the robot controller. You will also need to make sure tha the subnet mask matches that of the controller.
+2. Modify the robot controller's IP and subnet mask so it is on the subnet of the PC running the micro-ROS agent.
+3. Modify the robot controller's network settings to add a gateway which can reach the Agent's IP address.
+
+Refer to the relevant Yaskawa Motoman documentation for information on how to change the controller's network configuration.
```
ted-miller, line 41: Is there a public Knowledge Center article that we can link to?

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+Test if the two are connected by opening a terminal on the PC and attempting to ping the robot controller's IP. For example
+
+```shell
+ping 192.168.1.31
+```
+
+If you receive a near-immediate response upon attempting to ping the controller, then the potential network issues mentioned above are resolved. Try to connect again via the micro-ROS agent. If the micro-ROS agent still cannot connect to the controller, there may be a firewall configuration issue.
+
+If you do not receive a near-immediate after attempting to ping the controller, there is still some issue. You may have made a mistake previously, or there could be a firewall issue, or there could be a network hardware issue.
```
ted-miller, line 51: Maybe suggest plugging PC directly into the robot (bypassing any additional hardware, such as switches). Also suggest trying a different cable.

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+```shell
+sudo ufw show added
+```
+
+Look at the rules and remove any that may interfere with the connection for MotoROS2. For example, if the rule `ufw deny 8888/udp` was a present, you would want to remove it.
+
+Using `ros2 multicast send` and `ros2 multicast receive` can be helpful for diagnosing firewall problems.
+
+For further discussion on how to fix firewall issues, see [here](https://stackoverflow.com/a/75087882).
```
ted-miller, line 73: I don't like the "conclusion" of that link. It seems to contradict what you say here.

> So for now, I am simply assuming ros2 multicast is a tool which should not be used and that the following command must be ran when running ROS2 systems on several machines on the same network.

> `$ sudo ufw allow in proto udp from 192.168.1.0/24`

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+sudo ufw status
+```
+
+If the status shows as active, you should use the following command to see what firewall rules have been added that may prevent communication.
+
+```shell
+sudo ufw show added
+```
+
+Look at the rules and remove any that may interfere with the connection for MotoROS2. For example, if the rule `ufw deny 8888/udp` was a present, you would want to remove it.
```
ted-miller, line 69: What about explicitly allowing that port?

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+### Other potential network problems
+
+There are other potential network configuration problems.
+
+Make sure that there are no IP address conflicts with either the client PC or the robot controller.
+
+Make sure that all of the connections between the controller and the PC are solid, and that all of the cables and peripheral equipment works properly.
+
+If all else fails, Wireshark can be a valuable tool for finding where network problems lie.
```
ted-miller, line 87: I don't know that we should suggest this without providing some concrete definition of what to look for.

doc/troubleshooting.md
```diff
@@ -24,6 +24,68 @@ On Windows the `.cmd` file changes the `CWD` to the location of the script befor
...
+
+For further discussion on how to fix firewall issues, see [here](https://stackoverflow.com/a/75087882).
+
+If you wish to disable your firewall to test if it is the source of your problem, you may do so, but be aware of the risks associated with that.
+
+It is possible that firewall rules are not visible via `ufw`, but that they are still in place. That goes beyond the scope of this troubleshooting guide, but it is a possibility.
+
+### Other potential network problems
+
+There are other potential network configuration problems.
```
ted-miller, line 81: I think this line is not needed, due to the heading for this section.

doc/troubleshooting.md
ted-miller: This also applies to the other sections too.

doc/troubleshooting.md
```diff
@@ -1199,6 +1263,7 @@ The actual `rcl` return value is given in the text of the alarm (represented by
 Info about the return value indicating an error can be found [here](https://docs.ros2.org/latest/api/rcl/types_8h_source.html).
 Some problems are user-serviceable based on the diagnoses, and others are indicative of an error within MotoROS2.
 Please reference the troubleshooting steps of the secondary alarm code for more information.
+If the the secondary alarm code is not helpful, then consider checking for faults in the network connection between the PC and the robot controller. If the connection is not stable, `rcl` and `rclc` may return an error.
```
ted-miller, line 1266:
```suggestion
If the the secondary alarm code is not helpful, then consider checking for faults in the [network connection](#network-issues) between the PC and the robot controller. If the connection is not stable, `rcl` and `rclc` may return an error.
```

ted-miller (changes requested):

README.md
jimmy-mcelwain: We already have instructions in the troubleshooting guide saying to modify the gateway in case of networking problems.  [1](https://github.com/Yaskawa-Global/motoros2/blob/19b98386b4ca0b2cb20dbbc3520079f5d6cf9123/doc/troubleshooting.md?plain=1#L519), [2](https://github.com/Yaskawa-Global/motoros2/blob/19b98386b4ca0b2cb20dbbc3520079f5d6cf9123/doc/troubleshooting.md?plain=1#L689), [3](https://github.com/Yaskawa-Global/motoros2/blob/19b98386b4ca0b2cb20dbbc3520079f5d6cf9123/doc/troubleshooting.md?plain=1#L978). I haven't tested it either, I was just trying to keep consistency with the existing documentation

doc/troubleshooting.md
jimmy-mcelwain: [This](https://knowledge.motoman.com/hc/en-us/articles/4405490891543-Understanding-Network-Settings) is the best public article we have, I think. It isn't exactly what we're looking for, though. I was just copying what was already the troubleshooting guide under other network config problems.

README.md
ted-miller: Ok.

gavanderhoorn:
Seeing as there is quite some content added, should we perhaps introduce a separate file to discuss these network related issues / network setup?

The readme is already quite long, and we should perhaps not add more to it. Same for the troubleshooting document. Longer remedy sections don't always make things clearer.

doc/troubleshooting.md
gavanderhoorn: > I don't actually know why this is a thing.

because it makes the diffs much easier to read and merging changes easier.

doc/troubleshooting.md
gavanderhoorn: I would also suggest we don't include a dependency on an external site like this.

If we feel this is a good diagnostic step or remedy, we should include it in our documentation.

jimmy-mcelwain: I will just remove this both the `ros2 multicast` and the link. I have included more instructions. The `ros2 multicast` can sometimes be a useful diagnostic tool, but it depends on the specific set of firewall rules, and introducing it may make things more confusing.

jimmy-mcelwain:
> Seeing as there is quite some content added, should we perhaps introduce a separate file to discuss these network related issues / network setup?

I've separated it into a new file

README.md
```diff
@@ -54,15 +54,18 @@ The following sections document how to download, install, configure, use and tro
...
   - [Checking MotoPlus configuration](#checking-motoplus-configuration)
-  - [DX200, YRC1000 and YRC1000micro](#dx200-yrc1000-and-yrc1000micro)
-- [Building from source](doc/build_from_source.md)
-- [Updating the configuration](#updating-the-configuration)
-- [The micro-ROS Agent](#the-micro-ros-agent)
+  - [Installing MotoROS2 out file](#installing-motoros2-out-file)
+- [Example INFORM jobs](#example-inform-jobs)
+- [Understanding the micro-ROS Agent](#understanding-the-micro-ros-agent)
+- [Configuration](#configuration)
+  - [Network Configuration](#network-configuration)
```
ted-miller, line 64:
```suggestion
  - [Network Configuration](doc/network_configuration.md)
```

Better to skip straight to that document rather than forcing another click.

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
+﻿<!--
+SPDX-FileCopyrightText: 2023, Yaskawa America, Inc.
+SPDX-FileCopyrightText: 2023, Delft University of Technology
+
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Network Configuration
+
+You will need to use a LAN cable to allow the robot controller to communicate with the PC running the [micro-ROS Agent](../README.md#the-micro-ros-agent).
```
ted-miller, line 10:
```suggestion
You will need to use a LAN cable to allow the robot controller to communicate with the PC running the [micro-ROS Agent](../README.md##understanding-the-micro-ros-agent).
```

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Configuring Controller
+
+### YRC1000 and YRC1000micro
+
+1. boot the controller while holding `{MAIN MENU}` on the pendant keypad to enter *Maintenance* mode
+1. upgrade to *MANAGEMENT* security level by touching `[System Info]`→`[Security]` (default password is all `9`'s)
+1. touch `[System Info]`→`[Setup]` and select `OPTION FUNCTION`
+1. move to `LAN INTERFACE SETTING`
+1. make sure `IP ADDRESS SETTING(LAN[X])` for the port you are using is set to `MANUAL SETTING`. If it not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` for the NIC you are using (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
```
ted-miller, line 26:
```suggestion
1. make sure `IP ADDRESS SETTING(LAN[X])` for the port you are using is set to `MANUAL SETTING`.
If it not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` for the NIC you are using (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
```

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
...
+1. touch `[System Info]`→`[Setup]` and select `OPTION FUNCTION`
+1. move to `NETWORK SETTING` and then to `HOST SETUP`
+1. make sure that `IP ADDRESS SETTING` is set to `MANUAL SETTING`. If it is not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
+1. hit `{ENTER}` on the pendant keypad and touch `[OK]`
+
+## Get controller and PC on same subnet
+
+If the controller and PC are not on the same subnet, choose one of the following options:
+
+- Modify the `agent_ip_address` key in the `motoros2_config.yaml` file and specify an IP address that is on the robot's subnet. See [here](../README.md#configuration-file) for informationa about the config file. Then [propagate the changes to the Yaskawa controller](../README.md#updating-the-configuration). You will need to ensure that the PC running the micro-ROS agent application uses this static IP address on the network port connected to the robot controller.
```
ted-miller, line 42:
```suggestion
- Modify the `agent_ip_address` key in the `motoros2_config.yaml` file and specify an IP address that is on the robot's subnet.
See [here](../README.md#configuration-file) for informationa about the config file.
Then [propagate the changes to the Yaskawa controller](../README.md#updating-the-configuration).
You will need to ensure that the PC running the micro-ROS agent application uses this static IP address on the network port connected to the robot controller.
```

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
...
+sudo ufw show added
+```
+
+As an example of a rule that may interfere with MotoROS2 communication, `ufw deny from any proto udp` as a high priority rule would make it impossible to connect.
+
+There are many ways to fix the issue if it is a firewall.
+You could create firewall rules that specifically permit the connection with higher priority than the rules that deny communication.
+For example, if the rules from the following set of commands are given high priority, they will allow for UDP connection for host PC IP address `192.168.1.15` and controller IP address `192.168.1.31` on port `8888` for ROS2 communication and on port `21789` for debug communication.
+
+Note that this an extremely narrow set of rules that would permit communication. Much more concise rules could be applied.
```
ted-miller, line 101:
```suggestion
Note that this an extremely narrow set of rules that would permit communication.
Much more concise rules could be applied.
```

doc/troubleshooting.md
```diff
@@ -511,14 +515,7 @@ ALARM 8011
...
 
-Options:
-
-1. Modify the `agent_ip_address` key and specify an IP address that is on the robot's subnet.
-   Now follow the instructions [to propagate the changes to the Yaskawa controller](../README.md#updating-the-configuration).
-2. Modify the robot controller's IP so it is on the Agent's subnet.
-3. Modify the robot controller's network settings to add a gateway which can reach the Agent's IP address.
-
-Refer to the relevant Yaskawa Motoman documentation for information on how to change the controller's network configuration.
+Ensure that the robot controller's IP address is not assigned using DHCP and that the controller and the PC are on the same subnet. [See here](network_configuration.md#network-configuration).
```
ted-miller, line 518:
```suggestion
Ensure that the robot controller's IP address is not assigned using DHCP and that the controller and the PC are on the same subnet.
[See here](network_configuration.md#network-configuration).
```

doc/troubleshooting.md
```diff
@@ -681,14 +678,7 @@ Where `x` is either `1` or `2`.
...
```
ted-miller, line 681:
```suggestion
Ensure that the robot controller's IP address is not assigned using DHCP and that the controller and the PC are on the same subnet.
[See here](network_configuration.md#network-configuration).
```

doc/troubleshooting.md
```diff
@@ -975,7 +965,7 @@ If the alarm is raised again, and if auto-detection is not needed or desired, ma
 
 On YRC1000, set it to either `USER_LAN1` or `USER_LAN2`, depending on which LAN port is used to connect the controller to the PC running the micro-ROS Agent application.
 
-If auto-detection is to be used, verify `agent_ip_address` is set to an IP that can be reached by MotoROS2 over the LAN port which is connected to the PC running the micro-ROS Agent application (either directly, or via a default gateway configured on the controller).
+If auto-detection is to be used, verify `agent_ip_address` is set to an IP that can be reached by MotoROS2 over the LAN port which is connected to the PC running the micro-ROS Agent application (either directly, or via a default gateway configured on the controller). Also, make sure that on the robot controller, the `IP ADDRESS SETTING(LAN[X])` for the LAN port being used is set to `MANUAL SETTING`. [See here](../README.md#checking-network-configuration).
```
ted-miller, line 968:
```suggestion
If auto-detection is to be used, verify `agent_ip_address` is set to an IP that can be reached by MotoROS2 over the LAN port which is connected to the PC running the micro-ROS Agent application (either directly, or via a default gateway configured on the controller).
Also, make sure that on the robot controller, the `IP ADDRESS SETTING(LAN[X])` for the LAN port being used is set to `MANUAL SETTING`.
[See here](../README.md#checking-network-configuration).
```

doc/troubleshooting.md
```diff
@@ -1199,6 +1189,7 @@ The actual `rcl` return value is given in the text of the alarm (represented by
 Info about the return value indicating an error can be found [here](https://docs.ros2.org/latest/api/rcl/types_8h_source.html).
 Some problems are user-serviceable based on the diagnoses, and others are indicative of an error within MotoROS2.
 Please reference the troubleshooting steps of the secondary alarm code for more information.
+If the the secondary alarm code is not helpful, then consider checking for faults in the [network connection](network_configuration.md#network-issues) between the PC and the robot controller. If the connection is not stable, `rcl` and `rclc` may return an error.
```
ted-miller, line 1192:
```suggestion
If the the secondary alarm code is not helpful, then consider checking for faults in the [network connection](network_configuration.md#network-issues) between the PC and the robot controller.
If the connection is not stable, `rcl` and `rclc` may return an error.
```

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
...
+1. make sure `IP ADDRESS SETTING(LAN[X])` for the port you are using is set to `MANUAL SETTING`. If it not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` for the NIC you are using (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
+1. hit `{ENTER}` on the pendant keypad and touch `[OK]`
+
+### DX200
+
+1. boot the controller while holding `{MAIN MENU}` on the pendant keypad to enter *Maintenance* mode
+1. upgrade to *MANAGEMENT* security level by touching `[System Info]`→`[Security]` (default password is all `9`'s)
+1. touch `[System Info]`→`[Setup]` and select `OPTION FUNCTION`
+1. move to `NETWORK SETTING` and then to `HOST SETUP`
+1. make sure that `IP ADDRESS SETTING` is set to `MANUAL SETTING`. If it is not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
```
ted-miller, line 35:
```suggestion
1. make sure that `IP ADDRESS SETTING` is set to `MANUAL SETTING`.
If it is not, make sure that `DNS SETTING` and `SNTP SETTING` are not set to `DHCP SETTING` (change them if you need to), and then set `IP ADDRESS SETTING` to `MANUAL SETTING`
```

doc/network_configuration.md
ted-miller, line 42:
```suggestion
- Modify the `agent_ip_address` key in the `motoros2_config.yaml` file and specify an IP address that is on the robot's subnet.
See [here](../README.md#configuration-file) for information about the config file.
Then [propagate the changes to the Yaskawa controller](../README.md#updating-the-configuration).
You will need to ensure that the PC running the micro-ROS agent application uses this static IP address on the network port connected to the robot controller.
```

ted-miller: Note that this suggestion also fixes a typo, `informationa`

doc/network_configuration.md
```diff
@@ -0,0 +1,120 @@
...
+- Modify the robot controller's IP and subnet mask so it is on the subnet of the PC running the micro-ROS agent.
+- Modify the robot controller's network settings to add a gateway which can reach the IP address of the subnet of the PC running the micro-ROS agent.
+
+## Network issues
+
+If you are unable to connect the PC running the micro-ROS agent application and the controller, the network configuration may be the issue.
+Network problems can sometimes be the root of seemingly unrelated problems, such as rcl/rclc errors.
+
+To troubleshoot network issues, first ensure that the robot controller's IP address is `MANUAL SETTING`.
+It is important that it is [not using DHCP](#network-configuration).
```
ted-miller, line 52: I don't understand why `not using DHCP` is a link.

ted-miller (changes requested):

README.md
```diff
@@ -54,15 +54,18 @@ The following sections document how to download, install, configure, use and tro
   - [Downloading the files](#downloading-the-files)
   - [Extracting the files](#extracting-the-files)
   - [Verifying integrity of the download](#verifying-integrity-of-the-download)
-- [Configuration](#configuration)
-  - [Verifying YAML correctness](#verifying-yaml-correctness)
-- [Example INFORM jobs](#example-inform-jobs)
+- [Building from source](doc/build_from_source.md)
```
gavanderhoorn, line 57: it's a bit unclear from the diff, but does this re-order things such that building from source comes before installing the binaries provided by us?

If so: we should not encourage users to build from source, and in my experience, if that's listed first, that's what people assume they should do / attempt first.

gavanderhoorn: > It is immediately after the "download" section. 

True, but:

image

with people just following 'the steps in the readme' I would really recommend pushing *Building from source* below *Installation*.

jimmy-mcelwain: I moved the building from source section back below installation and added a bit more information about when/why to build from source vs just download binaries.

README.md
```diff
@@ -346,15 +288,92 @@ If `SAVE DATA CRC CHECK FUNC (FSU)` was disabled at the start of this procedure,
...
+| `two_arms`                  | Two robot groups     | DX200, YRC1000, YRC1000micro |
+| `sda_dual_arm`              | SDA robots *only*    | -                            |
+
+**These jobs are not required.**
+The INFORM job will be automatically generated at startup.
+These are only provided for convenience to show the required commands if a custom job is required.
+
+If needed, open a new issue on the [Issue tracker](https://github.com/yaskawa-global/motoros2/issues) to ask for support with creating a custom job.
+
+## Understanding the micro-ROS Agent
```
ted-miller, line 315: Shouldn't this be down around `Running the micro-ROS Agent`

README.md
```diff
@@ -346,15 +288,92 @@ If `SAVE DATA CRC CHECK FUNC (FSU)` was disabled at the start of this procedure,
...
+### Configuration file
+
+All MotoROS2 distribution `.zip` files come with a template `.yaml` file, which *must* be updated before deployment to a specific Yaskawa Motoman robot controller.
+
+For more information on all configuration settings, please refer to the `motoros2_config.yaml` file in the release `.zip`.
+
+All fields have defaults, except the IP address and UDP port number at which MotoROS2 expects the micro-ROS Agent to be reachable.
+These fields must be set to correct values by users, as otherwise MotoROS2 will not be able to communicate with the ROS 2 node graph.
+
+Edit the template `.yaml` and change the `agent_ip_address` and `agent_port_number` to point to the host which will run the micro-ROS Agent application (the PC, not the robot controller).
```
ted-miller, line 338: Maybe just say "see below" for information on `Understanding the micro-ROS Agent`

ted-miller: Something like "This will be the PC running ROS2."

README.md
jimmy-mcelwain: The reason that I put it here is because of https://github.com/Yaskawa-Global/motoros2/issues/260, where you said that there should be some explanation about the micro-ROS agent before the configuration section.

README.md
jimmy-mcelwain: Ok now the other comment that I just wrote is not as relevant. But I also wanted a description of the micro-ros agent before the network configuration section, because that terminology is used there a lot too.

ted-miller (approved):

PR #358, opened 2025-01-22
Simplify motion control

jimmy-mcelwain:
`MotionControl.c` had several instances of unused code, code that could have been replaced with a function call, and code that was otherwise duplicated. 

The only real functional change of this PR is that duplicate joint names are checked for all points in `queue_traj_point` mode now, not just for the first point. Everything else is a readability/code duplication removal change.

### DIFF
--- a/src/MotionControl.c
+++ b/src/MotionControl.c
@@ -31,6 +31,14 @@ Init_Trajectory_Status Ros_MotionControl_ConvertTrajectoryToJointMotionData(traj
 /// <returns>TRUE for success, FALSE for failure</returns>
 BOOL Ros_MotionControl_FindCtrlGroupAndIndex(rosidl_runtime_c__String jointName, int* group_idx, int* joint_idx);
 
+
+/// <summary>
+/// Go through a list of joint names and see if there are any duplicates. 
+/// </summary>
+/// <param name="jointNames">A pointer to the structure containing all of the request joint names</param>
+/// <returns>TRUE if there are duplicate names, FALSE if there are no duplicate names</returns>
+BOOL Ros_MotionControl_HasDuplicateNames(rosidl_runtime_c__String__Sequence* jointNames);
+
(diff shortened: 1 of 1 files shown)


src/MotionControl.c
```diff
@@ -21,6 +21,24 @@
 Init_Trajectory_Status Ros_MotionControl_ConvertTrajectoryToJointMotionData(trajectory_msgs__msg__JointTrajectoryPoint__Sequence* in_jointTrajData, 
     int incomingAxisIndex, CtrlGroup* ctrlGroup, int ctrlGroupAxisIndex, JointMotionData* out_jointMotionData);
 
+/// <summary>
+/// Given a joint name, search through each control group and axis to find
+/// the control group + axis number of the corresponding joint
+/// </summary>
+/// <param name="jointName">Pointer to the head of the incoming trajectory</param>
```
ted-miller, line 28: That argument is not actually a pointer. Should it be?

ted-miller (commented):
Just one pedantic issue

src/MotionControl.c
jimmy-mcelwain: It probably should be. I'll change that.

gavanderhoorn: We could change it, but [rosidl_runtime_c__String](https://docs.ros.org/en/iron/p/rosidl_runtime_c/generated/structrosidl__runtime__c____String.html) is essentially a wrapper around a pointer, grouped with its `size` and `capacity` fields.

That's probably 4 + 8 bytes on a YRC1. You'd save copying 8 bytes, which I would not expect to have a measurable performance impact.

jimmy-mcelwain: I chose to change it to a pointer anyways. I know that it likely won't have a real performance, but it's more consistent. Everywhere else, I am passing a pointer rather than the full struct.

ted-miller (approved):
Nice refactor. Looks good.

PR #364, opened 2025-02-03
Update changelog (0.2.0)

jimmy-mcelwain:
Everything in the `0.2.0` milestone is complete. I added a few PRs made since 0.1.3 that I thought were important enough to put into the changelog to the milestone. If anybody wants to add more stuff, delete what I have, or generally make edits to or give suggestions for the change log, feel free. 

Also, I would just like to mention 2 things that are slightly messy, but in my opinion acceptable, about the current state of this update. 

1. Right now, we are building against `micro_ros_motoplus` version `20250113` for `jazzy`, and version `20241211` for all other distributions. 
2. Since we're still at `0.2.0`, this may be okay, but with #341 (and more directly https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30) we are already guaranteeing that `0.2.1` will be another breaking release, even though this one is already breaking.

What do you think @gavanderhoorn @ted-miller

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,37 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # Changelog
 
+## Forthcoming
+
+**Added support for ROS2 Jazzy Jalisco!** ([#337](https://github.com/Yaskawa-Global/motoros2/pull/337))
+
+MotoROS2 is now built against `micro_ros_motoplus` version `20250207`.
+
+Breaking changes have been introduced in this version of MotoROS2, and failure to update and rebuild the workspace will lead to the inability to receive messages from MotoROS2 and/or crashes of client applications.
+
+**Breaking changes:**
+
+- Default joint names for single-group systems have been simplified (`joint_M`) ([#340](https://github.com/Yaskawa-Global/motoros2/pull/340))
+- Changes to `motoros2_interfaces` built into `micro_ros_motoplus` mean that you must update your client PC's ROS2 workspaces using `motoros2_client_interface_dependencies` version `0.1.1`. See the instructions [here](https://github.com/yaskawa-global/motoros2_client_interface_dependencies#overview).
+
+New functionality:
+
+- Allow users to insert USB with config file after bootup ([#299](https://github.com/Yaskawa-Global/motoros2/pull/299))
+- rcl and rclc errors now get forwarded to the user for simplified troubleshooting ([#306](https://github.com/Yaskawa-Global/motoros2/pull/306))
+- Users can enable/disable debug broadcast and choose debug broadcast port ([#309](https://github.com/Yaskawa-Global/motoros2/pull/309))
+
+Changes:
+
+- Error messages for `start_traj_mode` and `start_point_queue_mode` are more accurate and descriptive ([#297](https://github.com/Yaskawa-Global/motoros2/pull/297))
+- Fix response if `reset_error` service is called when major alarm is active ([#298](https://github.com/Yaskawa-Global/motoros2/pull/298))
+- Fixed condition so tool 63 can be selected properly ([#314](https://github.com/Yaskawa-Global/motoros2/pull/314))
+- The `queue_traj_point` service will give more detailed error messages if trajectory initialization fails ([#341](https://github.com/Yaskawa-Global/motoros2/pull/341))
+- Fixed FJT goal tolerance violation message ([#345](https://github.com/Yaskawa-Global/motoros2/pull/345))
+- Fix bug where `stop_traj_mode` would have response message indicating failure even after succeeding ([#352](https://github.com/Yaskawa-Global/motoros2/pull/352))
+- Always check for duplicate joint names in `queue_traj_point` ([#358](https://github.com/Yaskawa-Global/motoros2/pull/358))
+- General refactor of code that processes incoming trajectories ([#358](https://github.com/Yaskawa-Global/motoros2/pull/358))
+- The MAC address used for the default node name and key name now always matches the MAC address used for ROS traffic ([#365](https://github.com/Yaskawa-Global/motoros2/pull/365))
+
 ## 0.1.3 (2024-08-01)
 
 MotoROS2 is now built against `micro_ros_motoplus` version `20240710`.


ted-miller:
> Right now, we are building against micro_ros_motoplus version 20250113 for jazzy, and version 20241211 for all other distributions.

That's not an issue. They're the same version; just different build dates.

> Since we're still at 0.2.0, this may be okay, but with https://github.com/Yaskawa-Global/motoros2/pull/341 (and more directly https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30) we are already guaranteeing that 0.2.1 will be another breaking release, even though this one is already breaking

Ehhhh... I hate that we keep creeping more and more stuff in. But yeah, this issue bothers me. Let's resolve those so that the 020 release includes the latest mr2_interfaces.

CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
+New functionality:
+
+- Added support for Jazzy Jalisco ([#337](https://github.com/Yaskawa-Global/motoros2/pull/337))
+- rcl and rclc errors now get forwarded to the user for simplified troubleshooting ([#306](https://github.com/Yaskawa-Global/motoros2/pull/306))
+- Allow users to insert USB with config file after bootup ([#299](https://github.com/Yaskawa-Global/motoros2/pull/299))
+- Users can enable/disable debug broadcast and choose debug broadcast port ([#309](https://github.com/Yaskawa-Global/motoros2/pull/309))
+
+Changes:
+
+- Default joint names for single-group systems are simplified (`joint_M`) ([#340](https://github.com/Yaskawa-Global/motoros2/pull/340))
```
ted-miller, line 30: This can also be a breaking change and should be noted as such.

CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
+Users who are updating from a lower version must remove the `motoros2_interfaces` packages from their Colcon workspaces (from the `src`, `build` and `install` spaces), and follow the installation instructions in the [Build and installation](https://github.com/yaskawa-global/motoros2_client_interface_dependencies#build-and-installation) section in the `README` of the new `yaskawa-global/motoros2_client_interface_dependencies` package.
+Breaking changes have been introduced in this version of MotoROS2, and failure to update and rebuild the workspace will lead to the inability to receive messages from MotoROS2 and/or crashes of client applications.
+
+For Jazzy, MotoROS2 is now built against `micro_ros_motoplus` version `20250113`.
+
+For Humble, Galactic, and Foxy, MotoROS2 is now built against `micro_ros_motoplus` version `20241211`.
+
+New functionality:
+
+- Added support for Jazzy Jalisco ([#337](https://github.com/Yaskawa-Global/motoros2/pull/337))
```
ted-miller, line 23: Big. Bold. Excitement!

CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # Changelog
 
+## Forthcoming
+
+**Breaking change**:
+
+Users who are updating from a lower version must remove the `motoros2_interfaces` packages from their Colcon workspaces (from the `src`, `build` and `install` spaces), and follow the installation instructions in the [Build and installation](https://github.com/yaskawa-global/motoros2_client_interface_dependencies#build-and-installation) section in the `README` of the new `yaskawa-global/motoros2_client_interface_dependencies` package.
```
ted-miller, line 14: Should we do this?
https://github.com/Yaskawa-Global/motoros2_client_interface_dependencies/discussions/17

CHANGELOG.md
```diff
@@ -7,6 +7,34 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
+For Humble, Galactic, and Foxy, MotoROS2 is now built against `micro_ros_motoplus` version `20241211`.
+
+New functionality:
+
+- Added support for Jazzy Jalisco ([#337](https://github.com/Yaskawa-Global/motoros2/pull/337))
+- rcl and rclc errors now get forwarded to the user for simplified troubleshooting ([#306](https://github.com/Yaskawa-Global/motoros2/pull/306))
+- Allow users to insert USB with config file after bootup ([#299](https://github.com/Yaskawa-Global/motoros2/pull/299))
+- Users can enable/disable debug broadcast and choose debug broadcast port ([#309](https://github.com/Yaskawa-Global/motoros2/pull/309))
+
+Changes:
```
ted-miller, line 28: - General refactor of code that processes incoming trajectories

CHANGELOG.md
jimmy-mcelwain: It makes sense to me

jimmy-mcelwain:
> > Since we're still at 0.2.0, this may be okay, but with https://github.com/Yaskawa-Global/motoros2/pull/341 (and more directly https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30) we are already guaranteeing that 0.2.1 will be another breaking release, even though this one is already breaking

> Ehhhh... I hate that we keep creeping more and more stuff in. But yeah, this issue bothers me. Let's resolve those so that the 020 release includes the latest mr2_interfaces.

I opened a PR at https://github.com/Yaskawa-Global/motoros2_interfaces/pull/31 that will update the `motoros2_interfaces` version to `0.2.0` and then we can tag it as such and then build update the version in `micro_ros_motoplus` and rebuild `libmicroros` and have another release. 

I will also say that there are other potential breaking PRs open right now at `motoros2_interfaces`, so if those get merged anytime soon it may be that `0.2.1` is breaking regardless and what we're doing now doesn't have a big impact. That being said, I'm fine with including the recent changes from `motoros2_interfaces` in `0.2.0`.

jimmy-mcelwain:
@gavanderhoorn since we decided to add https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30 to `0.2.0`, we're going to need new `libmicroros` builds. Ted told me that you usually handle it, and I figured I would ping you to ask if you could build it. If you would rather not, I wouldn't mind doing it. As far as I know, the only things that you do differently than a normal debug build are: 
1. setting the flags the way they are set in the `build_info.yaml` for the other releases
2. fixing the version string

Is that right? 

Slightly related, but is it possible to use [motoplus_libmicroros_builder](https://github.com/Yaskawa-Global/motoplus_libmicroros_builder) and build with `extra_cflags` and `extra_cxxflags`? Those options don't appear when I try to run the workflow. But having that option would be nice, as my PC is much slower than the GitHub action.

ted-miller:
Also needs #365

gavanderhoorn:
@ted-miller wrote:
> @jimmy-mcelwain wrote:
> > Right now, we are building against micro_ros_motoplus version 20250113 for jazzy, and version 20241211 for all other distributions.
>
> That's not an issue. They're the same version; just different build dates.

content-wise, that was probably correct. However seeing as we're using calver, artefacts with different versions/stamp are really different releases.

@jimmy-mcelwain wrote:
> @gavanderhoorn since we decided to add [Yaskawa-Global/motoros2_interfaces#30](https://github.com/Yaskawa-Global/motoros2_interfaces/pull/30) to `0.2.0`, we're going to need new `libmicroros` builds. Ted told me that you usually handle it, and I figured I would ping you to ask if you could build it. 

All 12 variants of `libmicroros` version `20250207` are now available at [Yaskawa-Global/micro_ros_motoplus/releases](https://github.com/Yaskawa-Global/micro_ros_motoplus/releases?q=20250207&expanded=true).


> If you would rather not, I wouldn't mind doing it. As far as I know, the only things that you do differently than a normal debug build are:

There are a couple more steps actually. And it's not manual either, I have a couple of scripts that take care of the majority of the builds + creating the tags + releases.

> Slightly related, but is it possible to use [motoplus_libmicroros_builder](https://github.com/Yaskawa-Global/motoplus_libmicroros_builder) and build with `extra_cflags` and `extra_cxxflags`? Those options don't appear when I try to run the workflow. But having that option would be nice, as my PC is much slower than the GitHub action.

The online builder is not exactly the same as the off-line scripts right now. There are various reasons for this, one of which would be the fact Github workflows (still) don't support more than 10 arguments.

ted-miller:
#341 needs to be added as well

CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # Changelog
 
+## Forthcoming
+
+MotoROS2 is now built against `micro_ros_motoplus` version `20250207`.
```
ted-miller, line 12: On the client PC, this requires `motoros2_interfaces` version `0.2.0`.

CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
+
+Breaking changes have been introduced in this version of MotoROS2, and failure to update and rebuild the workspace will lead to the inability to receive messages from MotoROS2 and/or crashes of client applications.
+
+**Breaking changes:**
+
+- Default joint names for single-group systems have been simplified (`joint_M`) ([#340](https://github.com/Yaskawa-Global/motoros2/pull/340))
+- Users who are updating from a lower version must remove the `motoros2_interfaces` packages from their Colcon workspaces (from the `src`, `build` and `install` spaces), and follow the installation instructions in the [Build and installation](https://github.com/yaskawa-global/motoros2_client_interface_dependencies#build-and-installation) section in the `README` of the new `yaskawa-global/motoros2_client_interface_dependencies` package.
```
ted-miller, line 19: IMO, this should be moved above where you mention what version this is built against.

ted-miller: and explicitly mention the version of interfaces. (Of course, updating the target of this will help.
https://github.com/Yaskawa-Global/motoros2_client_interface_dependencies/blob/8fbe9eb276001981f18f958c583f24760b74e2e4/source_deps.repos#L2-L5

gavanderhoorn: I would perhaps suggest we don't put instructions like this in the changelog.

Building workspaces is not something that should be explained in a changelog. It would be enough to notice that version X of `motoros2_interfaces` is the only version compatible with this release of MotoROS2, and if users don't have that version, we should mention they need to make sure to build/install *that* version.

I would only include the link to the readme of `motoros2_client_interface_dependencies`, nothing more (no details about `build` and `install` spaces).

gavanderhoorn: Actually, we should perhaps not mention `motoros2_interfaces` here, other than note the version that's compatible with this version of MotoROS2.

If we're going to pin `motoros2_interfaces` in `motoros2_client_interface_dependencies`, it should be sufficient to state which version of `motoros2_client_interface_dependencies` is compatible with this version of MotoROS2.

CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
+
+New functionality:
+
+- **Added support for ROS2 Jazzy Jalisco!** ([#337](https://github.com/Yaskawa-Global/motoros2/pull/337))
```
gavanderhoorn, line 23: Suggestion: make this an item / section by itself. It's not really new functionality as the other items in this section are: it's compatibility with a new version of ROS 2.

As such, this seems more like an announcement. Moving it right where we now only have "MotoROS2 is now built against [..]" would give it the attention it deserves, and it can also not be overlooked (ie: when scanning the changelog for "when did MotoROS2 get Jazzy support?").

gavanderhoorn (changes requested):
High-level comment: please make sure to sort all PR references in ascending order (similar to the other changelogs of other releases).

CHANGELOG.md
jimmy-mcelwain: I just copied and pasted the instructions from one of the old changelog updates. I deleted the instructions in the most recent push though. 

We still don't have any sort of versioning scheme for `motoros2_client_interface_dependencies`, so for now what I pushed has the `motoros2_interfaces` version in it, but if we want to figure out the versioning scheme/tags for `motoros2_client_interface_dependencies` before finalizing the changelog that's fine by me.

ted-miller: > We still don't have any sort of versioning scheme for motoros2_client_interface_dependencies

How about we maintain parity and make the version tags match? MotoROS2 0.2.0 == client_dependencies 0.2.0

jimmy-mcelwain: That sounds good to me

gavanderhoorn: While I like it as it would make it clear which version to use, it would also prevent us from releasing updates to `motoros2_client_interface_dependencies` which are not breaking changes.

gavanderhoorn: @gavanderhoorn wrote:
> While I like it as it would make it clear which version to use, it would also prevent us from releasing updates to `motoros2_client_interface_dependencies` which are not breaking changes.

one work-around would be to release MotoROS2 in tandem with `motoros2_client_interfaces_dependencies`. That way we'd not be limited to just one release of `motoros2_client_interfaces_dependencies` per MotoROS2 release.

It might introduce quite a bit of churn on MotoROS2 though.

ted-miller: >  state the versions of MotoROS2 a particular release of motoros2_client_interfaces_dependencies is compatible with in the readme of the latter.

Jimmy and I like this idea. He's going to submit a PR to `motoros2_client_interfaces_dependencies` and then we can tag it and reference it here.

CHANGELOG.md
ted-miller: We will reference a specific tag for the client_dependencies

ted-miller (approved):

PR #365, opened 2025-02-11
Fix MAC address for name and key and simplify network settings

jimmy-mcelwain:
This is a fix for https://github.com/Yaskawa-Global/motoros2/issues/57

Now the default node name, and the key use the MAC address from the correct NIC rather than always using the MAC address from the first port. 

In doing so, I had to remove a lot of code duplication which ended up simplifying a lot of the stuff related to verifying network settings from the config file. 

When I removed the duplicate code, I also ended up removing alarm subcodes/assertions that were only used once, so I removed those from the `ErrorHandling.h` file as well. 

Removing the alarm subcodes had a (small) cascading effect so I had to change some of the other error numbers.


jimmy-mcelwain:
I realize that many people probably just look at the `main` branch when looking at the `troubleshooting.md` file, not the documentation from the specific commit that they are using. Does that mean that we should avoiding changing the numbers of the error subcodes, or is it acceptable to change the error subcode numbers rather than just adding new ones to the end?

gavanderhoorn:
I haven't checked the changes in detail yet, but:

> Does that mean that we should avoiding changing the numbers of the error subcodes

Indeed, we can't ever reuse errors or subcodes.

That's not even just because of documentation, but also for things like support requests and data 'from the past'. It would become impossible to deterministically interpret reported errors if we'd change things around.

gavanderhoorn:
High-level comment:

> In doing so, I had to remove a lot of code duplication which ended up simplifying a lot of the stuff related to verifying network settings from the config file.
> 
> When I removed the duplicate code, I also ended up removing alarm subcodes/assertions that were only used once, so I removed those from the ErrorHandling.h file as well.
> 
> Removing the alarm subcodes had a (small) cascading effect so I had to change some of the other error numbers.

for future PRs, please try to use multiple commits for changes as extensive as this.

Facilitates review but also makes it easier to revert/drop commits if needed.

src/ConfigFile.c
```diff
@@ -549,29 +525,54 @@ void Ros_ConfigFile_ValidateCriticalSettings()
...
+    status = Ros_GetMacAddress(g_Ros_Controller.rosTrafficLanPort, g_Ros_Controller.rosTrafficMacAddr);
+    if (status != OK)
+    {
+        Ros_Debug_BroadcastMsg("%s: Ros_GetMacAddress: iface: %d; error: %d",
+            __func__, g_Ros_Controller.rosTrafficLanPort, status);
+        motoRosAssert_withMsg(FALSE, SUBCODE_CONFIGURATION_FAIL_MP_NICDATA,
+            "Must enable ETHERNET function");
+    }
+
+    if (strlen(g_nodeConfigSettings.node_name) == 0)
```
ted-miller, line 566: Since you have a defined constant for `PLACEHOLDER_NODE_NAME`, it's probably better to explicitly compare to that.

src/ErrorHandling.h
ted-miller, line 1: I agree with gavanderhoorn that the alarm codes must not change.

If a previously used code is not needed anymore, just add a suffix to the enum like `_OBSOLETE_DO_NOT_USE`.

ted-miller (changes requested):

src/ErrorHandling.h
jimmy-mcelwain: I added a suffix and a comment to the newly obsoleted codes

jimmy-mcelwain:
I split it up into more commits, put the error codes where they were before, and marked the obsolete error codes. Any out-of-place whitespace changes are just from auto formatting upon copy/pasting. Visual Studio doesn't show me the whitespace changes in the diff.

ted-miller (approved):

PR #371, opened 2025-02-21
Include missing libmicroros references

jimmy-mcelwain:
I was preparing for the 0.2.0 release, and I took a look at the binaries from the batch build for DX200 and YRC1000 and realized that they were far smaller than they should have been and were missing necessary code. I figured out that it was because `libmicroros.dnLib_jazzy` and `libmicroros.yrcmLib_jazzy` were missing. They have been building properly, but wouldn't have run.

### DIFF
--- a/src/MotoROS2_AllControllers.vcxproj.filters
+++ b/src/MotoROS2_AllControllers.vcxproj.filters
@@ -268,8 +268,14 @@
     <None Include="..\libmicroros_yrc1000_jazzy\libmicroros.yrcLib_jazzy">
       <Filter>MotoPlus Libraries\micro-ROS</Filter>
     </None>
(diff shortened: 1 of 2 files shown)


src/MotoROS2_AllControllers.vcxproj.filters
```diff
@@ -268,9 +268,15 @@
     <None Include="..\libmicroros_yrc1000_jazzy\libmicroros.yrcLib_jazzy">
       <Filter>MotoPlus Libraries\micro-ROS</Filter>
     </None>
-	<None Include="..\doc\network_configuration.md">
+    <None Include="..\doc\network_configuration.md">
```
ted-miller, line 271: whitespace is auto-formatted by VS

ted-miller (approved):

PR #372, opened 2025-02-21
Add Jazzy fix for DX100 and YRC1000u to changelog

jimmy-mcelwain:


ted-miller (approved):

gavanderhoorn:
For future releases: such a fixup PR / commit doesn't need to be mentioned in the changelog necessarily.

There is no real change here that wouldn't have been part of the original #337, and it just makes the changelog more 'noisy'.

gavanderhoorn:
re: Add Jazzy fix for DX100 and YRC1000u to changelog: this should be DX200, correct?

ted-miller:
Heh. Jimmy actually made that argument too. I made him do the pr because I thought it was a pretty big bug.

And yes, dx200

PR #398, opened 2025-03-27
Fix debug log statement

jimmy-mcelwain:
It previously always assumed that there would be 6 axes. Now it makes no assumptions about the number of axes.

This is a lot of code for such a small fix. I would be fine with changing it to not print out the nice formatting (i.e. commas, parentheses, brackets) that require more code/calls to `snprintf`. Or we could just omit the position altogether and print only the timestamp and group number, that should be sufficient to identify the point sent, but it may make debugging harder. Or we could just go with what I have submitted here. I don't have a strong opinion. I'm happy to make changes.

### DIFF
--- a/src/CtrlGroup.c
+++ b/src/CtrlGroup.c
@@ -186,17 +186,17 @@ CtrlGroup* Ros_CtrlGroup_Create(int groupIndex, BOOL bIsLastGrpToInit, float int
         }
         Ros_Debug_BroadcastMsg(startupMessage);
 
-        Ros_Debug_BroadcastMsg("maxInc[%d] (in motoman joint order): %d, %d, %d, %d, %d, %d, %d",
+        Ros_Debug_BroadcastMsg("maxInc[%d] (in motoman joint order): %d, %d, %d, %d, %d, %d, %d, %d",
             groupIndex,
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
While this is the correct approach, for the sake of simplicity, we could also just decide to always print `MP_GRP_AXES_NUM`, as that's what the size of `trajectoryIterator->pos` is ([here](https://github.com/Yaskawa-Global/motoros2/blob/48d71146fb055c6a71d037abb0014d2f3cfa2031/src/CtrlGroup.h#L48)), regardless of how many axes the group actually has.

We do that elsewhere, and it would just result in some zeroes in the output.

ted-miller:
> we could also just decide to always print MP_GRP_AXES_NUM, as that's what the size of trajectoryIterator->pos is

I like that idea

jimmy-mcelwain:
Yeah that makes a lot more sense. For some reason I was thinking that the array was dynamically sized so we couldn't index points past the "end". I will change it to the simpler `MP_GRP_AXES_NUM` point version.

ted-miller (approved):
Looks good to me

gavanderhoorn (approved):
this looks OK.

Have you tested it on real hw?

ted-miller:
> Have you tested it on real hw?

Yes.

PR #411, opened 2025-04-07
Time utils fix

### LINKED ISSUE #402: Bad floating point cast to integer in time conversion utils
To convert from nanoseconds to milliseconds, which are both stored as integer types, we are multiplying by `0.000001` and then casting to an integer. But this means that they are losing precision and then getting inappropriately truncated. For example, `0.2 s`, aka `200000000 ns` gets truncated to `199 ms`, `0.3 s` gets truncated to `299` ms, etc. 

We should just need to change it to integer division instead of floating point multiplication. 

The problem is in at least two spots

https://github.com/Yaskawa-Global/motoros2/blob/48d71146fb055c6a71d037abb0014d2f3cfa2031/src/TimeConversionUtils.h#L15-L18

https://github.com/Yaskawa-Global/motoros2/blob/48d71146fb055c6a71d037abb0014d2f3cfa2031/src/TimeConversionUtils.h#L37-L40

This may only matter if the duration or timestamp falls right on a whole millisecond value.

jimmy-mcelwain:
Resolves #402

This PR gets rid of the floating point math error that results in inappropriate truncation of ms timestamps. 

It also changes the time utils to work for negative times and negative durations. 

I added a bunch of unit tests. Note that due the problem discussed in #407, the controller will crash if you set the testing flag. So make sure to comment out [these lines](https://github.com/Yaskawa-Global/motoros2/blob/ec13454105c21c9397449f9aa1491a224147cc14/src/main.c#L75-L78) before setting that flag and building. 

This PR is a lot more broad than I was originally planning on it being, mostly because dealing with negative times/durations is more inconvenient than I originally thought. I'm happy to make changes though.

### DIFF
--- a/src/Tests_TestUtils.c
+++ b/src/Tests_TestUtils.c
@@ -30,4 +30,32 @@ BOOL Ros_Testing_CompareLong(long a, long b)
     return ok;
 }
 
+BOOL Ros_Testing_INT64_Equals(INT64 a, INT64 b)
+{
+    BOOL ok = a == b;
+
+    if (!ok)
+        Ros_Debug_BroadcastMsg("Fail: %lld != %lld", a, b);
+
+    return ok;
+}
+
+BOOL Ros_Testing_Timespec_Equals(const struct timespec* lhs, const struct timespec* rhs)
+{
+    BOOL ok = lhs->tv_sec == rhs->tv_sec;
+
+    if (!ok)
+    {
+        Ros_Debug_BroadcastMsg("Fail timespec seconds: %d != %d", lhs->tv_sec, rhs->tv_sec);
+    }
+
+    ok &= lhs->tv_nsec == rhs->tv_nsec;
+
+    if (!ok)
+    {
+        Ros_Debug_BroadcastMsg("Fail timespec nanos: %d != %d", lhs->tv_nsec, rhs->tv_nsec);
+    }
+    return ok;
+}
+
 #endif //MOTOROS2_TESTING_ENABLE
--- a/src/Tests_TimeConversionUtils.h
+++ b/src/Tests_TimeConversionUtils.h
@@ -1,0 +1,17 @@
+// Tests_TimeConversionUtils.h
+
+// SPDX-FileCopyrightText: 2025, Yaskawa America, Inc.
+// SPDX-FileCopyrightText: 2025, Delft University of Technology
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef MOTOROS2_TESTS_TIME_CONVERSION_UTILS_H
+#define MOTOROS2_TESTS_TIME_CONVERSION_UTILS_H
+
+#ifdef MOTOROS2_TESTING_ENABLE
+
+extern BOOL Ros_Testing_TimeConversionUtils();
+
+#endif //MOTOROS2_TESTING_ENABLE
+
+#endif  // MOTOROS2_TESTS_TIME_CONVERSION_UTILS_H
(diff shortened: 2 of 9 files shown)


src/Tests_TestUtils.c
```diff
@@ -30,4 +30,32 @@ BOOL Ros_Testing_CompareLong(long a, long b)
     return ok;
 }
 
+BOOL Ros_Testing_CompareINT64(INT64 a, INT64 b)
```
gavanderhoorn, line 33: Should we rename this to indicate it actually checks for *equality*?

I realise the existing functions follow a similar naming scheme, but we can take this opportunity to introduce more aptly named functions already (and address renaming the others at a later time).

src/Tests_TestUtils.c
```diff
@@ -30,4 +30,32 @@ BOOL Ros_Testing_CompareLong(long a, long b)
...
+{
+    BOOL ok = a == b;
+
+    if (!ok)
+        Ros_Debug_BroadcastMsg("Fail: %lld != %lld", a, b);
+
+    return ok;
+}
+
+BOOL Ros_Testing_CompareTimespec(const struct timespec* lhs, const struct timespec* rhs)
```
gavanderhoorn, line 43: Same here.

src/Tests_TimeConversionUtils.h
```diff
@@ -0,0 +1,17 @@
+// Tests_TimeConversionUtils.h
+
+// SPDX-FileCopyrightText: 2023, Yaskawa America, Inc.
+// SPDX-FileCopyrightText: 2023, Delft University of Technology
```
gavanderhoorn, line 4:
```suggestion
// SPDX-FileCopyrightText: 2025, Yaskawa America, Inc.
// SPDX-FileCopyrightText: 2025, Delft University of Technology
```

src/TimeConversionUtils.h
```diff
@@ -11,56 +11,76 @@
...
+        *quotient -=  1;
+    }
+
+    *remainder = abs(dividend - *quotient * divisor);
+}
 
 static inline INT64 Ros_Duration_Msg_To_Millis(builtin_interfaces__msg__Duration const* const x)
 {
-    return ((INT64)x->sec * 1000) + (INT64)(x->nanosec * 0.000001);
+    return (x->sec * 1000LL) + (x->nanosec / (1000LL * 1000LL));
```
gavanderhoorn, line 28: I like the fact you used two `1000LL` (instead of `1000000LL`). Would `1e6` also work?

Perhaps more a comment below, where it would become `1e9`.

jimmy-mcelwain: I thought about that, but thought I would just keep everything that I could in the integer space. I figured with the casting that would be needed to avoid floating point math, it wouldn't be much simpler that way but I recognize the way it is now is messy. I could switch to scientific notation with casts before any math happens, I think that should be fine.

jimmy-mcelwain:
The first round of suggestions have all been addressed

ted-miller:
Looks good to me. But I haven't tried to run it yet. I'm assuming that #412 should be done first.

jimmy-mcelwain:
> Looks good to me. But I haven't tried to run it yet. I'm assuming that https://github.com/Yaskawa-Global/motoros2/pull/412 should be done first.

It would probably make more sense that way

ted-miller (approved):
> Looks good to me. But I haven't tried to run it yet. I'm assuming that https://github.com/Yaskawa-Global/motoros2/pull/412 should be done first.

I just validated by commenting out the other test-routines.

gavanderhoorn:
Thanks for the changes @jimmy-mcelwain :+1:

PR #412, opened 2025-04-08
Fix unit tests

### LINKED ISSUE #408: `Tests_ActionServer_FJT` incorrectly initialises `JointTrajectoryPoint` fields
Context: #407.

Example:

https://github.com/Yaskawa-Global/motoros2/blob/48d71146fb055c6a71d037abb0014d2f3cfa2031/src/Tests_ActionServer_FJT.c#L476-L478

`trajectory_msgs__msg__JointTrajectoryPoint__init(..)` initialises all sequence members with a `capacity` and `size` of `0`.

Without resizing the sequence, the line `traj_point.positions.data[0] = 0.0` indexes into uninitialised memory.

jimmy-mcelwain:
Among other things, resolves #408

This fixes a bunch of problems with unit tests. Most of the problems were from memory management. 

There was a lot of double allocation or allocating and failing to free memory. 

The reason that it was crashing with the `MOTOROS2_TESTING_ENABLED` flag set (see #407) is because of a stack overflow error. The `CtrlGroup` struct takes up `39392 bytes`, but the default stack size for a thread created via `mpCreateTask(...)` is only `40KB`, so that was getting filled whenever a CtrlGroup was created on the stack. 

With the tests fixed, I also enabled the `Tests_ControllerStatusIO.c` (assuming that `MOTOROS2_TESTING_ENABLED` is set). I am also printing out the dynamic memory lost during testing (should always be 0).

### DIFF
--- a/src/CtrlGroup.c
+++ b/src/CtrlGroup.c
@@ -55,7 +55,9 @@ const char* Ros_CtrlGroup_GRP_ID_String[] =
 
 CtrlGroup* Ros_CtrlGroup_Ctor()
 {
-    return (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
+    CtrlGroup* grp = (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
+    bzero(grp, sizeof(CtrlGroup));
+    return grp;
 }
 
 void Ros_CtrlGroup_Dtor(CtrlGroup* grp)
--- a/src/CtrlGroup.c
+++ b/src/CtrlGroup.c
@@ -53,6 +53,16 @@ const char* Ros_CtrlGroup_GRP_ID_String[] =
     "s24",
 };
 
+CtrlGroup* Ros_CtrlGroup_Ctor()
+{
+    return (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
+}
+
+void Ros_CtrlGroup_Dtor(CtrlGroup* grp)
+{
+    mpFree(grp);
+}
+
 //-------------------------------------------------------------------
 // Create a CtrlGroup data structure for existing group otherwise
 // return NULL
--- a/src/Tests_ActionServer_FJT.c
+++ b/src/Tests_ActionServer_FJT.c
@@ -56,6 +56,9 @@ static BOOL Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_empty_jtole
     const size_t NUM_JOINTS = 6;
     double posTolerances[NUM_JOINTS];
 
+    rosidl_runtime_c__String__Sequence__init(&joint_names, NUM_JOINTS);
+    joint_names.size = 0;
+
     //init capacity to NUM_JOINTS, but don't add any elements
     control_msgs__msg__JointTolerance__Sequence__init(&joint_tolerances, NUM_JOINTS);
     joint_tolerances.size = 0;
@@ -92,7 +95,6 @@ static BOOL Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_out_array_t
     BOOL bSuccess = TRUE;
 
     STATUS status = 0;
-    control_msgs__msg__JointTolerance joint_tolerance;
     control_msgs__msg__JointTolerance__Sequence joint_tolerances;
     rosidl_runtime_c__String__Sequence joint_names;
     const size_t NUM_JOINTS = 6;
@@ -111,9 +113,6 @@ static BOOL Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_out_array_t
     joint_tolerances.size = 0;
 
     //add (at least) a single JointTolerance (don't need to init it, just add it)
-    bzero(&joint_tolerance, sizeof(joint_tolerance));
-    control_msgs__msg__JointTolerance__init(&joint_tolerance);
-    joint_tolerances.data[0] = joint_tolerance;
     joint_tolerances.size = 1;
 
     //call
@@ -139,7 +138,6 @@ static BOOL Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_more_jtol_t
     BOOL bSuccess = TRUE;
 
     STATUS status = 0;
-    control_msgs__msg__JointTolerance joint_tolerance;
     control_msgs__msg__JointTolerance__Sequence joint_tolerances;
     rosidl_runtime_c__String__Sequence joint_names;
     const size_t NUM_JOINTS = 6;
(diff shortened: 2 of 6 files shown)


gavanderhoorn:
High-level comment: thanks for the fixes. Some seem like there were some issues there indeed.

src/Tests_ActionServer_FJT.c
```diff
@@ -347,19 +328,20 @@ static BOOL Ros_Testing_Ros_ActionServer_FJT_Parse_GoalPosTolerances_last_settin
     //check joint tols
 
     //had no JointTolerance instance, so should be default
-    BOOL bT01 = posTolerances[0] == DEFAULT_FJT_GOAL_POSITION_TOLERANCE;
+    BOOL bT01 = Ros_Testing_CompareDouble(posTolerances[0], DEFAULT_FJT_GOAL_POSITION_TOLERANCE);
```
gavanderhoorn, line 331: I can see why you made this change, but I don't believe it's necessary in this case: the function-under-test (FUT) is supposed to only re-arrange values, not perform any calculations. So the `DEFAULT_FJT_GOAL_POSITION_TOLERANCE` is more of a *sentinel value* here, rather than the result of a calculation.

The test is therefore not to see whether `posTolerances[0]` is an expected value (or: falls in some range) after processing by the FUT, but to see whether that exact value is returned. It should not be any other value, whatsoever. If it is, something is wrong.

Same comment for the other values and comparisons below.

src/Tests_ControllerStatusIO.c
```diff
@@ -85,13 +83,12 @@ BOOL Ros_Testing_ControllerStatusIO_ShouldWarnNoCalibDataLoaded_R1()
     BOOL bSuccess = TRUE;
 
     Controller controller;
-    CtrlGroup grp0;
 
+    CtrlGroup* grp0 = (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
     controller.numGroup = 1;
-    controller.numRobot = 1;
```
gavanderhoorn, line 91: is this initialised by something else? I don't believe `Ros_Testing_ControllerStatusIO_MakeFake6dofRobot(..)` touches this, does it?

jimmy-mcelwain: That property has been commented out. 

https://github.com/Yaskawa-Global/motoros2/blob/316b9993b5d1f1be940ffd46147934261726e596/src/ControllerStatusIO.h#L74

src/Tests_ActionServer_FJT.c
jimmy-mcelwain: That makes sense, I'll revert that

src/Tests_ControllerStatusIO.c
```diff
@@ -85,13 +83,12 @@ BOOL Ros_Testing_ControllerStatusIO_ShouldWarnNoCalibDataLoaded_R1()
     BOOL bSuccess = TRUE;
 
     Controller controller;
-    CtrlGroup grp0;
 
+    CtrlGroup* grp0 = (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
```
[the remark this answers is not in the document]
jimmy-mcelwain: Would this theoretically be to create multiple control groups/free multiple groups at once? I think I prefer the way it is right now, but I wouldn't be against this.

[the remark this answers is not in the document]
gavanderhoorn: No, I was just thinking of something like:

```c
CtrlGroup* Ros_CtrlGroup_Ctor()
{
  return (CtrlGroup*) mpMalloc(sizeof(CtrlGroup));
}
```

and:

```c
void Ros_CtrlGroup_Ctor(CtrlGroup* grp)
{
  mpFree(grp);
}
```

We already have a `_Create(..)`, but that does both creation *and* initialisation.

Creating these two functions would allow us to refactor things a bit more easily in the future (ie: zero-initialise some fields in the ctor, or set them to sane defaults, etc).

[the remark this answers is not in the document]
jimmy-mcelwain: I added those and use them in the unit tests now

src/Tests_CtrlGroup.c
```diff
@@ -237,30 +237,34 @@ BOOL Ros_Testing_CtrlGroup_PosConverters()
...
+    Ros_CtrlGroup_Dtor(group);
+
     return bAllTestsPassed;
 }
 
 BOOL Ros_Testing_CtrlGroup_HasBaseTrack()
 {
-    CtrlGroup group;
+    CtrlGroup* group = Ros_CtrlGroup_Ctor();
+    bzero(group, sizeof(CtrlGroup));
```
gavanderhoorn, line 248: Sorry for only suggesting this now, but we could move this to the `Ctor()` you just introduced.

jimmy-mcelwain: No problem, just pushed it

src/CtrlGroup.c
```diff
@@ -53,6 +53,18 @@ const char* Ros_CtrlGroup_GRP_ID_String[] =
     "s24",
 };
 
+CtrlGroup* Ros_CtrlGroup_Ctor()
```
ted-miller, line 56: Shouldn't this be used in `Ros_CtrlGroup_Create`?

src/CtrlGroup.c
```diff
@@ -53,6 +53,18 @@ const char* Ros_CtrlGroup_GRP_ID_String[] =
...
+{
+    CtrlGroup* grp = (CtrlGroup*)mpMalloc(sizeof(CtrlGroup));
+    bzero(grp, sizeof(CtrlGroup));
+    return grp;
+}
+
+void Ros_CtrlGroup_Dtor(CtrlGroup* grp)
```
ted-miller, line 63: Should this get merged with `Ros_CtrlGrp_Cleanup` or called in `Ros_Controller_Cleanup`

gavanderhoorn: The former does some [rather specific things](https://github.com/Yaskawa-Global/motoros2/blob/316b9993b5d1f1be940ffd46147934261726e596/src/CtrlGroup.c#L236-L242) and doesn't release the memory of the `CtrlGroup` passed in.

We should indeed start using this in `Ros_Controller_Cleanup(..)` [here](https://github.com/Yaskawa-Global/motoros2/blob/316b9993b5d1f1be940ffd46147934261726e596/src/ControllerStatusIO.c#L246). But I would again suggest that's not in scope in this PR.

ted-miller (approved):
Looks good

gavanderhoorn:
Thanks for fixing this @jimmy-mcelwain :+1:

PR #415, opened 2025-04-14
Fixing FJT feedback message

jimmy-mcelwain:
This has ended up being less straightforward than I was expecting, so I'm putting this out there as a draft and will make any requested modifications, clean everything up, and break it into different commits before finalizing the PR. 

First, I realized that the way that we handle FSU/accept that slowdowns can happen will make this difficult. When populating the `desired` feedback, it is difficult to choose what point to use. Specifically for a trajectory that "falls behind", do we want desired to be the most recently commanded position, or what position "should be commanded" at this time (if there was no slowdown). I would like to do the latter, but unfortunately if there is a slowdown, the increment point queue can quickly fill up and the position that "should be commanded" may not have been calculated yet. The increment queue has a max length of 200 points, or less than a second in time so that will likely happen if there is an FSU. We could increase that queue size, but the problem remains. I don't think that we could make the queue "big enough" to not worry. Plus if we do what "should" be commanded, that may make the code more complicated since we would have to traverse the increment queue separately. So I am reporting what currently is commanded as `desired`, instead, even if it doesn't match up perfectly with the request. This should only really matter if there is an FSU.

Second, desired and actual feedback messages are populated in two separate threads, so the `time_from_start` in reality going to be different between the two. So error can't be "properly" calculated. Right now I am just subtracting the most recent actual from desired, and the time differences are small enough that I don't feel it really matters, plus there's not much that can be done. If anybody would prefer a different method for populating `time_from_start`, let me know. 

Third, for `desired` `time_from_start`, at the end of a motion while waiting for the feedback to catch up, do we want to continue incrementing the `time_from_start` in the message? Or just leave it? E.g. say the `time_from_start` of the last point in the FJT request is `1 sec`, but the trajectory takes 1.3 seconds to complete. At `1.2 sec`, do we want desired to say `time_from_start` is `1.2 sec`, or `1 sec`? 

Fourth, which values (`position`, `velocity`, `acceleration`, `effort`) do we want to be populated for each of the point types (`actual`, `desired`, `error`)? For now, I was planning on having `position` and `velocity` for all three, skipping `acceleration` (like we do for `joint_states`), and only having `effort` for actual, because calculating the `desired` effort is difficult and we don't respect the effort submitted by the user in the FJT.

### DIFF
--- a/src/ActionServer_FJT.c
+++ b/src/ActionServer_FJT.c
@@ -331,10 +331,7 @@ void Ros_ActionServer_FJT_ResetProgressTracker()
     bzero(feedback->desired.velocities.data,
         sizeof(double) * g_messages_PositionMonitor.jointStateAllGroups->velocity.size);
 
-    //--------------------
-    //desired effort
-    bzero(feedback->desired.effort.data,
-        sizeof(double) * g_messages_PositionMonitor.jointStateAllGroups->effort.size);
+    feedback->desired.velocities.size = g_messages_PositionMonitor.jointStateAllGroups->velocity.size;
 
     //===========================================
     //error in position (zero)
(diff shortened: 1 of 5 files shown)


gavanderhoorn:
> When populating the `desired` feedback, it is difficult to choose what point to use. 

Yes, so perhaps we should focus first on mapping the terminology used in the feedback message on the state(s) available to us in MotoROS2 / M+.

Without doing that immediately, intuitively I'm wondering whether we could use whatever values have already been submitted via the M+ motion API. If there are getters for current feedback position, current commanded position, etc, those could be candidates.

We'd avoid having to deal with our own queue, and in the end, it would seem to be the most 'honest', as that's really what the robot is moving towards.

jimmy-mcelwain:
Yes there are getters for the current feedback position and commanded position. With this PR, `actual` is simply set to the most recent feedback position, and I am setting `desired` to the most recently read command position + the increment that was just submitted, because `mpGetPulsePos(...)`, which returns the command position, does not immediately reflect the newest increment that was submitted. So I am pretty much doing what you said above, but I'm just making sure that the `desired` also includes the most recent increment move.

PR #427, opened 2025-06-24
Experiment with real-time interface

ted-miller:
Not an actual PR. Do not merge. This is just to review the concept.

wrt #186 

I implemented something very similar to [tinglest's approach](https://github.com/ros-industrial/motoman/discussions/465#discussioncomment-2142603).

The idea is that this is just another "motion mode". So there's a new service to activate `start_rt_mode`. 

Previously, the `MP_PRI_IP_CLK_TAKE` task was started during init and never stopped. Now we would have each motion mode activate their desired version of it.

This new version of `mpExRcsIncrementMove` thread will use a direct TCP connection from the client. The blocking nature of the socket, combined with `mpClkAnnounce(MP_INTERPOLATION_CLK)`, will keep the PC synchronized with the robot's clock.

In order to maintain the deterministicness (deterministicality?) of this loop, I don't think we'll be able to do too much validation or conversion of the incoming stream. So I think we'll have to "trust" the incoming data more than we currently do.

For the FSU speed limit, I'm not sure if we should "buffer" some cycles and delay the next `mpRecv`. Since the whole point of this interface is to minimize buffering and latency, I'm leaning toward just letting the increments get truncated. We can always notify the user in the response packet of how much was lost. Then they can decide how to react.

Any thoughts?


jimmy-mcelwain:
I will look at this more tomorrow and give some feedback, but I have a very rough draft of a real time servo implementation that I made a few months ago. I just now pushed it to [my fork](https://github.com/jimmy-mcelwain/motoros2/tree/RealTimeServoing). Mine uses ROS2 messages instead, which it seems like users in #186 did not want.

Issue #11, opened 2023-05-30
Certain group cfgs make JointState output stop working


ted-miller:
On an `R1+R2+S1` system, the `/joint_states` topic does not publish. 

I did check the size of the joint_states message with `R1+R2+S1`. It was somewhere around 1100 bytes. So I don't think the size of the message is a problem.

Tested `R1 S1 S2 S3` for other reasons. It worked ok. So it's not the number of groups. But it's still possible it's the number of axes.

I have not checked yet to verify if the publish API is returning an error code.

ted-miller:
Additionally, this issue occurs on Foxy, Galactic, and Humble. The ROS2 version is not the reason for this issue.

ted-miller:
I restored an old CMOS.bin for other reasons.

When trying to run my FJT script, it says that I didn't have velocity data for all axes. Well... scrolling up through the log, I see that all of the joint names are blank. 

So now I'm thinking that the JointStates publisher is broken because of the lack of joint-names.
[CMOS.zip](https://github.com/Yaskawa-Global/motoros2/files/12768240/CMOS.zip)

I will continue debug next week.

gavanderhoorn:
> When trying to run my FJT script, it says that I didn't have velocity data for all axes. Well... scrolling up through the log, I see that all of the joint names are blank.

if you don't configure custom ones, you should get the default ones generated:

https://github.com/Yaskawa-Global/motoros2/blob/6c5614b64f27adc40bff1958b63e1467baeffeae/src/ControllerStatusIO.c#L77-L102

ted-miller:
Yeah... I just now saw that too. 
So now I don't know what's up and my system is extra broken.

gavanderhoorn:
Would you have a sample of the `/joint_states` topic?

ted-miller:
/joint_states won't publish. `topic echo` just freezes until I `ctrl-c`. 

However, I am able to view the topics for each individual control-group, such as `/ctrl_groups/r1/joint_states`. That works fine.

gavanderhoorn:
Hm, wireshark? Any XRCE-DDS traffic?

ted-miller:
@jimmy-mcelwain I just had another thought. Rather than a max number of bytes, the uros allocator may be limiting the size of the sequence (array of joint names). 

In the FJT action, we're explicitly defining some allocation rules to that regard.

[...]

PR #430, opened 2025-07-03
Update outdated item from known issues

jimmy-mcelwain:
I believe that this section in the known issues was referencing #11, but that's been solved with https://github.com/Yaskawa-Global/micro_ros_motoplus_buildscripts/pull/163

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ The following sections document how to download, install, configure, use and tro
   - [No support for partial goals](#no-support-for-partial-goals)
   - [Upper limit to publishing frequency](#upper-limit-to-publishing-frequency)
   - [Incorrect transform tree origin with multi-robot setups](#incorrect-transform-tree-origin-with-multi-robot-setups)
+  - [Certain topics won't publish data](#certain-topics-wont-publish-data)
 - [Provisional roadmap](#provisional-roadmap)
 - [Frequently Asked Questions](doc/faq.md)
 - [Troubleshooting](doc/troubleshooting.md)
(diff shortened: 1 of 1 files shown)


jimmy-mcelwain:
I think that it is unlikely that users will run into this problem anymore, though it is possible. If you would prefer, I could amend this PR to instead clarify the actual cause now that we know it, and how it can be fixed on the user's side (reducing the length of joint names in the config file).

ted-miller (approved):
Looks good to me

README.md
```diff
@@ -818,16 +819,25 @@ Refer to the relevant Yaskawa Motoman documentation for more information on how
 
 After robot-calibration, the transform between the shared `world` frame and each robot's `base` frame will be known, and MotoROS2 will include it in the transforms it broadcasts.
 
-### Some group combinations won't publish data
+### Best-effort reliability topics fail to publish large messages
```
gavanderhoorn, line 822: I'd suggest using a heading that only describes the symptom, not a/the cause.

Users are more likely to search for symptoms they've observed.

The description below would also need to be updated to match -- or perhaps we keep the version that @ted-miller initially wrote.

The `cause of this behaviour is unknown` would be where this would go into details and link the issue that's linked in the *description* now.

README.md
```diff
@@ -818,16 +819,25 @@ Refer to the relevant Yaskawa Motoman documentation for more information on how
...
-All other topics and services work as expected.
+**Description**: Topics with QoS set to best-effort reliability will fail to publish large messages.
+This behavior is most commonly observed with `/joint_states`.
+As an example, if your system has a large number of joints or long joint names which result in a large message, `/joint_states` will not be published.
+This can occur with other best-effort reliability topics as well.
+The root cause of this behavior is described [here](https://github.com/eProsima/Micro-XRCE-DDS-Client/issues/394#issuecomment-2519526357).
 
-**Work-around**: such a multi-group system would need to be broken up into a independent systems.
-(E.g. `R1+R2+S1` would be broken up into `R1+S1` and another `R1`).
+**Work-arounds**: There are two options for fixing this behavior.
```
gavanderhoorn, line 830: are they *work-arounds* or *fixes* ? :)

they can't be both :)

README.md
```diff
@@ -818,16 +819,25 @@ Refer to the relevant Yaskawa Motoman documentation for more information on how
...
 
-The cause of this behavior is unknown.
-The issue is being investigated.
+1. Reduce the length of your messages.
+For example, you can reduce the length of the `/joint_states` message by modifying `joint_names` in [the configuration file](#configuration) to reduce the length of joint names.
```
gavanderhoorn, line 833: The message is `sensor_msgs/JointState`, the topic is `joint_states`.

If referring to configuration items in the configuration file, I'd suggest using a perma-link to the exact item, to avoid users having to search for them.

jimmy-mcelwain:
I should have been more clear in my previous comment -- I addressed @gavanderhoorn's comments and pushed the revisions. I just didn't click "resolve conversation". I'm never sure if I should resolve somebody else's comments after I address them, or let them decide for themselves if it is resolved.

Issue #385, opened 2025-03-05
Consider implementing a 'stop motion' ROS service


gavanderhoorn:
Context: #384 (but perhaps also #41).

`FollowJointTrajectory` action goals can be cancelled, but motion controlled via the `queue_traj_point` service doesn't have such an interface.

In fact, the [current implementation](https://github.com/Yaskawa-Global/motoros2/blob/34bfdaf4c16a1381ffd4e0d03af5d05001f1eea4/src/ServiceStopTrajMode.c#L43-L75) of `stop_traj_mode` will refuse to do anything if the robot is in motion (also documented [here](https://github.com/Yaskawa-Global/motoros2/blob/34bfdaf4c16a1381ffd4e0d03af5d05001f1eea4/doc/ros_api.md#stop_traj_mode)).

In order to allow for more control by clients, a `stop_motion` (or similarly named) service should be added which stops any ongoing motion.

Any implementation should take care not to interfere with the `FollowJointTrajectory` action state machine (ie: regular action goals) in unexpected ways. Either such a `stop_motion` service should not affect active goals (requiring action clients to cancel goals instead) or it should `ABORT` active goals, similarly to how any on-controller error or alarm would (note: not `CANCELLED`, as that should probably be only returned when clients cancel their action goals).

jimmy-mcelwain:
I like the idea of a stop motion service. I can spend some time working on this. 

Related to this, it might be too late in MotoROS2's development, but especially if we add a new service I think that it would be good to change the name of the already existing `stop_traj_mode` service. I have seen users confused, thinking that it would stop motion. It confused me at first. Maybe `reset_motion_mode`, `clear_motion_mode`, or something along those lines would be more intuitive. But that would obviously be a breaking change, so we could put it on hold or just not mess with it.

gavanderhoorn:
> I can spend some time working on this.

ok. Let's use this ticket to discuss potential designs.

> I think that it would be good to change the name of the already existing `stop_traj_mode` service. I have seen users confused, thinking that it would stop motion

[...]

PR #449, opened 2025-08-29
Realtime Interface (Joint and Cartesian)

ted-miller:
~~This is just a draft~~. Both @gavanderhoorn and @jimmy-mcelwain have made some feedback already. I'd like to have a centralized place for that feedback.

This PR adds a real-time control interface, intended to be used in a closed loop system. The idea is to minimize as much communication latency as possible. This will take the user data and pump it directly into the motion API with only the bare minimal processing.

There are currently two motion modes, which require a planned trajectory. This adds a new motion mode, which is activated by the `StartRtMode` [service](https://github.com/ted-miller/motoros2_interfaces/blob/experiment_rt/srv/StartRtMode.srv). 
(Note: I will be getting rid of [this argument](https://github.com/ted-miller/motoros2_interfaces/blob/0a578a45cd34360fe2da5d9640439fa749b2b9cd/srv/StartRtMode.srv#L22-L24) in the service.)

Once that is active, the client must open a UDP connection to port 8889 (*configurable). Once that is open, the client must send the first [command packet](https://github.com/ted-miller/motoros2/blob/bc0a616f4fa20ba4e761165f2214f95013056a19/src/RealTimeMotionControl.h#L18-L32), with an ID of `0`. Then it should wait for a reply from the robot, which contains the current feedback position of the robot. (_To be expanded with additional status info_...) Upon receiving that feedback packet, the client must immediately send the next command packet.

Please note that each command packet is an incremental motion **relative** to the current position. The return value of the `StartRtMode` service call will indicate the time period for each increment. (Default `4 ms` for a single robot arm.)

For this interface, the `StartRtMode` service invocation will indicate whether the increments will be in joint space (radians) or cartesian (currently mm and deg; this will probably change).

There is a test app which uses a USB joystick to move the arm in real time. [This one](https://github.com/ted-miller/rt_control_tester/tree/joints) is joint space. [This](https://github.com/ted-miller/rt_control_tester/tree/cartesian) is cartesian.

All testing has been done with Jazzy and YRC1000.


jimmy-mcelwain (commented):
I keep on getting the message `Micro-ROS PC Agent Disconnected` ~5 seconds after `Starting UserLan link state monitor (port: 1)`. But then it automatically reconnects, and then it disconnects again 5 seconds later. This is after a lot of messing around with combinations of `/start_traj_mode`, `/start_rt_mode`, and `stop_traj_mode`. I may have gotten into some sort of weird state. I've attached a large debug log here, it was happening near the end (ignore the reset alarm messages, those are unrelated). Since it happens almost exactly 5 seconds after the link state monitor begins, it seems like it's just failing to ping my PC. I don't know why, but I'll look into it more. 

I'm also seeing a memory leak. Probably related, but maybe not. 

[20250829T130735_motoros2_debug_log.txt](https://github.com/user-attachments/files/22052129/20250829T130735_motoros2_debug_log.txt)

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,315 @@
...
+    {
+        Ros_Debug_BroadcastMsg("ERROR: Failed to bind UDP socket for real-time motion control");
+        mpClose(*sockServer);
+        return false;
+    }
+
+    return true;
+}
+
+void Ros_RtMotionControl_PopulateReplyMessage(int sequenceId, RtReply* reply)
```
jimmy-mcelwain, line 302: It looks like right now, this doesn't give cartesian positions if you are in cartesian mode. It only gives joint angles.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,315 @@
...
+        FD_ZERO(&fds);
+        FD_SET(sockRtCommandListener, &fds);
+
+        tv.tv_usec = 0;
+        tv.tv_sec = g_nodeConfigSettings.timeout_for_rt_msg;
+
+        if (g_nodeConfigSettings.timeout_for_rt_msg != -1)
+            timeout = &tv;
+        else
+            timeout = NULL;
```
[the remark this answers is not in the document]
ted-miller: It does need to be reset every time.
mpSelect will modify the value of timeout, based on how much time elapsed.

src/RealTimeMotionControl.c
ted-miller: > I would suggest to always return both joint and Cartesian poses,

What's the rationale for that? 

It's easy to implement. But I don't see why someone would want both.

gavanderhoorn:
> Are we OK with mandating little-endianness?

it would be against *best practices* for network protocols (I'm old-fashioned when it comes to network protocols), but I guess with 'everything' being and/or defaulting to little-endian, it'd probably be OK. Would also make (de)serialisation easier, as clients could just use `std::memcpy(..)` from internal variables for fields/packets.

ted-miller:
Note to self:
The reason that the FSU detection isn't working properly is the order of operations. The command position needs to be checked before calling `mpExRcsIncrementMove`.

So now the question is whether I should read feedback position before or after...

jimmy-mcelwain:
> > I'm also seeing a memory leak.

> I'm not explicitly allocating any new memory on the heap. And AFAICT, I'm de-init'ing everything properly.

> I have reason to suspect that my libmicroros is outdated. For your testing, did you get that library from me? Or did you build yourself?

Whoops I missed this. I got the libmicroros from you. That's probably the problem.

doc/rt_control.md
```diff
@@ -0,0 +1,140 @@
...
+The command packet is a *packed* `RtPacket` structure.
+This contains a sequence ID, the increments for each control group, and the tool number to use for each control group.
+
+```c
+//##########################################################################
+//                  !All data is little-endian!
+//##########################################################################
+struct RtPacket
+{
+    UINT32 sequenceId;
```
jimmy-mcelwain, line 54: With the sequence increasing by `250` every second, the `UINT32` will overflow after 198 days of operation. I think it is extremely unlikely that this system will run for such a long period of time with no intervention considering the robot cannot run into any alarms and the PC and network would need to be nearly perfectly reliable, but I thought I'd bring it up at least.

doc/rt_control.md
```diff
@@ -0,0 +1,140 @@
...
+#### Joints
+
+When the `control_mode` is `JOINT_ANGLES (1)`, the order of the joints in the `delta` array must be in the order of `S L U R B T E 8`.
+Please note that for seven axis robots, the `E` joint is phyically mounted in the middle of the arm.
+But it must be sent at the end of the joint array.
+
+See `JointIndeces` enum.
+
+```c
+enum JointIndeces
```
jimmy-mcelwain, line 69: indices

doc/rt_control.md
```diff
@@ -0,0 +1,140 @@
...
+
+    MAX_JOINTS
+}
+```
+
+#### Cartesian
+
+When the `control_mode` is `CARTESIAN (2)`, the order of the joints in the `delta` array must be in order of `X Y Z Rx Ry Rz Re 8`.
+
+See `CartesianIndeces` enum.
```
jimmy-mcelwain, line 88: CartesianIndices

jimmy-mcelwain:
How is the client application supposed to know if the RT session has ended? It looks like the controller exits the increment loop and doesn't send anything to the client no matter what the problem is. A client-side timeout in case of e-stop, timeout, missed packets, etc and tracking `/robot_status` for errors/alarms may be sufficient. But maybe there could be an "end session" packet. It's probably okay as-is but I just want to check.

jimmy-mcelwain (changes requested):

ted-miller:
> How is the client application supposed to know if the RT session has ended? It looks like the controller exits the increment loop and doesn't send anything to the client no matter what the problem is. A client-side timeout in case of e-stop, timeout, missed packets, etc and tracking /robot_status for errors/alarms may be sufficient. But maybe there could be an "end session" packet. It's probably okay as-is but I just want to check.

I think that if we had an end-session packet, that we would want to include some reason code. But IMO, that information could just as easily be obtained from `robot_status`. So, I'm inclined to just let the client time out.

I have added some text to the docs/comments that the client should use the same timeout as the server.

ted-miller:
> speed limit

Since this can be configured on the client app and also through the FSU, I don't think it's necessary.

> reliability

I think that's due to ubuntu and not the robot controller.

> Also, do we want to "force" users to read the feedback messages? Technically they could just have a 4ms (or whatever the interpolation period is) timer on their client PC and fire packets without reading feedback. 

The [latest commit](https://github.com/ted-miller/motoros2/blob/4b72532d25a55de6f7fa1f39a5dbf7ba525ee11e/src/RealTimeMotionControl.c#L155) ensures that the two systems stay in sync. If the client attempts to send anything before the robot's "trigger", then the data gets discarded.

jimmy-mcelwain (commented):
A couple more notes:

There is a bug where the rotational portions of the cartesian feedback messages are being truncated to the nearest degree. This isn't very obvious because it is converted to radians before being sent to the client, but if you convert it to degrees it is easy to see. So the cartesian feedback messages can be off by up to 1 degree. 

If I end a session with an e-stop and then I start a new session, then I get the "you are being slowed down" message as a response to the first packet. It seems like it "remembers" from the previous session that it failed to complete a movement (because of the e-stop). I think something isn't being cleared properly between sessions. 

I am able to control the robot for several seconds past when the agent disconnects. I can kill the micro-ros agent on my PC and still control via a joystick until a few seconds later the cleanup happens. This is because we chose to use our own UDP message rather than ROS2. This means you can't call `/stop_traj_mode` or monitor the `/robot_status` topic during these few seconds. I don't know if we want to do something to avoid this behavior.

Unlike other services, you must do `/stop_traj_mode` and then `/start_rt_mode` to restart again. I don't have a problem with this, but with say trajectory mode, I can just call `/start_traj_mode` after an estop and it works. But with the RT interfaces, if I e-stop, then I I need to call `/stop_traj_mode` and then `/start_traj_mode` to start it up again.

jimmy-mcelwain:
It's also worth noting that the `/stop_traj_mode` actually does stop the robot during execution of real time motion. For other motion modes, `/stop_traj_mode` fails if the robot is actively executing a motion. Since this motion type doesn't have a clear "end" unlike the other motion modes, maybe this is okay. If we don't like this, we could instead only allow `/stop_traj_mode` after the session times out, or we could think about working on https://github.com/Yaskawa-Global/motoros2/issues/385

ted-miller:
So, the memory leak is not related to this PR. It is something to do with a multi-robot configuration.

This PR should be good to merge.

gavanderhoorn (changes requested):
I have some comments/questions, but can only post those later, so marking as *request changes*. Apologies for not posting them sooner @ted-miller.

config/motoros2_config.yaml
```diff
@@ -318,3 +318,35 @@ publisher_qos:
...
+#-----------------------------------------------------------------------------
+# For the real time motion interface, which port should the UDP messages
+# be transmitted on?
+#
+# DEFAULT: 8889
+#rt_udp_port_number: 8889
+
+#-----------------------------------------------------------------------------
+# Timeout for real time motion commands. If a command packet is not received
+# within this number of milliseconds, the motion mode will be cancelled.
```
gavanderhoorn: Does this affect only the 'first' packet, or all traffic?

doc/ros_api.md
```diff
@@ -133,6 +133,21 @@ Check the relevant fields of the `RobotStatus` messages to determine overall con
 
 The `reset_error` service can be used to attempt to reset errors and alarms
 
+### start_rt_mode
+
+Type: [motoros2_interfaces/srv/StartRtMode](https://github.com/yaskawa-global/motoros2_interfaces/srv/StartRtMode.srv)
+
+Attempts to enable servo drives, activate the real-time UDP server, and set the job-cycle mode to allow execution of INIT_ROS.
```
gavanderhoorn: suggestion: link to the documentation of *the real-time UDP server* here.

That link is also two lines down, but it's a bit confusing to refer to *the [..] server* here without immediately linking to it.

doc/rt_control.md
```diff
@@ -0,0 +1,142 @@
...
+This indicates the rate at which increment commands will be expected by the robot.
+The default `period` for a single robot arm is 4 milliseconds.
+However, that value will increase as additional axes or arms are added to the system.
+
+## Usage
+
+### Command Flow
+
+Once activated, a UDP server will listen on port `8889` (default).
+The user then sends the first increment with a `sequenceId = 0`.
```
gavanderhoorn:
```suggestion
The user then sends the first increment with the `sequenceId` field set to `0`.
```

doc/rt_control.md
```diff
@@ -0,0 +1,142 @@
...
+### Data format (command)
+
+The command packet is a *packed* `RtPacket` structure.
+This contains a sequence ID, the increments for each control group, and the tool number to use for each control group.
+
+```c
+//##########################################################################
+//                  !All data is little-endian!
+//##########################################################################
+struct RtPacket
```
gavanderhoorn: should this perhaps also include a bitmask/integer array indicating to which groups increments are being commanded?

That would make it an even thinner wrapper around `mpExRcsIncrementMove(..)` while making it possible to *not* have to send `0` increments for groups that *should not move*.

gavanderhoorn (commented):
I've commented mostly on the documentation and configuration file.

The implementation might need changes depending on discussion, so I didn't want to post comments on that yet.

config/motoros2_config.yaml
```diff
@@ -318,3 +318,35 @@ publisher_qos:
 # OPTIONS: USER_LAN1, USER_LAN2
 # DEFAULT: (all available network ports)
 #debug_broadcast_port: USER_LAN1
+
+#-----------------------------------------------------------------------------
+# For the real time motion interface, which port should the UDP messages
+# be transmitted on?
+#
+# DEFAULT: 8889
+#rt_udp_port_number: 8889
```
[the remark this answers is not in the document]
ted-miller: You're right, we should definitely move this to be within the range specified by the M+ manual. Given the upcoming CRA changes, I suspect that restrictions are going to get more... "restrictive". 

(We'll probably need to move the micro-ROS agent port too.)

doc/rt_control.md
```diff
@@ -0,0 +1,142 @@
...
+    TCP_Rz,
+    TCP_Re,
+
+    TCP_8,          //pulse
+
+    MAX_AXES
+}
+```
+
+Please note that rotations are applied in the order of `Z Y X`.
```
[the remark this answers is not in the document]
ted-miller: My brain doesn't like quaternions. My first instinct is, "no".

But... quats would be consistent with our `/tf` publisher. So, "maybe"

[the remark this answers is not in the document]
ted-miller: Ehhh... I really prefer yaw, pitch, roll. Since the idea is to be a minimal wrapper around the motion API, I think the data should be as close as possible to the 'final form'.

@jimmy-mcelwain Do you have an opinion?

BadBotX:
Hello @ted-miller

I'm working on MOTOROS2 and implementing some custom features. We build our own binaries, and the real-time capabilities are valuable for our application. We are using the Humble distribution and targeting the YRC1000 controller.

Could you please provide the custom libmicroros_yrc1000_humble library along with the .out file? Thank you.

ted-miller:
Hello @BadBotX 

Please note these compile fine, but have not in other way been tested.

[micro_ros_motoplus_yrc1000_humble-20240917-dbg_1776948819.zip](https://github.com/user-attachments/files/27012020/micro_ros_motoplus_yrc1000_humble-20240917-dbg_1776948819.zip)

[mr2_yrc1_h.zip](https://github.com/user-attachments/files/27012026/mr2_yrc1_h.zip)

Let us know any feedback.

> I'm working on MOTOROS2 and implementing some custom features

When they're ready, please consider submitting another PR with your new features.

cadkin:
@ted-miller 
Hello, sorry for yet another compile request but we're looking to try this on our YRC1000 with Jazzy. If you're willing, could you provide a `.out` file for this configuration?

Thank you!

cadkin:
> @ted-miller Hello, sorry for yet another compile request but we're looking to try this on our YRC1000 with Jazzy. If you're willing, could you provide a `.out` file for this configuration?
> 
> Thank you!

@ted-miller Hello, just wanted to ping again about getting a build of this PR for the YRC1000. Appreciate all your work on this!

ted-miller:
Hi @cadkin 

Here is a YRC1000 Jazzy build of this PR. Please let me know any feedback you have during testing.

[micro_ros_motoplus_yrc1000_jazzy-20240917-dbg_1778073143.zip](https://github.com/user-attachments/files/27443058/micro_ros_motoplus_yrc1000_jazzy-20240917-dbg_1778073143.zip)

~[mr2_yrc1_j.zip](https://github.com/user-attachments/files/27443073/mr2_yrc1_j.zip)~ _(Outdated)_

jimmy-mcelwain:
I'm going to look at this again soon. I would like to see if this could be made compatible with `ros2_control`.

cadkin:
> Here is a YRC1000 Jazzy build of this PR. Please let me know any feedback you have during testing.

@ted-miller Thanks again for providing this build, it's been very helpful for our use case. 

A small bit of feedback, it would be very nice if we could switch between cartesian and joint space movements on the fly. For example, our YRC1000 is configured with two movement groups, a GP12 arm and a D500F positioner. The cartesian movement works well for the arm, but joint-space movement makes a lot more sense for the positioner. 

Not sure how feasible this is, and it can certainly be worked around using an external IK solver for cartesian movement, so not super important.

> I'm going to look at this again soon. I would like to see if this could be made compatible with `ros2_control`.

@jimmy-mcelwain I've actually got a prototype `ros2_control` hardware interface using the build that I was provided.  It's undocumented, but might be good inspiration for a more official one:
https://github.com/cadkin/component_yaskawa_yrc1000/tree/master

It only supports a joint position interface, but it could certainly be extended to provide more.

jimmy-mcelwain:
@cadkin thank you for the link! I will take a look at that. I'll think about your feedback too, though I think for now it is unlikely.

src/MotionControl.c
```diff
@@ -1512,6 +1552,8 @@ MotionNotReadyCode Ros_MotionControl_StartMotionMode(MOTION_MODE mode, rosidl_ru
         }
     }
 
+    StartInterpolationTask(mode);
```
jimmy-mcelwain: `StartInterpolationTask` returns a bool that indicates whether it succeded or failed, but that bool is discarded

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+
+static int sockRtCommandListener = -1;
+static int sockRtStatusSender = -1;
+
+static struct sockaddr_in client_addr_status_messages;
+
+static LONG prevRtCmdPosition[MAX_GROUPS][MAX_AXES];
+static LONG howMuchShouldIHaveMoved[MAX_GROUPS][MAX_AXES];
+
+void Ros_RtMotionControl_HyperRobotCommanderX5(MOTION_MODE mode)
```
jimmy-mcelwain, line 31: Is this name just for fun, or is there some meaning behind it?

src/ConfigFile.h
```diff
@@ -109,6 +109,15 @@ typedef enum
...
 #define DEFAULT_ULAN_DEBUG_BROADCAST_PORT        CFG_ROS_USER_LAN1
 #endif
+
+#define DEFAULT_RT_LISTENER_UDP_PORT_NUMBER     "22000"
+#define DEFAULT_RT_STATUS_UDP_PORT_NUMBER       "22001"
+#define DEFAULT_RT_STATUS_SLEEP_PERIOD          10
+
+#define DEFAULT_TIMEOUT_FOR_RT_MSG      30000
+
+#define DEFAULT_MAX_SEQUENCE_DIFFERENCE 10
```
jimmy-mcelwain: These are different from the yaml file now

jimmy-mcelwain: Should be 
`DEFAULT_TIMEOUT_FOR_RT_MSG      5000`
and
`DEFAULT_MAX_SEQUENCE_DIFFERENCE 3`
to match (I think)

src/main.c
```diff
@@ -105,6 +105,14 @@ void RosInitTask()
     Ros_Controller_SetIOState(IO_FEEDBACK_RESERVED_7, FALSE);
     Ros_Controller_SetIOState(IO_FEEDBACK_RESERVED_8, FALSE);
 
+    //==================================
+    //REUSEADDR isn't working (not officially supported anyway), so
+    //I need to keep the R/T socket open forever. But I also can't
+    //open it on a subtask which will be deleted as connections come
+    //and go. So, I'm opening on the one persistent task that never
+    //ends.
+    Ros_RtMotionControl_OpenSocket();
```
jimmy-mcelwain, line 114: `Ros_RtMotionControl_OpenSocket` returns a bool that indicates whether it succeded or failed, but that bool is discarded

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        Ros_CtrlGroup_ConvertRosUnitsToMotoUnits(ctrlGroup, incomingCommand->delta[groupNo], pulse_increments);
+
+        // Copy pulse increments to moveData
+        for (i = 0; i < ctrlGroup->numAxes; i++)
+        {
+            moveData->grp_pos_info[groupNo].pos[i] = pulse_increments[i];
+            
+            if (abs(pulse_increments[i]) > ctrlGroup->maxInc.maxIncrement[i])
+            {
+                Ros_Debug_BroadcastMsg("ERROR: The increment for axis [%d] exceeds the maximum limit of [%d] pulse counts", pulse_increments[i], ctrlGroup->maxInc.maxIncrement[i]);
```
jimmy-mcelwain: `The increment for axis [%d]` is paired with `pulse_increments[i]`, meaning that it is not saying the axis number, but rather the pulse count. 

A rewording like: 

```c
Ros_Debug_BroadcastMsg("ERROR: Group [%d] Axis [%d] has been commanded to move [%d] pulse counts this increment, exceeding the maximum limit of [%d] pulse counts", groupNo, i, pulse_increments[i], ctrlGroup->maxInc.maxIncrement[i]);
```

or something more terse could be good. (notably `i` is in moto joint order)

gavanderhoorn:
> I'm going to look at this again soon. I would like to see if this could be made compatible with `ros2_control`.

Just to keep things connected: adding `ros2_control` compatibility was always (one of the) (a) goal(s) of this effort.

We've looked at previous approaches and implementations, such as [adv4ncr/motoman_ROS2](https://github.com/adv4ncr/motoman_ROS2) and [tingelst/motoman@realtime-motion-server](https://github.com/tingelst/motoman/tree/realtime-motion-server).

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        stateMsg.drives_powered = g_messages_RobotStatus.msgRobotStatus->drives_powered.val;
+        stateMsg.e_stopped = g_messages_RobotStatus.msgRobotStatus->e_stopped.val;
+        stateMsg.in_motion = g_messages_RobotStatus.msgRobotStatus->in_motion.val;
+        stateMsg.play_mode = (g_messages_RobotStatus.msgRobotStatus->mode.val == industrial_msgs__msg__RobotMode__AUTO);
+        stateMsg.motion_possible = g_messages_RobotStatus.msgRobotStatus->motion_possible.val;
+        stateMsg.error = g_messages_RobotStatus.msgRobotStatus->in_error.val;
+        if (g_messages_RobotStatus.msgRobotStatus->error_codes.size > 0)
+            stateMsg.error_code = g_messages_RobotStatus.msgRobotStatus->error_codes.data[0];
+        else
+            stateMsg.error_code = 0;
```
jimmy-mcelwain, line 659: These are all pulling from `g_messages_RobotStatus.msgRobotStatus`, but that is a NULL pointer while the agent is not connected. This code is unconditionally reached even before the agent spawns, and thus before `g_messages_RobotStatus.msgRobotStatus` is set. 

After a disconnection, it can also use after free. 

At the very least, this should just pull directly from the same array (or accessors) that `g_messages_RobotStatus.msgRobotStatus` is populated with. 

That being said, I don't even like the existence of this separate robot status message/thread personally. I think that it should just be a bitmask that is included in the reply.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+
+        moveData->grp_pos_info[groupNo].pos[TCP_Rx] = RAD_TO_DEG_0001(incomingCommand->delta[groupNo][TCP_Rx]);
+        moveData->grp_pos_info[groupNo].pos[TCP_Ry] = RAD_TO_DEG_0001(incomingCommand->delta[groupNo][TCP_Ry]);
+        moveData->grp_pos_info[groupNo].pos[TCP_Rz] = RAD_TO_DEG_0001(incomingCommand->delta[groupNo][TCP_Rz]);
+
+        moveData->grp_pos_info[groupNo].pos[TCP_Re] = RAD_TO_DEG_0001(incomingCommand->delta[groupNo][TCP_Re]);
+
+        moveData->grp_pos_info[groupNo].pos[TCP_8] = incomingCommand->delta[groupNo][TCP_8]; //pulse or micron (no known manipulators use this axis)
+
+        double vector = sqrt(pow(incomingCommand->delta[groupNo][TCP_X], 2) + //x^2
```
jimmy-mcelwain: should be magnitude, not vector

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+                             pow(incomingCommand->delta[groupNo][TCP_Z], 2)); //z^2
+
+        // Assuming 'elapsed_ms' is your variable for time in milliseconds.
+        const double max_speed_mm_per_ms = 1.5; // 1500 mm/sec is 1.5 mm/ms
+
+        if (vector > (max_speed_mm_per_ms * g_Ros_Controller.interpolPeriod))
+        {
+            Ros_Debug_BroadcastMsg("ERROR: The increment for the TCP exceeds the maximum limit of 1500 mm/sec");
+            return false;
+        }
```
jimmy-mcelwain: This is not going to work as expected (I think). `vector` is given in meters, and then it is compared to a unit given in millimeters ` (max_speed_mm_per_ms * g_Ros_Controller.interpolPeriod)`.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+    struct timeval tv;
+    struct timeval* timeout;
+
+    bool fsuLimitingDetected;
+
+    bool packetTypeOK;
+
+    //=========================================================================================
+
+    bzero(prevRtCmdPosition, MAX_GROUPS * MAX_AXES * sizeof(LONG));
```
jimmy-mcelwain, line 56: You should also zero out `howMuchShouldIHaveMoved`

src/MotionControl.c
```diff
@@ -1594,6 +1639,9 @@ void Ros_MotionControl_StopTrajMode()
     ioWriteData.ulAddr = g_Ros_Controller.ioStatusAddr[IO_ROBOTSTATUS_WAITING_ROS].ulAddr;
     ioWriteData.ulValue = 0;
     mpWriteIO(&ioWriteData, 1);
+
+    mpDeleteTask(g_Ros_Controller.tidIncMoveThread);
```
jimmy-mcelwain: What if `g_Ros_Controller.tidIncMoveThread` is `INVALID_TASK`? E.g if you call `/stop_traj_mode` twice in a row. This should be gated.

jimmy-mcelwain: Also, if the RT motion times out and the task ends, then this delete task is deleting a stale tid. This may be unlikely, but if that tid is recycled for another task it could be a big problem.

src/MotionControl.c
jimmy-mcelwain, line 1673: Should this include RT too? I am not sure.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+                        break; //drop the connection
+                }
+                else
+                {
+                    if (!Ros_RtMotionControl_ParseCartesian(&incomingCommand, &moveData))
+                        break; //drop the connection
+                }
+
+                fsuLimitingDetected = Ros_RtMotionControl_CheckForFsuInterference(mode, incomingCommand.toolIndex);
+
```
jimmy-mcelwain, line 186: The `bStopMotion` and `Ros_Controller_IsMotionReady()` should be considered before moving. If a user submits a `/stop_traj_mode`, it will set a hold but will continue running `mpExRcsIncrementMove` (and fail) if another increment is submitted.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        //NOTE: This isn't the best method for this. During testing, I had tool #2 selected
+        //      on the pendant, but I was commanding increments on tool #0. Because of this,
+        //      the first motion on each axis would trigger the FSU detection mechanism. But
+        //      it immediately recovers after one cycle.
+        mpGetToolNo(MP_R1_GID + i, &getToolResp);
+
+        cartSendData.sRobotNo = i;
+        cartSendData.sFrame = 1; //1 = RF
+        cartSendData.sToolNo = getToolResp.sToolNo;
```
jimmy-mcelwain: What about positioners and base axes, i.e. non manipulator control groups?

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        //================================================================================
+        //FB pos
+        //================================================================================
+        Ros_CtrlGroup_GetFBPulsePos(group, pulsePos_moto);
+
+        //Angles
+        Ros_CtrlGroup_ConvertMotoUnitsToRosUnits(group, pulsePos_moto, reply->feedbackPositionJoints[groupIndex]);
+        
+        for (int axis = 0; axis < MP_GRP_AXES_NUM; axis += 1)
+            degrees[axis] = RAD_TO_DEG_0001(reply->feedbackPositionJoints[groupIndex][axis]);
```
jimmy-mcelwain: What about prismatic axes?

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+
+        //Cart
+        mpConvAxesToCartPos(groupIndex, degrees, command->toolIndex[groupIndex], &figure, &coord);
```
jimmy-mcelwain: How about base and stations groups?

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        //        See https://github.com/Yaskawa-Global/motoros2/discussions/455
+        ctrlGroup.sCtrlGrp = groupIndex;
+        mpGetPulsePos(&ctrlGroup, &cmdPulse);
+
+        //rad
+        Ros_CtrlGroup_ConvertMotoUnitsToRosUnits(group, cmdPulse.lPos, reply->previousCommandPositionJoints[groupIndex]);
+        
+        //deg
+        for (int axis = 0; axis < MP_GRP_AXES_NUM; axis += 1)
+            degrees[axis] = RAD_TO_DEG_0001(reply->previousCommandPositionJoints[groupIndex][axis]);
```
jimmy-mcelwain: Again, what about prismatic axes?

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+
+        //Cart
+        mpConvAxesToCartPos(groupIndex, degrees, command->toolIndex[groupIndex], &figure, &coord);
```
jimmy-mcelwain: Again, what about base and station axes

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+
+        tv.tv_usec = (g_nodeConfigSettings.timeout_for_rt_msg % 1000) * 1000;
+        tv.tv_sec = g_nodeConfigSettings.timeout_for_rt_msg / 1000;
+
+        if (g_nodeConfigSettings.timeout_for_rt_msg != -1)
+            timeout = &tv;
+        else
+            timeout = NULL;
+
+        if (mpSelect(sockRtCommandListener + 1, &fds, NULL, NULL, timeout) > 0)
```
jimmy-mcelwain, line 86: `bStopMotion` should be considered (as I say below), but `bStopMotion` is a short-lived signal, and it is likely to become true and then false during `mpSelect` if the user stops sending commands after calling `/stop_traj_mode`. This makes things difficult.

src/MotionControl.c
jimmy-mcelwain: False shouldn't *always* be fatal, though, considering that the `ERROR - IncMoveTask is already allocated` path is false but shouldn't be harmful.

jimmy-mcelwain: If `StartInterpolationTask` succeeds but a later call (e.g the `mpStartJob` below) fails, then the interpolation task exists and doesn't get cleaned up until `/stop_traj_mode` is called. Since `Ros_MotionControl_ActiveMotionMode = mode` has not yet been set, the motion mode is still `INACTIVE`. Then if `/start_traj_mode` or `/start_point_queue_mode` is called (before calling `/stop_traj_mod`), they may be accepted even though they aren't in charge of the motion task. 

This should go just before `Ros_MotionControl_ActiveMotionMode = mode` and the return value needs to be verified.

src/RealTimeMotionControl.c
jimmy-mcelwain: Performing `mpDeleteTask` during the select may be what caused the socket to be un-rebindable. It may be better to periodically check if `/stop_traj_mode` was called, and if it was, exit the task gracefully.

src/RealTimeMotionControl.h
```diff
@@ -0,0 +1,171 @@
...
+typedef struct
+{
+    BOOL drives_powered;
+    BOOL e_stopped;
+    BOOL in_motion;
+    BOOL play_mode;
+    BOOL motion_possible;
+    BOOL error;
+    int error_code;
+} RobotState;
```
jimmy-mcelwain: Not packed (or versioned, not sure if that is necessary)

doc/rt_control.md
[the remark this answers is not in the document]
jimmy-mcelwain: If it is really applied in the order `X Y Z`, then maybe quaternions do make more sense.

cadkin:
Interesting behavior to report: we're currently looking into controlling our positioner with this PR and we're getting some weird looking feedback.

The configuration is a YRC1000 with a GP12 arm as the first group, and a D500F positioner as the second group. On an initial read, I'm getting the following RT packet back:

```
read: RtReply: {
  seq: 0
  fsu_interference: false
  target:
    - group 0
    -- current: s -1.3089205517410658
    -- current: l -0.608179280449005
    -- current: u -0.6831614118464461
    -- current: r -0.9930658733445238
    -- current: b -1.3446971460404773
    -- current: t 4.291187084746855
    -- current: e 0
    -- current: 8 0
    -- commanded: s -1.3089205517410658
    -- commanded: l -0.608179280449005
    -- commanded: u -0.6831614118464461
    -- commanded: r -0.9930658733445238
    -- commanded: b -1.3446971460404773
    -- commanded: t 4.291187084746855
    -- commanded: e 0
    -- commanded: 8 0
    - group 1
    -- current: s 0.10645976690497291
    -- current: l 0.30202762400545746
    -- current: u 0
    -- current: r 0
    -- current: b 0
    -- current: t 0
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.9420354795302762
    -- commanded: l -0.48971814891076343
    -- commanded: u 0
    -- commanded: r 0
    -- commanded: b 0
    -- commanded: t 0
    -- commanded: e 0
    -- commanded: 8 0
}

```

Notice how the first group's commanded position matches the current, but second group does not. This causes a lot of problems in the current implementation of my `ros2_control` hardware interface as it relies on these values corresponding to each other as appropriate.
 
I'm not sure if this is something that's specific to our setup or if this is a bug in the realtime mode. Any suggestions?

EricMarcil:
@cadkin I'm not sure what could be causing this. We can start by checking your system configuration. Can you send us a copy of your system CMOS.BIN (zipped so it is not too large).

cadkin:
> @cadkin I'm not sure what could be causing this. We can start by checking your system configuration. Can you send us a copy of your system CMOS.BIN (zipped so it is not too large).

@EricMarcil Thanks for taking a look, here's the CMOS from the pendant: 
[yaskawa_yrc1000_cmos.zip](https://github.com/user-attachments/files/30273228/yaskawa_yrc1000_cmos.zip)

(If you want to move this conversation to another channel so we don't clog up the PR, let me know.)

EricMarcil:
When looking at the positioner's configuration, it shows stars for the reducer values. I'm not sure if that is because it is a preset and they don't want users to change it, or if it is because this positioner is using special parameters which might explain why we are not calculating the same ratio. I will need to investigate further and get back to you.

EricMarcil:
@cadkin I've looked closer at the parameters and compared the conversion to pulses with our simulator. The parameters normally used are matching the conversion in our simulator. So, the controller settings are correct.

Would you be able to get me another set of values like the ones above, but also incude the controller pulse positions. You can display the controller pulse positions on the pendant by going to management mode, then ROBOT --> CURRENT POSITION then use the PAGE key to switch to the S1 group.  I also want the feedback pulses under ROBOT --> SERVO MONITOR (You need to be in management security to access that menu). You can also check the ROBOT --> COMMAND POSITION to see if you have values displaying (I'm not sure if motion through MotoRos populates those or not)

That should help us understand if the issue is on the command side or the current position side.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        LONG howMuchDidIActuallyMove[MAX_AXES];
+        bzero(howMuchDidIActuallyMove, sizeof(LONG) * MAX_AXES);
+
+        if (mode == MOTION_MODE_RT_JOINT)
+        {
+            //Should this be Ros_CtrlGroup_GetPulsePosCmd?
+            //Answer: No, it should not. That should only be used when converting incoming
+            //        positional commands that contain an absolute position.
+            //        See https://github.com/Yaskawa-Global/motoros2/discussions/455
+            ctrlGroup.sCtrlGrp = groupIndex;
```
jimmy-mcelwain: ctrlGroup.sCtrlGrp = g_Ros_Controller.ctrlGroups[groupIndex]->groupId;

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+    bzero(moveData, sizeof(MP_EXPOS_DATA));
+
+    for (i = 0; i < g_Ros_Controller.numGroup; i++)
+    {
+        moveData->ctrl_grp |= (0x01 << i);
+        moveData->grp_pos_info[i].pos_tag.data[0] = Ros_CtrlGroup_GetAxisConfig(g_Ros_Controller.ctrlGroups[i]);
+        moveData->grp_pos_info[i].pos_tag.data[3] = MP_INC_PULSE_DTYPE;
+
+
+        ctrlGroup.sCtrlGrp = i;
```
jimmy-mcelwain: ctrlGroup.sCtrlGrp = g_Ros_Controller.ctrlGroups[i]->groupId;

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,611 @@
...
+        reply->feedbackPositionCartesian[groupIndex][TCP_Re] = DEG_0001_TO_RAD(coord.ex1);
+
+        //================================================================================
+        //CMD pos
+        //================================================================================
+        //Should this be Ros_CtrlGroup_GetPulsePosCmd?
+        //Answer: No, it should not. That should only be used when converting incoming
+        //        positional commands that contain an absolute position.
+        //        See https://github.com/Yaskawa-Global/motoros2/discussions/455
+        ctrlGroup.sCtrlGrp = groupIndex;
```
jimmy-mcelwain: ctrlGroup.sCtrlGrp = g_Ros_Controller.ctrlGroups[groupIndex]->groupId;

jimmy-mcelwain (changes requested):

jimmy-mcelwain:
> going to maintenance mode, the ROBOT --> CURRENT POSITION then use the PAGE key to switch to the S1 group

Just a quick note, I'm guessing that @EricMarcil intended to say management mode. Not 100% sure. 

I would also like to say @cadkin that I'm guessing your problem is caused by a bug in this PR that I just left another review for. 

I do not have access to any external axes at my setup right now. I could try to provide a new build for you if you would like, @cadkin

cadkin:
> @cadkin I've looked closer at the parameters and compared the conversion to pulses with our simulator. The parameters normally used are matching the conversion in our simulator. So, the controller settings are correct.
> 
> Would you be able to get me another set of values like the ones above, but also incude the controller pulse positions. You can display the controller pulse positions on the pendant by going to management mode, then ROBOT --> CURRENT POSITION then use the PAGE key to switch to the S1 group. I also want the feedback pulses under ROBOT --> SERVO MONITOR (You need to be in management security to access that menu). You can also check the ROBOT --> COMMAND POSITION to see if you have values displaying (I'm not sure if motion through MotoRos populates those or not)
> 
> That should help us understand if the issue is on the command side or the current position side.

Sorry for the delayed reply, but here's what I collected for this:

<details>
<summary>RtReply packet</summary>

```
RtReply: {
  seq: 0
  fsu_interference: false
  target:
    - group 0
    -- current: s -0.9837605863538408
    -- current: l -0.5938710746500018
    -- current: u -0.7816552502750073
    -- current: r -1.002710608855453
    -- current: b -1.0126313942343401
    -- current: t 4.211933166636486
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.9837605863538408
    -- commanded: l -0.5938710746500018
    -- commanded: u -0.7816675221213104
    -- commanded: r -1.0027286027649882
    -- commanded: b -1.0126313942343401
    -- commanded: t 4.211971546257847
    -- commanded: e 0
    -- commanded: 8 0
    - group 1
    -- current: s 2.6253949914913175e-05
    -- current: l -0.2804117625629485
    -- current: u 0
    -- current: r 0
    -- current: b 0
    -- current: t 0
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.7080165213053785
    -- commanded: l -0.47819689476190613
    -- commanded: u 0
    -- commanded: r 0
    -- commanded: b 0
    -- commanded: t 0
    -- commanded: e 0
    -- commanded: 8 0
}
```

</details>

<details>
<summary>Pulse Values GP12</summary>
pulse_gp12

</details>

<details>
<summary>Pulse Values D500F</summary> 
pulse_d500f

</details>

<details>
<summary>Servo Monitor</summary>
servo_monitor

</details>

Let me know if you need any extra info from our pendant. Thanks again for the help!

> > going to maintenance mode, the ROBOT --> CURRENT POSITION then use the PAGE key to switch to the S1 group
> 
> Just a quick note, I'm guessing that @EricMarcil intended to say management mode. Not 100% sure.
> 
> I would also like to say @cadkin that I'm guessing your problem is caused by a bug in this PR that I just left another review for.
> 
> I do not have access to any external axes at my setup right now. I could try to provide a new build for you if you would like, @cadkin

If you could provide another build that would be wonderful. I've worked around it for the time being (deadlines are looming), but I definitely want to circle back around to this.

On a separate note, we've also been getting intermittent issues that require us to restart the entire controller:
<details>
<summary>Pendant Disconnection Error</summary>

pendant_disconnect

</details>

This seems to happen somewhat infrequently, but I've noticed that when it does it's usually when we first start the micro-ros bridge and only when using the preview build.

I'd be happy to try a new build and see if some of the tweaks have caused this issue to clear up.

EricMarcil:
The values supplied confirm that the reported current position is correct.

So, the issue would with the command position of the 2nd group. The values don't even make sense. It doesn't look like ratio error. I don't see how you could go from an almost zero value (2.6253949914913175e-05) to -0.7080165213053785. My guess is that the commanded values are garbage (not initialized or overwritten by some other data).

You say that this is at initialization, so you haven't sent any motion yet?

cadkin:
Yes, that is correct - this is the very first packet I get back after initializing the UDP connection. I should note that if I just ignore the commanded value, this offset doesn't seem to go away. E.g. here's the start packet vs a random packet from later in the lifecycle:

<details>
<summary>Packet Seq 0</summary>

```
RtReply: {
  seq: 0
  fsu_interference: false
  target:
    - group 0
    -- current: s -0.9748962351789675
    -- current: l -0.6016828718123094
    -- current: u -0.7087604832346817
    -- current: r -1.0200027559188911
    -- current: b -0.9091657597171571
    -- current: t 4.265511118056377
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.9748962351789675
    -- commanded: l -0.6016828718123094
    -- commanded: u -0.7087604832346817
    -- commanded: r -1.0200207498284264
    -- commanded: b -0.9091657597171571
    -- commanded: t 4.265511118056377
    -- commanded: e 0
    -- commanded: 8 0
    - group 1
    -- current: s 0.04084239475096659
    -- current: l -0.0594544268976211
    -- current: u 0
    -- current: r 0
    -- current: b 0
    -- current: t 0
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.7016368114760546
    -- commanded: l -0.48448711044167625
    -- commanded: u 0
    -- commanded: r 0
    -- commanded: b 0
    -- commanded: t 0
    -- commanded: e 0
    -- commanded: 8 0
}
```

</details>

<details>
<summary>Packet Seq 1405</summary>

```
read: RtReply: {
  seq: 1405
  fsu_interference: false
  target:
    - group 0
    -- current: s -0.9748962351789675
    -- current: l -0.6016828718123094
    -- current: u -0.7087604832346817
    -- current: r -1.0200207498284264
    -- current: b -0.9091657597171571
    -- current: t 4.265511118056377
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.9748962351789675
    -- commanded: l -0.6016828718123094
    -- commanded: u -0.7087604832346817
    -- commanded: r -1.0200207498284264
    -- commanded: b -0.9091657597171571
    -- commanded: t 4.265511118056377
    -- commanded: e 0
    -- commanded: 8 0
    - group 1
    -- current: s 0.04084239475096659
    -- current: l -0.0594544268976211
    -- current: u 0
    -- current: r 0
    -- current: b 0
    -- current: t 0
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.7016368114760546
    -- commanded: l -0.48448711044167625
    -- commanded: u 0
    -- commanded: r 0
    -- commanded: b 0
    -- commanded: t 0
    -- commanded: e 0
    -- commanded: 8 0
}
```
</details>

If I then rotate our positioner:

<details>
<summary>Packet Seq 5615</summary>

```
read: RtReply: {
  seq: 5615
  fsu_interference: false
  target:
    - group 0
    -- current: s -0.9748962351789675
    -- current: l -0.6016828718123094
    -- current: u -0.7087604832346817
    -- current: r -1.0200207498284264
    -- current: b -0.9091657597171571
    -- current: t 4.265511118056377
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.9748962351789675
    -- commanded: l -0.6016828718123094
    -- commanded: u -0.7087604832346817
    -- commanded: r -1.0200207498284264
    -- commanded: b -0.9091657597171571
    -- commanded: t 4.265511118056377
    -- commanded: e 0
    -- commanded: 8 0
    - group 1
    -- current: s 0.04084239475096659
    -- current: l -1.0574371338368207
    -- current: u 0
    -- current: r 0
    -- current: b 0
    -- current: t 0
    -- current: e 0
    -- current: 8 0
    -- commanded: s -0.7016368114760546
    -- commanded: l -0.48448711044167625
    -- commanded: u 0
    -- commanded: r 0
    -- commanded: b 0
    -- commanded: t 0
    -- commanded: e 0
    -- commanded: 8 0
}
```
</details>

It seems like the commanded value just never changes.

src/RealTimeMotionControl.c
ted-miller: The meaning is that it is the mark 5 version of the commander which gives hyper control of the robot. I thought that was clear.

src/RealTimeMotionControl.c
ted-miller: It gets zeroed [down below](https://github.com/ted-miller/motoros2/blob/2b18361dc0e530357bf1fc0c09bc4179b3540343/src/RealTimeMotionControl.c#L212). Is that not sufficient?

src/MotionControl.c
ted-miller: I don't understand this comment

src/RealTimeMotionControl.c
ted-miller: IMO, the current behavior is fine (and intended). When `mpExRcsIncrementMove` fails, it breaks out of the loop and ends the RT session.

src/MotionControl.c
jimmy-mcelwain: [Here is the whole function](https://github.com/ted-miller/motoros2/blob/9bdf8b836c97f432de7b49cb48e04b17cb338dfd/src/MotionControl.c#L1669-L1686).  

My question is whether:
```
if (Ros_MotionControl_IsMotionMode_PointQueue(){...}
```

Should be changed to 


```
if (Ros_MotionControl_IsMotionMode_PointQueue() && Ros_MotionControl_IsMotionMode_RealTime()) {...}
```

src/RealTimeMotionControl.c
jimmy-mcelwain: No because for the first increment of each session, it has data from the last session left over.

src/MotionControl.c
ted-miller: I don't think its necessary. Given the incremental nature of the r/t interface, the user must constantly monitor feedback position. So, they will know where they are and there wouldn't be concern of starting motion from the wrong location.

src/RealTimeMotionControl.c
ted-miller: I suppose that I should switch this to BF. 

As for positioners, the idea of commanding cartesian doesn't make any sense. So, we would have to mix cart and pulse. I don't think I want to support that at this time.

ted-miller: > I suppose that I should switch this to BF.

No, I shouldn't. Then the incoming command could conflict with the feedback data.

src/MathConstants.h
```diff
@@ -16,6 +16,7 @@
 #define DEGREES_PER_RAD (57.295779513082)
 
 // macro
+#define METERS_TO_MILLIMETERS(x)    (x * 0.001)
```
jimmy-mcelwain: This would be `MILLIMETERS_TO_METERS`, not `METERS_TO_MILLIMETERS`. Or should it be `x * 1000`?

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+        cartSendData.sRobotNo = i;
+
+        CtrlGroup* group = g_Ros_Controller.ctrlGroups[i];
+        if (group->groupId <= MP_R8_GID) //is a robot and not an external axis
+            cartSendData.sFrame = 1; //1 = RF
+        else if (group->groupId <= MP_B8_GID) //is a base track
+            cartSendData.sFrame = 0; //0 = BF
+        else
+        {
+            Ros_Debug_BroadcastMsg("ERROR: Group [%d] is an external positioner. Cartesian control mode is not supported for this group.");
```
jimmy-mcelwain: `%d` in string, but no matching argument

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+
+            difference = howMuchShouldIHaveMoved[groupIndex][axis] - howMuchDidIActuallyMove[axis];
+            if (abs(difference) > MAX_INCREMENT_DEVIATION_FOR_FSU_DETECTION)
+            {
+                //Ros_Debug_BroadcastMsg("howMuchShouldIHaveMoved[%d][%d] = %d", groupIndex, axis, howMuchShouldIHaveMoved[groupIndex][axis]);
+                //Ros_Debug_BroadcastMsg("howMuchDidIActuallyMove[%d] = %d", axis, howMuchDidIActuallyMove[axis]);
+                //Ros_Debug_BroadcastMsg("difference = %d", difference);
+                //Ros_Debug_BroadcastMsg("---------");
+
+                return TRUE;
```
jimmy-mcelwain: With a multi-group system, if this conditional block evaluates to true and it returns `TRUE` here, the outer for loop exits. So for example, if group 0 returns here, then the `prevRtCmdPosition` is inaccurate for groups 1 to N.

jimmy-mcelwain (commented):
I have a lot more comments to make, but just as a start here are a few things.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+            //Answer: No, it should not. That should only be used when converting incoming
+            //        positional commands that contain an absolute position.
+            //        See https://github.com/Yaskawa-Global/motoros2/discussions/455
+            ctrlGroup.sCtrlGrp = g_Ros_Controller.ctrlGroups[groupIndex]->groupId;
+            mpGetPulsePos(&ctrlGroup, &cmdPulse);
+        }
+        else if (mode == MOTION_MODE_RT_CARTESIAN)
+        {
+            cartSendData.sRobotNo = groupIndex;
+            cartSendData.sFrame = 1; //1 = RF
```
jimmy-mcelwain: What about for base track? One of your latest commits fixed this during the init ([see resolved comment thread](https://github.com/Yaskawa-Global/motoros2/pull/449#discussion_r3815998252)), but it is still a problem here.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+        //rad (or meter for linear track)
+        Ros_CtrlGroup_ConvertMotoUnitsToRosUnits(group, cmdPulse.lPos, reply->previousCommandPositionJoints[groupIndex]);
+
+        if (group->groupId <= MP_R8_GID) //is a robot and not an external axis
+        {
+            //deg
+            for (int axis = 0; axis < MP_GRP_AXES_NUM; axis += 1)
+                degrees[axis] = RAD_TO_DEG_0001(reply->previousCommandPositionJoints[groupIndex][axis]);
+
+            if (group->groupId <= MP_R8_GID) //is a robot and not an external axis
```
jimmy-mcelwain: `if (group->groupId <= MP_R8_GID)` appears twice here

jimmy-mcelwain: Also, just generally, we have [these helper functions](https://github.com/Yaskawa-Global/motoros2/blob/351e7c81b616e4c56987449cac81d9a24b7da14e/src/CtrlGroup.h#L134-L138). you can get rid of the manual checks if you use these

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+}
+
+//Essentially a clone of the /robot_status topic. But decoupled from the industrial_msgs/RobotStatus type.
+void Ros_RtMotionControl_SendRobotStatus()
+{
+    RobotState stateMsg;
+
+    int client_addr_len = sizeof(client_addr_status_messages);
+
+    if (g_messages_RobotStatus.msgRobotStatus == NULL) //may already be allocated in ControllerStatusIO.c
```
jimmy-mcelwain, line 637: This is not sufficient, because it only really works on the first connection. After the agent disconnects, `msgRobotStatus` is destroyed, but this is already in the `while (TRUE)` loop below and continues referencing the freed memory even through connect/disconnect cycles.

jimmy-mcelwain: I generally do not like the way that this `Ros_RtMotionControl_SendRobotStatus` is handled. It is always alive, sending messages once every 10 ms even if there is no client connected. The fact that it is reading from `msgRobotStatus` means that it only gives useful information when connected to the micro-ros agent and the `ControllerStatusIO` code is running. I mentioned this elsewhere, but I still feel that it would be better if this was a bitfield included in the main feedback message.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+    {
+        moveData->ctrl_grp |= (1 << i);
+        moveData->grp_pos_info[i].pos_tag.data[0] = Ros_CtrlGroup_GetAxisConfig(g_Ros_Controller.ctrlGroups[i]);
+        moveData->grp_pos_info[i].pos_tag.data[3] = MP_INC_RF_DTYPE;
+
+        //NOTE: This isn't the best method for this. During testing, I had tool #2 selected
+        //      on the pendant, but I was commanding increments on tool #0. Because of this,
+        //      the first motion on each axis would trigger the FSU detection mechanism. But
+        //      it immediately recovers after one cycle.
+        mpGetToolNo(MP_R1_GID + i, &getToolResp);
```
jimmy-mcelwain: Doesn't work for non-robot groups. Plus, has the same group index not matching the `groupId` problem

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+
+        cartSendData.sRobotNo = i;
```
jimmy-mcelwain: Using group index instead of `groupId`

src/RealTimeMotionControl.h
```diff
@@ -0,0 +1,175 @@
...
+
+} PACKED;
+typedef struct RtPacket_ RtPacket;
+
+
+//##########################################################################
+//                  !All data is little-endian!
+//##########################################################################
+
+struct RtReply_
```
jimmy-mcelwain, line 115: I do not feel strongly about this, but @gavanderhoorn mentioned before 

> a version field: clients should implement version checks and refuse to work with servers of incompatible versions (and vice-versa). This should make it at least a little 'safer' to introduce breaking changes on the server side

But right now only the command packet has the version, so it is only a one-way check. 

Considering the server rejects it if the command packet doesn't match, I don't have a big problem (if the command and reply have tightly coupled versions). But I thought I would mention it.

doc/rt_control.md
```diff
@@ -0,0 +1,222 @@
...
+    //Please note that for seven axis robots, the 'E' joint is phyically
+    //mounted in the middle of the arm. But it must be sent at the end
+    //of the joint array. See JointIndices enum.
+    //
+    //For joint-space, this will be radians of each joint.
+    //
+    //For cartesian, this will be meters and radians of the TCP.
+    //The order of the joints must be in the order of [X Y Z Rx Ry Rz Re 8].
+    //See CartesianIndices enum.
+    //Rotations are applied in the order of ZYX.
```
jimmy-mcelwain: ZYX not true for incremental motion per earlier [comment thread](https://github.com/Yaskawa-Global/motoros2/pull/449/changes#r2643972401). Decision still needs to be made

src/RealTimeMotionControl.c
jimmy-mcelwain: `msgRobotStatus` should probably be set to NULL upon being destroyed, but that only solves one part of the problem.

src/RealTimeMotionControl.c
jimmy-mcelwain: Also, as the comment implies, I do not think this is the best way of doing this. I think that first we make sure that it isn't called improperly for base/station axes, but maybe we can get rid of the check on the first cycle at some point.

src/ControllerStatusIO.c
```diff
@@ -168,27 +168,10 @@ BOOL Ros_Controller_Initialize()
     //==================================
     //create message for robot status
     //TODO(gavanderhoorn): use micro_ros_utilities_create_message_memory(..) instead
-    g_messages_RobotStatus.msgRobotStatus = industrial_msgs__msg__RobotStatus__create();
-    rosidl_runtime_c__int32__Sequence__init(&g_messages_RobotStatus.msgRobotStatus->error_codes, MAX_ALARM_COUNT + 1);
-
-    //==================================
-    // If not started, start the IncMoveTask (there should be only one instance of this thread)
-    if (g_Ros_Controller.tidIncMoveThread == INVALID_TASK)
+    if (g_messages_RobotStatus.msgRobotStatus == NULL) //may already be allocated in RealTimeMotionControl.c
```
jimmy-mcelwain, line 171: This is never set to NULL after free/destroy. This will publish freed memory.

src/RealTimeMotionControl.c
jimmy-mcelwain, line 295: `sRobotNo` only encodes R1-R8, but this is going up to `g_Ros_Controller.numGroup`, which doesn't really make sense. As it is right now, cartesian mode really wouldn't work with any robots that have any non-robot control groups. I'm not really sure what to do here, it seems like this will need a new approach.

src/RealTimeMotionControl.c
```diff
@@ -0,0 +1,646 @@
...
+            //Should this be Ros_CtrlGroup_GetPulsePosCmd?
+            //Answer: No, it should not. That should only be used when converting incoming
+            //        positional commands that contain an absolute position.
+            //        See https://github.com/Yaskawa-Global/motoros2/discussions/455
+            ctrlGroup.sCtrlGrp = g_Ros_Controller.ctrlGroups[groupIndex]->groupId;
+            mpGetPulsePos(&ctrlGroup, &cmdPulse);
+        }
+        else if (mode == MOTION_MODE_RT_CARTESIAN)
+        {
+            cartSendData.sRobotNo = groupIndex;
```
jimmy-mcelwain, line 549: Same as comment above, `sRobotNo` is only valid for robot control groups. This will not behave properly for anything with any base or station axes.

src/RealTimeMotionControl.h
ted-miller: I'm considering the version in the command packet to be a "session version". So, if we modify the reply packet, we'll increment the version in the command packet.

src/RealTimeMotionControl.c
ted-miller: I'm punting this for a future issue. I want to test behavior on actual hardware. So for now, only robots are allowed in cartesian mode.

doc/rt_control.md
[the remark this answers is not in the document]
ted-miller: I have changed it to use quats.

Fadi-AlMasalmah:
Hi @ted-miller and everyone,

Following up on my previous comment (30 March), I managed to build and test the `rt_standalone` branch on our robot. This is a report on my testing and a question about the delay and the control frequency. I apologize in advance if this is only-partially relevant to Motoros2, but this might be interesting for anyone who is trying to have realtime control.

### Our configuration

* Robot: Yaskawa HC20DTP
* Controller: YRC1000micro
* Controller software: `YBS4.24.01A(EN/FR)-00`
* External axes: none
* ROS 2 Jazzy
* Motoros2 commit: Mar 27, 2026 
* hash: 2b18361dc0e530357bf1fc0c09bc4179b3540343

### Mororos2 Setup
I locally merged this branch with [PR #188](https://github.com/Yaskawa-Global/motoros2/pull/188) to also publish the estimated TCP forces. I am using the RT interface through a custom `ros2_control` hardware interface forked from [motoros2_hw_interfaces](https://github.com/b-robotized/motoros2_hw_interfaces), and modified to follow the [rt_control_tester](https://github.com/ted-miller/rt_control_tester) repo. 

### Goal
To implement admittance control for contact with hard objects.

### Result
In general, the RT interface works with no apparent issues for me, it runs at 125hz and usually doesn't drop any packets. I was able to implement a stable admittance control for hand guidance. However, I cannot stabilize it for contact with hard objects, mainly due to the delay in command execution, as well as the poor quality of internal joint torque sensors.

### The ISSUE: 
We measure a delay of 220~240 ms between the commanded joint position and the corresponding measured position from `/joint_states`. A sample of this delay is shown in the images below.

Pasted Image 20260820171951_204

Pasted Image 20260820171607_924

I realized that this doesn't come from Motoros2 because I then measured it using the Pendant Oscilloscope with no Motoros2 involved and it gave similare delay.

### Attempts to Reduce the delay
For context, initially, 'start_rt_mode' responded with 8 ms as the cycle period.
I tested the solutions suggested in [ros-industrial/motoman#219](https://github.com/ros-industrial/motoman/issues/219) and other solutions from the discussion with Yaskawa support:
1. Set `S2C1584=1` to remove the controller averaging filter.

   This reduced the delay from  220 ms to approximately 180 ms.


2. Enabled `HTRAJON` in `INIT_ROS`.

   This produced almost no measurable improvement in our case.
   
3. Disabling PFL and arm interference check (Be careful)
	I had an exchange with Yaskawa support, which provided me with a CMOS.BIN to disable the PFL and arm interference check. This didn't work until I removed Motoros2 from the robot (to be investigated later), but it managed to reduce the cycle period down to 4 ms, and reduced the delay to 144 ms instead of 220 ms. 
	
### Now my final questions are:
- Did anyone manage to reduce the delay of motion execution? what is the minimal delay that we might get? 
- How can we take the advantage of 2~4 ms RT control period when the command has such a delay to be executed?

Thank you again for your work on this interface.

ted-miller:
@Fadi-AlMasalmah 

> Did anyone manage to reduce the delay of motion execution? 

At this time, the remedies in https://github.com/ros-industrial/motoman/issues/219 are the best solution that we have. 

So far, the other things that we have tried will damage the robot. This has been an ongoing issue and we are working with Yaskawa Japan to try to find a better solution.
(In my personal opinion, I don't think this will be improved on the YRC1000 platform.)

> what is the minimal delay that we might get?

It varies depending on the robot model. I personally haven't done any testing on an HC model. 

For a GP-12 / AR-1440, I have gotten about ~100ms delay using S2C1584.

> How can we take the advantage of 2~4 ms RT control period when the command has such a delay to be executed?

Unfortunately, I don't have an answer for this. Other people (smarter than me) have made various real-time adaptive applications work. I think that they are generally slow-moving applications, such as grinding or dispensing.

Issue #434, opened 2025-07-16
No alarm if `motoros2_config.yaml` fails to parse


ted-miller:
I was editing my joint names in the config file. I uncommented the array and modified the name. But I forgot to uncomment the key `joint_names:` above the array.

Everything booted just fine without any notification of an error. 

The debug log says, `Failed to parse: did not find expected key`.

We should raise an alarm if the yaml is invalid.

gavanderhoorn:
I was convinced we had an issue for this already, but apparently not, as I can't find it anymore.

ted-miller:
> I am in favor of 1

I concur

PR #452, opened 2025-09-05
Alarm and exit if config file is not properly parsed

jimmy-mcelwain:
Addresses #434 

It just calls `motoRosAssert_withMsg` if the parser encounters an problem. So it alarms and then enters the infinite loop, so users will have to upload a new config file and reboot. 

I need to test this still so I'm going to make it a draft.


src/ErrorHandling.h
```diff
@@ -176,6 +176,8 @@ typedef enum
     SUBCODE_FAIL_INVALID_BASE_TRACK_MOTION_TYPE,
     SUBCODE_DEBUG_INIT_FAIL_MP_NICDATA,
 
+    SUBCODE_CONFIGURATION_FILE_YAML_PARSING_ERROR,
```
ted-miller, line 179: The white space above this is triggering me. The others are in groups of 10.

src/ConfigFile.c
```diff
@@ -817,15 +817,13 @@ void Ros_ConfigFile_Parse()
...
-                    bOkToInit = FALSE;
-                    continue;
+                    motoRosAssert_withMsg(FALSE, SUBCODE_CONFIGURATION_FILE_YAML_PARSING_ERROR, "Cfg file yaml parsing error");
                 }
                 Ros_ConfigFile_CheckYamlEvent(&event);
                 event_type = event.type;
                 yaml_event_delete(&event);
             } while (event_type != YAML_STREAM_END_EVENT);
-
+
```
ted-miller, line 826: Since I've already complained about white space... I'm gonna do it again.

jimmy-mcelwain: Visual Studio doesn't show this on my PC, but I agree that this is annoying and I will get rid of it.

src/ErrorHandling.h
jimmy-mcelwain: Good point. I didn't realize that they were in groups of 10, I assumed that they were just grouped with alarms that were added at the same time. I'll fix that

ted-miller (approved):

PR #454, opened 2025-09-19
Clarify supported controllers

ted-miller:
Minor updates to README


gavanderhoorn (approved):
> no intent to support FS100

:( but perhaps it's good to be clear about it.

I won't forget about it though :)

PR #480, opened 2026-03-17
Update README to include Smart Pendant

### LINKED ISSUE #137: Setup procedure for Smart Pendant
I happen to be using the SP right now for another project. While trying to test MotoROS2, I have no idea how I'm supposed to set this up w/o switching to the annoying Classic Interface. 

We should really document the process.

Or... better yet, the SP supports [YIP packages](https://developer.motoman.com/en/yip/smart-packager). We could greatly simplify the installation process by maintaining an installer.

ted-miller:
Fixes #137 

Add clarification on how to use the Smart Pendant during setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ This is a virtual recreation of the legacy pendant interface.
 
 ### To access *Normal Operation* mode
 
- - touch `[Connect]`
+- touch `[Connect]`
 
 When you want to return to the Smart Pendant interface, touch the "power" icon in the top right.
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The following sections document how to download, install, configure, use and tro
   - YRC1000micro: `mr2_yrc1m_*.out`
 - *Online* mode:
   - DX200: insert USB stick with `motoros2_config.yaml` into `CN106` in the controller cabinet
-  - YC1000[micro]: load `motoros2_config.yaml` via `USER DEFINED FILES`
+  - YC1000[micro]: load `motoros2_config.yaml` via `USER DEFINED FILES` (Smart Pendant must use the *Classic Interface*)
 - start an instance of the [micro-ROS Agent](#running-the-micro-ros-agent) on a PC
 - reboot controller
 - verify [MotoROS2 is running](#verifying-successful-installation)
@@ -51,6 +51,7 @@ The following sections document how to download, install, configure, use and tro
 - [General Requirements](#general-requirements)
   - [Checking the system software version](#checking-the-system-software-version)
 - [Option Function compatibility](#option-function-compatibility)
+- [Using the Smart Pendant](#using-the-smart-pendant)
 - [Download](#download)
   - [Downloading the files](#downloading-the-files)
   - [Extracting the files](#extracting-the-files)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -206,6 +241,7 @@ Insert the storage device into the robot's programming pendant.
 If the controller is configured with the Functional Safety Unit (FSU), then `SAVE DATA CRC CHECK FUNC (FSU)` must be temporarily disabled during the installation procedure.
 
 Turn on the robot controller to enter *Normal Operation* mode.
+(see [Using the Smart Pendant](#using-the-smart-pendant))
```
jimmy-mcelwain: Mention specifically that classic interface is needed

README.md
```diff
@@ -251,6 +287,8 @@ In *Normal Operation* mode:
 
 #### YRC1000 and YRC1000 micro
 
+(see [Using the Smart Pendant](#using-the-smart-pendant))
```
jimmy-mcelwain: Mention specifically that classic interface is needed

README.md
```diff
@@ -274,6 +312,7 @@ Reboot again and there should be no alarms.
 If you receive any errors or alarms after rebooting, please refer to the [Troubleshooting](#troubleshooting) section for information on how to remedy the issue.
 
 If `SAVE DATA CRC CHECK FUNC (FSU)` was disabled at the start of this procedure, then it can now be reenabled.
+(see [Using the Smart Pendant](#using-the-smart-pendant))
```
jimmy-mcelwain: Mention that classic interface is needed

doc/troubleshooting.md
```diff
@@ -230,6 +232,8 @@ For YRC1000 and YRC1000micro: ensure the controller is updated to at least `YAS2
 If the system software version is below this, please contact Yaskawa Motoman for assistance with upgrading the controller.
 
 Then boot the controller into *Maintenance* mode by holding `{Main Menu}` on the keypad.
+(see [Using the Smart Pendant](../README.md#using-the-smart-pendant))
```
jimmy-mcelwain: Not linked in a consistent way. Sometimes it is with the "boot into maintenance mode" step and sometimes it is at the beginning of the troubleshooting instructions. 

More importantly though, I think that maybe it would be better if the text said something along the lines of "To access maintenance mode on the smart pendant, follow the instructions [here]"

jimmy-mcelwain (commented):
I think that this is generally complete enough, but rather than just linking the smart pendant section with little context, you should mention that smart pendant users must use classic mode (normal operation or maintenance) in addition to the link to instructions.

doc/troubleshooting.md
ted-miller: >  Sometimes it is with the "boot into maintenance mode" step and sometimes it is at the beginning of the troubleshooting instructions.

I don't understand this comment. Can you elaborate?

jimmy-mcelwain:
This seems like [this](https://github.com/Yaskawa-Global/motoros2/blob/4b70dde75cc8fcdd0ca668fefa8cb00d83037816/README.md?plain=1#L452-L457) should have a smart pendant note too. That being said, it is under the "Older Controller Software" header, which is probably an inappropriate place for it to be in the first place.

Issue #489, opened 2026-04-14
Inaccurate `IONAME.DAT` files


jimmy-mcelwain:
The `IONAME.DAT` files were seemingly copied from MotoROS1. They are no longer accurate. 

https://github.com/Yaskawa-Global/motoros2/blob/8ca01457a9fb76c5d1716458b4eb725ce0f55931/robot_jobs/single_arm/IONAME.DAT#L481-L484

Actual allocation:
https://github.com/Yaskawa-Global/motoros2/blob/8ca01457a9fb76c5d1716458b4eb725ce0f55931/src/ControllerStatusIO.h#L11-L18

The indices of `IONAME.DAT` and the actual allocations match, but the names sometimes differ from the purposes, and some allocations are not actually used at all. 

A colleague is trying to compile a list of reserved IO for different common applications along with their names and purpose. He asked about MotoROS(1) and MotoROS2 and pointed out that the IO has the same names, which it should not. 

We also reserve the following 8 bits. I suppose that we should include that in the list of IO that my colleague and the `IONAME.DAT` even though it isn't being used. 

https://github.com/Yaskawa-Global/motoros2/blob/8ca01457a9fb76c5d1716458b4eb725ce0f55931/src/ControllerStatusIO.h#L20-L27

ted-miller:
Can you elaborate on specifically what is wrong? It is difficult to correlate the indeces with the specific numbers and/or names. For instance, does `IO_FEEDBACK_WAITING_MP_INCMOVE` directly link to `line 482: ROS_READY`?

ted-miller:
AFAIK, nothing has changed about this since ROS1

jimmy-mcelwain:
This shows the relationship between the source code names and the IONAME.DAT names

| Source code | IONAME.DAT | IO index | OT# | Acceptable |
| -------- | -------- | -------- | -------- | -------- |
| `IO_FEEDBACK_WAITING_MP_INCMOVE` | `ROS_READY`| 11120   | 889 | Yes |
| `IO_FEEDBACK_MP_INCMOVE_DONE`    | `ROS_DONE` | 11121   | 890 | Yes |
| `IO_FEEDBACK_INITIALIZATION_DONE`| `INIT_DONE` | 11122   | 891 | Yes |
| `IO_FEEDBACK_CONTROLLERRUNNING`  | `CONNECTION_SRV` | 11123   | 892 | No |
[...]

PR #491, opened 2026-04-24
Update IONAME.DAT files

jimmy-mcelwain:
Addresses #489 

I'm open to changing the names if anybody has suggestions.


ted-miller (changes requested):
For reference: https://github.com/Yaskawa-Global/motoros2/blob/8ca01457a9fb76c5d1716458b4eb725ce0f55931/src/ControllerStatusIO.h#L11-L27

My only complaint is the `obsolete` label. We should just mark that also as `reserved`.

jimmy-mcelwain:
I just changed them to reserved 9 and 10. It's a little silly since they come before 1-8, but this way we don't have to change any source code. I can change the names in the source code instead if needed so it will be 1-10 in order. I just chose "obsolete" because they changed from being used in MotoROS [like this](https://github.com/ros-industrial/motoman/blob/846fbbc57c60223f12bf4795ea6404982d57709e/motoman_driver/MotoPlus/Controller.h#L39-L55) to being named `IO_FEEDBACK__` and `IO_FEEDBACK___` instead of being "reserved" like the others, so I assumed that meant they won't be used again.

gavanderhoorn (approved):
I've not tested this on-controller, but the changes make sense to me.

I don't feel we need to rename anything in any `.c`s: the changes here keep things compatible and avoid unexpected changes in behaviour on deployed systems.

gavanderhoorn:
@jimmy-mcelwain: please `rebase` first to reduce complicating history even more.

ted-miller (approved):

PR #499, opened 2026-06-12
Kilted support

jimmy-mcelwain:
Adds kilted support. 

Depends on https://github.com/Yaskawa-Global/micro_ros_motoplus/pull/22 (and consequently https://github.com/Yaskawa-Global/micro_ros_motoplus/pull/21) and https://github.com/Yaskawa-Global/micro_ros_motoplus_buildscripts/pull/167. 

I added `--ipc=host` to the docker run example because otherwise the micro-ros agent wasn't publishing properly. I could list topics and nodes, but not echo. But after adding `--ipc=host`, it worked. 

I haven't thoroughly tested. I loaded on a binary that I built and verified that `/joint_states` was publishing.

I don't know when Lyrical will make it upstream for micro-ros, but hopefully having done all of this will make it easier to port once it does come.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -349,6 +349,9 @@ healthchecksdb
 .vscode/
 
 # M+ libmicroros distribution directories
+libmicroros_yrc1000_kilted/
+libmicroros_yrc1000u_kilted/
+libmicroros_dx200_kilted/
 libmicroros_yrc1000_jazzy/
 libmicroros_yrc1000u_jazzy/
 libmicroros_dx200_jazzy/
--- a/MotoROS2.sln
+++ b/MotoROS2.sln
@@ -11,14 +11,17 @@ Global
 		DX200_galactic|x86 = DX200_galactic|x86
 		DX200_humble|x86 = DX200_humble|x86
 		DX200_jazzy|x86 = DX200_jazzy|x86
+		DX200_kilted|x86 = DX200_kilted|x86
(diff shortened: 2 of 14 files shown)


gavanderhoorn:
I haven't tested anything, but as a first comment: please update the `build VS proj` workflow to add Kilted to the [matrix](https://github.com/Yaskawa-Global/motoros2/blob/348ce29ec9607741242191bafad37f6c4621d314/.github/workflows/ci_msbuild.yaml#L62-L67).

It will fail until Yaskawa-Global/micro_ros_motoplus#22 is merged, but that would be a good way to figure out whether this PR can be merged or not.

gavanderhoorn (changes requested):
Marking as `changes requested` to avoid accidental merge.

jimmy-mcelwain:
> please update the `build VS proj` workflow to add Kilted to the [matrix](https://github.com/Yaskawa-Global/motoros2/blob/348ce29ec9607741242191bafad37f6c4621d314/.github/workflows/ci_msbuild.yaml#L62-L67)

I haven't done that yet because I do not know what the name of the `micro_ros_motoplus` release will be (since it has the build date in it). Should I have a placeholder, or should I put up a Kilted pre-release to `micro_ros_motoplus`?

ted-miller:
I just realized that support for Kilted ends in November. Do we still care about this?

PR #502, opened 2026-07-09
ci: use CI token for u-ROS M+ download

gavanderhoorn:
As per title.

This switches the MSBuild CI workflow to use a regular `github.token` instead of a special token we acquire using `actions/create-github-app-token`.

We previously used that action because `yaskawa-global/micro_ros_motoplus` was a private repository, but that is no longer the case.

See [runs/29005919982](https://github.com/Yaskawa-Global/motoros2/actions/runs/29005919982) for a test run with this change (ignore the failed jobs, they're failing for a different reason -- the `Download M+ libmicroros` step is successful, which is what this PR changes).

### DIFF
--- a/.github/workflows/ci_msbuild.yaml
+++ b/.github/workflows/ci_msbuild.yaml
@@ -83,19 +83,10 @@ jobs:
     - name: Find MSBuild and add to PATH
       uses: microsoft/setup-msbuild@v3
 
-    - name: Generate token (micro_ros_motoplus)
-      id: gen_token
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
This avoids this warning as well:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/create-github-app-token@v1

gavanderhoorn:
#477 can be closed if we merge this.

ted-miller (approved):
Looks good

PR #503, opened 2026-07-09
ci: switch to Bash for SDK download

gavanderhoorn:
As per title.

Bash -- as a shell -- has better support for exit-early behaviour in case commands fail.

Windows Cmd does not.

This also switches to `curl` (from `wget`) as that's more universally available on the runners it seems.


gavanderhoorn:
This would've been helpful when diagnosing the recent CI issues (broken SDK downloads) as jobs would have failed due to the failed download.

ted-miller (approved):
That is much nicer.

PR #508, opened 2026-09-08
Add a deeper motion queue for drip-feeding continuous motion

fake-name:
Hello!

This is a relatively simple tweak that extends the drip-feed interface to allow a compile-time-configurable ring-buffer of enqueued moves.

This basically exists because I have a project that needs to feed the robot complex (far more then 200) point trajectories, and the existing one-move-at-a-time interface has nearly no buffer, so if you are doing things like doing linear interpolation in joint space, it can be difficult to not accidentally let the buffer run out and produce motion discontinuities. 

It has two modes. Unless enabled, it should behave *exactly* as the existing code does (allowing one buffered move, returning BUSY if it's full). 

When the queue is turned on, it allows enqueuing up to `POINT_QUEUE_DEPTH` moves (default: 32). 

I cannot test this fully myself, as this involves a new message, and apparently you need some closed-source components to add messages?

The ring-fifo was exhaustively tested at least (see `/test/host`), and the rest of it reads fine.

Sidenote: it turns out you can build motoros2 on linux without issue via WINE. I can write that up of y'all are interested.

Disclosure: LLMs were used heavily in writing the code here. 
The architectural decisions were all mine, though. And I have reviewed the generated code.


fake-name:
Ok, so more things! Now tested on actual hardware (HC10DTP + YRC1000micro). Happily streams 10 points/sec indefinitely with zero stops.

Other things here that may be intriguing:
 - Everything I did here was exclusively built on linux. Turns out the windows tooling works fine under WINE.
 - I believe I (and the clanker) wound up reverse-engineering the proprietary bits needed to build things to the extent that we were able to add new messages and services to the binary.

ted-miller:
Hi @fake-name 

I'm not sure that I understand the motivation behind this modification. The queue on top of another queue seems complicated. Have you tried simply increasing the `Q_SIZE` by a significant amount? https://github.com/Yaskawa-Global/motoros2/blob/351e7c81b616e4c56987449cac81d9a24b7da14e/src/CtrlGroup.h#L12

The current value of 200 gives you 0.8 seconds of queued increments. If you crank that up to something like 1000, then you'd have 4 seconds of buffered motion increments.

fake-name:
Hello!

Yes, I initially looked at FollowJointTrajectory, but the issue I have there is that it's discrete movement sections, with a zero-velocity stop between each section. There does not seem to be a way to add *more* points to a in-progress FJT command while it is executing. 

I need to stream arbitrarily long generated trajectories, so I need to be able to enqueue things while the current motion is in progress. 

A fair bit of the complexity here comes from the fact that I tried to make this a purely additive change, e.g. it maintains the existing behavioural semantics *exactly*, just adds a new mode. I had to add a new queue on top of the existing structure to do that, unfortunately.

It might make more sense to restructure things so that the enqueue motion command (either the existing one, or my additions) can drop new motion commands directly into the `ControlGroup.inc_q`, but honestly this is big enough and complex enough that I didn't want to do that kind of substantial refactor if I could avoid it.

Since there was already an API that was pretty obviously *trying* do what I needed here, I just went with "let's extend that".

ted-miller:
I'm not suggesting using FJT. Rather, use the existing `queue_traj_point`. 

When the `QueueTrajPointRequest` is received, it is broken up into small 4 ms increments (same mechanism as FJT). But that buffer of increments can currently only hold 200 (note: this is unrelated to the 200 limit on FJT points). So, the buffer can only hold 0.8 seconds of motion. 

If the buffer fills up before processing the entire `QueueTrajPointRequest`, then the service will return `BUSY`. Once the entire `QueueTrajPointRequest` has been processed, you can queue another point. You essentially have 0.8 seconds to submit that point before your incremental buffer is emptied. 

Increasing the `Q_SIZE` would give you more time between the end of BUSY-status and the buffer emptying.

fake-name:
~~Yeah, I get that, but that means if I'm feeding points at ~10 Hz, and I have a network hiccup (this has happened in my testing even with a dedicated wire right from my computer to the robot), 100 ms of delay can cause a motion discontinuity.~~

~~Increasing Q_SIZE let's you more densely interpolate a *single motion*, but I'm specifically streaming extremely dense point paths, to allow me to effectively perform Cartesian motion with a negligible chord error.~~

~~If you're streaming *lower rate* points (where individual motion segments aren't in the tens/hundreds of milliseconds range), the existing interface is fine. I am not doing that.~~

Oooh, I think I see what you're saying. `queue_traj_point` requests are *asynchronous to the motion*, processed into the internal queue. So in that case this is pointless. Guh.

fake-name:
~~Really, I would have been 100% fine just adding a new direct-to-internal-`Q_SIZE` queue object, but that seems like a lot more rework, so I did the lower effort thing here.~~

~~Actually, all that said, would that be a preferred approach? I can rework things to use that approach if you would like that better.~~

~~(this actually started as "just let me call `queue_traj_point` more", but it then grew enough complex semantics that it wound up needing it's own service, making the origin of where I started everything kind of irrelevant. As usual, the first time I do anything I wind up figuring out how to not do it stupid *the next time*.)~~

gavanderhoorn:
> Oooh, I think I see what you're saying. queue_traj_point requests are asynchronous to the motion, processed into the internal queue.

exactly.

fake-name:
Gah. Well, sorry about the PR churn.

gavanderhoorn:
No need to apologise.

In fact, thank you for making the effort to draft this and submitting a PR. We very much welcome contributions like this.

I would however perhaps suggest to first come and discuss potential enhancements/fixes (post something on the [Discussion forum](https://github.com/Yaskawa-Global/motoros2/discussions). It could save you (quite) some time and potentially prevent duplicated effort as well.

fake-name:
There may still be _some_ value here, in that I at least did figure out how to do the entire build under linux+wine. 

If that's of interest, I can probably pull that out into a separate PR. Despite the Official SDK being very "you have to use this in visual studio", it works completely fine without it.

